PR #505, opened 2026-07-24
Implement infinite scrolling using Intersection Observer API

man-singh-dev:
This PR refactors the `ActivityFeed` component from a spinner-based pagination experience to a smooth infinite scrolling implementation using the native **Intersection Observer API**, without introducing any additional dependencies.

The primary changes are contained in `ActivityFeed.tsx`, with a small cleanup in `ActivityItem.tsx` and the addition of a reusable `useIntersectionObserver.ts` hook under `src/hooks/`. The new hook exposes a stable ref callback that observes a lightweight sentinel element placed at the end of the feed. It uses a `rootMargin` of `200px` to trigger the next page fetch before the user reaches the end of the list, providing a seamless scrolling experience. The observer is properly disconnected whenever additional fetching is disabled or the component unmounts, preventing unnecessary observers and potential memory leaks.

To improve the user experience, the initial loading spinner has been replaced with skeleton cards that match the feed layout, while subsequent page loads display appended skeleton placeholders instead of blocking the interface with a spinner. Duplicate requests are prevented through both TanStack Query's built-in request deduplication and conditional observer activation, ensuring additional fetches are only triggered when appropriate. Existing loading, error, and empty states have been preserved, with an additional end-of-feed indicator displayed once all available content has been loaded.

As part of the cleanup, the unnecessary wrapper element previously used to attach a reference to the last feed item was removed, reducing redundant DOM nodes, and the unused default React import was removed from `ActivityItem.tsx` to align with React 19's automatic JSX transform and the project's existing code style.


nensii21:
@man-singh-dev CI jobs are failing.Update it.

man-singh-dev:
Got it. I'll fix the CI

nensii21:
Let me know once you do.So i can merge it.

PR #826, opened 2026-08-01
feat: add Redis caching for profiles, popular feed, and notifications

### LINKED ISSUE #529: Redis Caching Layer
Cache

Profiles
Projects
Popular feeds
Notifications

revatikadam0607:
## Summary
This PR extends the existing Redis caching layer (introduced for Projects) to cover the remaining areas requested in #529: Profiles, Popular Feeds, and Notifications.

## Changes
- Cached `UserService.get_user`, used by the profile endpoint, so repeated profile views hit Redis instead of the database.
- Applied the existing (previously unused) `@cached` decorator to the `get_feed` route in `activities.py`, caching the public/popular feed for 60 seconds.
- Added Redis caching to `NotificationService.unread_count`, which is polled frequently by the frontend for the notification badge, with a short 30-second TTL to keep it responsive.

## Why
These endpoints are read-heavy and frequently polled, but were not yet using the multi-level (in-memory + Redis) cache already set up in `app/core/cache.py`. This reduces database load without changing any response formats.

## Testing
- Verified each endpoint still returns the same response shape.
- Confirmed cache keys are scoped correctly per user/resource so no data leaks across users.

Closes #529 

@nensii21 
I've resolved the issue. Kindly review it, and if everything looks good, please consider merging the corresponding PR.

### DIFF
--- a/backend/app/routers/activities.py
+++ b/backend/app/routers/activities.py
@@ -71,8 +71,8 @@ def get_activity(
     "/",
     response_model=list[ActivityResponse],
 )
-def get_feed(
-    limit: int = Query(50, ge=1, le=100),
+@cached(ttl=60, key_prefix="feed")
+def get_feed(    limit: int = Query(50, ge=1, le=100),
     cursor: datetime | None = Query(
         None, description="Cursor for pagination (created_at timestamp)"
     ),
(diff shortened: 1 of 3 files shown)


revatikadam0607:
Hey @nensii21 ! I noticed that you removed the ECSoC26-L1 label. That label is important because it's used to determine my contribution points. Could you please add the ECSoC26 label back? Once it's re-added, the ECSoC bot should automatically apply the Level 1 label again.

PR #859, opened 2026-08-03
Add ETag support and conditional request handling

### LINKED ISSUE #854: Add ETag / conditional request support to GET endpoints
## Description

Every `GET` endpoint in the API currently returns a full response body on every request, even when the underlying data has not changed since the client last fetched it. There is no `ETag` header emitted anywhere in the backend and no handling of the `If-None-Match` request header, so browsers and the TanStack Query cache in the frontend cannot take advantage of HTTP-level revalidation.

This is wasteful on hot read paths such as the dashboard, the project list, the feed, and the notification poller, all of which are re-fetched frequently and usually return identical bytes.

`app/core/cache.py` already gives us an L1/L2 cache, but that only saves database work — the full payload is still serialised and pushed over the wire.

## Proposed solution

Add ETag support as ASGI middleware so it applies uniformly without touching every router:

- Compute a strong `ETag` from the response body for successful (`200`) `GET` and `HEAD` responses.
- When the incoming request carries `If-None-Match` and it matches, return `304 Not Modified` with an empty body and the caching headers preserved.
- Emit a sensible `Cache-Control` default (`private, no-cache`) so clients revalidate rather than serve stale content.
- Skip streaming responses, file responses, WebSocket upgrades, and any endpoint that already sets its own `ETag`.

## Acceptance criteria

- [ ] `ETag` present on 200 JSON `GET`/`HEAD` responses.
- [ ] Matching `If-None-Match` yields `304` with no body.
- [ ] Non-matching `If-None-Match` yields the normal `200` payload.
- [ ] `POST`/`PUT`/`PATCH`/`DELETE` responses are untouched.
- [ ] Weak validators (`W/"..."`) and the `*` wildcard are handled per RFC 9110.
- [ ] Streaming and file responses pass through unmodified.
- [ ] Unit tests covering the above.
- [ ] Documented in `docs/api.md`.

MOHITKOURAV01:
# Pull Request

## Summary

Adds `ETag` headers to safe API responses and handles `If-None-Match`, so a client that already holds the current representation gets a bodyless `304 Not Modified` instead of the same JSON again.

The dashboard, project list, feed and notification poller all re-fetch on a short cadence and almost always get back identical bytes. `app/core/cache.py` already avoids the database round-trip for those, but the payload was still being serialised and pushed over the wire every time. This closes that gap.

---

## Related Issue

Closes #854

---

## Type of Change

- [ ] Bug fix
- [x] New feature
- [x] Documentation update
- [x] Performance improvement
- [ ] Refactoring
- [ ] UI/UX enhancement
- [x] Tests
- [ ] Other

---

## Changes Made

- `app/core/http_cache.py` — new module with the tag primitives: `generate_etag`, `parse_if_none_match`, `etag_matches`, `is_etaggable_content_type`, `merge_vary`. Kept separate from the middleware so the RFC 9110 comparison rules are testable on their own.
- `app/middleware/etag.py` — `ETagMiddleware`. Tags `200` JSON/text responses to `GET`/`HEAD`, and converts a matching conditional request into a `304`.
- `app/core/config.py` — `ENABLE_ETAG`, `ETAG_MAX_BODY_SIZE`, `ETAG_CACHE_CONTROL`.
- `app/main.py` — registers the middleware just outside `SecurityHeadersMiddleware`, so a `304` still carries the security headers.
- `tests/test_etag.py` — 38 tests.
- `docs/api.md` — new "Conditional Requests (ETag)" section with a worked request/response pair.

### Notes on the implementation

**Why raw ASGI instead of `BaseHTTPMiddleware`.** Hashing requires the full body. `BaseHTTPMiddleware` hands you an already-committed async iterator, so once you have consumed it to hash it you can no longer decide to stream instead. Writing it at the ASGI level lets the middleware hold the `http.response.start` message back, buffer as it goes, and abandon buffering entirely once the body passes `ETAG_MAX_BODY_SIZE` — at which point everything collected so far is flushed and the remainder streams through untouched. That keeps large exports and file downloads off the heap.

**What is deliberately skipped:** non-`200` responses, non-JSON/text content types, already-encoded bodies, responses where the handler set its own `ETag`, and anything marked `Cache-Control: no-store`. Empty bodies get no tag either — hashing `b""` would hand every empty response the same validator, which is worse than having none.

**`Vary: Authorization`** is added because payloads are user-scoped. Combined with `Cache-Control: private`, a shared proxy cannot serve one user's response to another.

`304` responses drop `Content-Length`, `Content-Type` and the other payload-describing headers, per RFC 9110 §15.4.5 — leaving them on an empty body confuses intermediaries.

---

## Testing

- [x] Tested locally
- [x] Existing tests pass
- [x] Added new tests
- [ ] UI verified

Additional testing notes:

`pytest backend/tests/test_etag.py` — 38 passed. Coverage includes:

- ETag stability across identical bodies, and divergence when the body changes.
- Weak validators, comma-separated tag lists, and the `*` wildcard.
- `304` carries the ETag but no `Content-Type` and no body.
- A stale `If-None-Match` still returns the full payload with a fresh tag.
- `POST`, `404`, binary, streaming, handler-supplied-ETag, `no-store` and empty-body responses are all left alone.
- A response over the size cap streams through intact and untagged.
- `ENABLE_ETAG=false` disables the whole thing.

---

## Checklist

---

## Additional Notes

This is transparent to existing clients — a client that never sends `If-None-Match` sees exactly what it saw before, plus one extra response header.

Follow-up worth considering separately: wiring `If-Match` into the write paths for optimistic concurrency. The comparison helpers in `http_cache.py` already cover the strong-comparison case that would need, but the semantics on writes are a different discussion and did not belong in this PR.

### DIFF
--- a/backend/app/core/config.py
+++ b/backend/app/core/config.py
@@ -216,6 +216,21 @@ class Settings(BaseSettings):
     CROSS_ORIGIN_EMBEDDER_POLICY: str = "require-corp"
 
     # ==========================================================
+    # HTTP Caching (ETag / conditional requests)
+    # ==========================================================
+
+    ENABLE_ETAG: bool = True
+
+    # Responses larger than this are streamed through without a validator
+    # rather than buffered in memory to be hashed. 1 MiB comfortably covers
+    # every JSON payload the API produces today.
+    ETAG_MAX_BODY_SIZE: int = 1024 * 1024
+
+    # "no-cache" means "revalidate before reuse", not "do not store", so the
+    # client keeps the body and we get to answer with a 304.
+    ETAG_CACHE_CONTROL: str = "private, no-cache"
+
+    # ==========================================================
     # Celery
     # ==========================================================
 
--- a/backend/app/core/http_cache.py
+++ b/backend/app/core/http_cache.py
@@ -1,0 +1,124 @@
+"""
+HTTP caching primitives.
+
+Small, dependency-free helpers for entity tags and conditional requests as
+described in RFC 9110 sections 8.8 and 13.1. Kept separate from the middleware
+so the parsing/comparison rules can be unit tested on their own.
+"""
+
+import hashlib
+from typing import Iterable, List, Optional
+
+# Media types we are willing to fingerprint. Anything else (images, PDFs,
+# octet-streams) is either already cheap to revalidate or too large to buffer.
+ETAGGABLE_CONTENT_TYPES: tuple[str, ...] = (
+    "application/json",
+    "application/problem+json",
+    "text/plain",
+    "text/html",
+)
+
+# Header names are compared lowercase throughout; ASGI guarantees byte keys but
+# not a particular case.
+_WILDCARD = "*"
+
+
+def generate_etag(body: bytes, weak: bool = False) -> str:
+    """
+    Build an entity tag for a fully buffered response body.
+
+    Uses SHA-256 truncated to 32 hex characters. That is 128 bits of digest,
+    which is far more than enough to make an accidental collision between two
+    payloads of the same resource impossible in practice, while keeping the
+    header short.
+    """
+    digest = hashlib.sha256(body).hexdigest()[:32]
+    tag = f'"{digest}"'
+    return f"W/{tag}" if weak else tag
+
+
+def parse_if_none_match(header_value: Optional[str]) -> List[str]:
+    """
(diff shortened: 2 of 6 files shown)


MOHITKOURAV01:

| :--- | :--- | :--- | :--- |

PR #860, opened 2026-08-03
Screen passwords against known-compromised values

### LINKED ISSUE #855: Reject known-compromised passwords at registration and password change
## Description

`app/core/security.py::validate_password_strength` only enforces structural rules — length, upper, lower, digit, symbol. That check happily accepts `Password1!`, `Welcome123!`, `Qwerty123!` and similar passwords that sit at the very top of every credential-stuffing wordlist.

Structural complexity rules are a weak signal on their own. NIST SP 800-63B explicitly recommends screening candidate passwords against a list of known-compromised values instead of relying on composition rules.

We already track `password_history` and have `httpx` available, so the plumbing is mostly there.

## Proposed solution

Two complementary layers:

1. **Local blocklist** — an in-repo list of the most common leaked passwords, checked case-insensitively with basic normalisation (leetspeak, trailing digits) so `P@ssw0rd1` is caught too. Also reject passwords that contain the user's email local-part or username.
2. **Have I Been Pwned range API** — k-anonymity lookup: SHA-1 the password, send only the first 5 hex characters, and match the suffix locally. The full password never leaves the process. Must fail open (network error must not lock users out of signup) and be disableable via settings for offline/CI environments.

## Acceptance criteria

- [ ] Common/leaked passwords rejected with a clear, non-leaky error message.
- [ ] Passwords containing the username or email local-part rejected.
- [ ] HIBP checked via k-anonymity; only the 5-char SHA-1 prefix is ever sent.
- [ ] Network failure or timeout fails open and is logged.
- [ ] Feature toggleable via config (default off in tests).
- [ ] Applied to signup, password change, and password reset.
- [ ] Tests, including a mocked HIBP response and a network-failure case.

MOHITKOURAV01:
# Pull Request

## Summary

`validate_password` only checked structure — length, upper, lower, digit, symbol. `Password1!`, `Welcome123!` and `Qwerty123!` all pass that, and all three sit near the top of every credential-stuffing wordlist.

NIST SP 800-63B is explicit that composition rules are the weak option and that screening candidates against known-compromised values is what actually helps. This PR adds that screening in three layers.

---

## Related Issue

Closes #855

---

## Type of Change

- [ ] Bug fix
- [x] New feature
- [x] Documentation update
- [ ] Performance improvement
- [ ] Refactoring
- [ ] UI/UX enhancement
- [x] Tests
- [ ] Other

---

### 1. Local blocklist — `app/core/password_blocklist.py`

An in-repo list of the most commonly observed passwords. Candidates are normalised before comparison, so one entry covers a family of decorated variants:

| Input           | Normalised | Rejected |
| :-------------- | :--------- | :------- |
| `password`      | `password` | yes      |
| `P@ssw0rd`      | `password` | yes      |
| `Password2025!` | `password` | yes      |
| `Welcome123!`   | `welcome`  | yes      |

Normalisation lowercases, applies NFKD folding, strips trailing digits and punctuation, then undoes leetspeak. **Order matters and I got it wrong on the first pass** — folding leetspeak first turns the `2025` in `Password2025!` into letters, leaving nothing for the trailing-noise pattern to strip, so it never matched. The tests caught it; the ordering constraint is now commented in the function.

Because normalisation collapses variants, the list stays small enough to read and review rather than being a multi-megabyte wordlist vendored into the repo. HIBP covers the long tail.

### 2. Personal information

A password containing the user's own username or email local-part is rejected. Tokens under four characters are ignored (a two-letter username would match almost anything) and the email domain is not considered, since every user shares it.

### 3. Have I Been Pwned — `app/services/password_breach_service.py`

k-anonymity range lookup: SHA-1 the password, send **only the first five hex characters**, match the suffix locally. The password never leaves the process and neither does its full hash. `Add-Padding: true` is set so response size leaks nothing. Range responses cache for 24h.

**Fails open.** Timeout, connection error, 5xx and malformed responses are all logged and treated as "no reason to reject". An outage at a third party must not become an outage of our signup form.

### Wiring

`register`, `change_password` and `reset_password` now pass the account's username and email so the personal-information check has context. `reset_password` resolves the user *before* validating for that reason — the token checks that follow are unchanged.

### Test fixtures

Fourteen test files used `Passw0rd!` or `Password123!` as the fixture password. Both are now rejected, so they are updated to a value that is not on a wordlist. This is the change touching the most files, but it is mechanical.

---

## Testing

- [x] Tested locally
- [x] Existing tests pass
- [x] Added new tests
- [ ] UI verified

Additional testing notes:

`tests/test_password_screening.py` — 48 passed. The HIBP transport is mocked throughout; the suite must not depend on a third party being reachable. Coverage includes:

- Normalisation folding, and the guard that an entry never normalises down to `""` (which would reject every password whose core folds away).
- Blocklist hits and misses, including the composition-rule survivors.
- Username/email matching, dotted usernames, the domain-only non-match, and the short-token floor.
- **Privacy**: an explicit assertion that neither the password nor the non-prefix part of its digest appears in the outbound URL.
- Padding is requested.
- Breach counts, the minimum-count threshold, and cache reuse.
- Timeout, connection error, 5xx and malformed-count paths all fail open.
- End-to-end through `validate_password`, including that a HIBP outage does not block registration and that error messages never echo the password.

Regression check on the files whose fixtures changed, compared against a clean `main` worktree:

| File | main | this branch |
| :--- | :--- | :--- |
| `test_auth` | 14 passed | 14 passed |
| `test_users` | 20 passed | 20 passed |
| `test_mfa` | 6 passed | 6 passed |
| `test_refresh_tokens` | 6 passed | 6 passed |
| `test_last_seen` | 3 passed | 3 passed |
| `test_oauth` | 3 passed | 3 passed |
| `test_rbac` | 20 failed, 27 passed | 20 failed, 27 passed |
| `test_organizations` | 9 failed | 9 failed |
| `test_conversations` | 1 failed, 4 passed | 1 failed, 4 passed |
| `test_profile_completion` | 1 failed, 4 passed | 1 failed, 4 passed |
| `test_notifications_events` | 1 failed, 2 passed | 1 failed, 2 passed |
| `test_notification_tasks` | 1 failed, 2 passed | 1 failed, 2 passed |

Identical on both sides — those failures are pre-existing and unrelated (they post to `/organizations/` rather than `/api/organizations/`). No regressions from this change.

---

## Checklist

---

## Additional Notes

**Existing users are unaffected.** Screening runs on password *entry*, not on login, so nobody is locked out of an account whose password is now on the list. Prompting existing users to rotate a compromised password is a separate, larger piece of work.

Both layers can be turned off independently (`ENABLE_PASSWORD_BLOCKLIST`, `ENABLE_HIBP_CHECK`) — see the new section in `docs/security.md` for the full settings table.

### DIFF
--- a/backend/app/core/config.py
+++ b/backend/app/core/config.py
@@ -54,6 +54,33 @@ class Settings(BaseSettings):
 
     PASSWORD_HASH_SCHEME: str = "bcrypt"
 
+    # ----------------------------------------------------------
+    # Password screening
+    # ----------------------------------------------------------
+
+    # Offline blocklist of common/guessable passwords, plus the check that a
+    # password does not simply repeat the user's own username or email.
+    ENABLE_PASSWORD_BLOCKLIST: bool = True
+
+    # Have I Been Pwned range lookup. Off during tests so the suite never
+    # touches the network; the behaviour itself is covered with a mocked
+    # transport.
+    ENABLE_HIBP_CHECK: bool = Field(
+        default_factory=lambda: not (
+            os.getenv("TESTING") == "true" or "pytest" in sys.modules
+        )
+    )
+    HIBP_API_URL: str = "https://api.pwnedpasswords.com/range"
+    HIBP_TIMEOUT_SECONDS: float = 3.0
+
+    # A password is rejected once it has been seen at least this many times.
+    # A count of 1 is often a corpus artefact rather than a password in active
+    # circulation on stuffing lists.
+    HIBP_MIN_BREACH_COUNT: int = 5
+
+    # Range responses are effectively static, so they cache well.
+    HIBP_CACHE_TTL_SECONDS: int = 60 * 60 * 24
+
     # ==========================================================
     # Database
     # ==========================================================
--- a/backend/app/core/password_blocklist.py
+++ b/backend/app/core/password_blocklist.py
@@ -1,0 +1,243 @@
+"""
+Local screening for obviously-guessable passwords.
+
+Composition rules (upper + lower + digit + symbol) are a weak signal on their
+own: ``Password1!``, ``Welcome123!`` and ``Qwerty123!`` all satisfy them and all
+sit near the top of every credential-stuffing wordlist. NIST SP 800-63B
+recommends screening candidates against known-bad values instead of leaning on
+composition alone.
+
+This module is the offline half of that. The network half -- the Have I Been
+Pwned range API -- lives in ``app/services/password_breach_service.py``.
+"""
+
+import re
+import unicodedata
+from typing import Iterable, Optional, Set
+
+# The most frequently observed passwords across public breach corpora, plus the
+# ones this project's own domain invites ("devlink", "github", ...). Entries are
+# stored already normalised (see ``normalise_password``) so lookup is a single
+# set membership test.
+#
+# Deliberately kept small and in-repo rather than shipping a multi-megabyte
+# wordlist: HIBP covers the long tail, and this list only needs to catch the
+# passwords people actually reach for first, including when HIBP is unreachable.
+_RAW_COMMON_PASSWORDS: tuple[str, ...] = (
+    # Classic top-25
+    "123456",
+    "password",
+    "123456789",
+    "12345678",
+    "12345",
+    "1234567",
+    "1234567890",
+    "qwerty",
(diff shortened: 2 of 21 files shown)


MOHITKOURAV01:

| :--- | :--- | :--- | :--- |

Issue #582, opened 2026-07-27
Plugin & Extension System


nensii21:
Design a plugin architecture that allows third-party developers to extend DevLink with custom integrations, widgets, and workflows through a documented extension API.

nayanraj864-cmyk:
i want to work on this assign to me @nensii21

nayanraj864-cmyk:
i working on the same project but other issue i need some time

PR #861, opened 2026-08-03
Add a runtime feature flag service

### LINKED ISSUE #856: Introduce a runtime feature flag service
## Description

DevLink ships a large number of half-finished and experimental surfaces (the graph view, hackathons, the AI recommendation panel, the design-system route, maintenance mode). Right now the only way to turn any of them on or off is to edit the router and redeploy.

There is no feature flag mechanism anywhere in the codebase — `grep -ri feature_flag backend/app` returns nothing. That means:

- Incomplete work has to be merged behind a comment or not merged at all.
- We cannot do a percentage rollout of anything risky.
- We cannot kill a misbehaving feature without shipping a release.

Note this is deliberately *not* the same as #582 (Plugin & Extension System) — that issue is about third-party extensibility, this one is about first-party runtime toggles.

## Proposed solution

A small, dependency-free flag service:

- Flags defined in config with a default state, so the system works with zero external infrastructure.
- Per-flag rollout strategies: `on`, `off`, `percentage` (stable hash of user id, so a user's bucket never flickers), and `allowlist` of user ids.
- Overrides persisted in Redis via the existing `cache_manager`, so an operator can flip a flag without a deploy.
- A `GET /api/feature-flags` endpoint returning the evaluated flag set for the current user, so the frontend can gate UI from one request.
- A FastAPI dependency (`require_flag("...")`) that 404s a route when its flag is off.

## Acceptance criteria

- [ ] Flags evaluate from config defaults with no Redis present.
- [ ] Percentage rollout is deterministic and stable per user.
- [ ] Allowlisted users always get the flag regardless of percentage.
- [ ] Anonymous users resolve to the default state.
- [ ] Redis override takes precedence over config.
- [ ] `GET /api/feature-flags` returns the evaluated map for the caller.
- [ ] `require_flag` dependency blocks disabled routes.
- [ ] Unit tests for each strategy.

MOHITKOURAV01:
# Pull Request

## Summary

Adds runtime feature flags: toggles that can be flipped without a deploy, targeted at a subset of users, and rolled out gradually.

The `ENABLE_*` booleans already in `config.py` are build-time switches — changing one means editing the environment and restarting. They are the right tool for "does this deployment have email configured". They are the wrong tool for "is the graph view ready for everyone yet", and that second question is the one we keep hitting: several half-finished surfaces (graph view, hackathons, the AI recommendation panel, the design-system route) are currently gated by nothing at all, so the choice is merge them live or don't merge them.

> Deliberately **not** #582. That issue is about third-party extensibility; this is first-party runtime toggling.

---

## Related Issue

Closes #856

---

## Type of Change

- [ ] Bug fix
- [x] New feature
- [x] Documentation update
- [ ] Performance improvement
- [ ] Refactoring
- [ ] UI/UX enhancement
- [x] Tests
- [ ] Other

---

## Changes Made

- **`app/core/feature_flags.py`** — the registry and the evaluation function. Evaluation is pure: definition + optional override + user id in, boolean out, no I/O. That is what makes the strategy rules testable without standing anything up.
- **`app/services/feature_flag_service.py`** — joins definitions with Redis overrides.
- **`app/routers/feature_flags.py`** — `GET /api/feature-flags` plus admin override endpoints.
- **`app/dependencies.py`** — `require_flag()`.
- **`docs/feature_flags.md`** — how to define, use, roll out, and retire a flag.

### Strategies

`on`, `off`, `percentage`, `allowlist`. New flags are defined `off` — a flag somebody forgot to configure should stay dark.

### Percentage bucketing

Users are bucketed on `sha256("<flag_key>:<user_id>")`, which buys three things:

- **Stable** — a user's bucket never changes, so a feature cannot flicker on and off between requests. This is why Python's built-in `hash()` is unusable: it is salted per process, so two uvicorn workers would disagree about the same user.
- **Independent per flag** — the flag key is in the hash, so the same unlucky cohort is not enrolled in every rollout.
- **Monotonic** — raising the percentage only ever *adds* users. Widening a rollout never takes a feature away from someone who already had it. There is a test for this specifically, because getting it wrong is the kind of bug you only notice from a support ticket.

Allowlisted users bypass the percentage entirely, so the team can dogfood something sitting at 0%.

### `require_flag` returns 404, not 403

A 403 confirms the route exists and merely refuses you, which leaks the shape of unreleased work. A 404 is indistinguishable from the route not being deployed yet.

### Failure behaviour

Every path fails closed and stays up:

| Situation | Behaviour |
| :--- | :--- |
| No Redis | Flags resolve to their definitions |
| Override write with no backend | `503` — better than reporting a success that evaporates |
| Unknown flag key | `False` + warning; a typo leaves a feature dark, it does not break the endpoint |
| Malformed / unknown-strategy / bad-type override | Discarded with a warning, definition used |
| Cache raises | Caught, definition used |

Overrides expire after seven days — long enough to be effectively permanent for an incident kill switch, short enough that forgotten experiments do not linger.

---

## Testing

- [x] Tested locally
- [x] Existing tests pass
- [x] Added new tests
- [ ] UI verified

Additional testing notes:

`tests/test_feature_flags.py` — 40 passed. Covers:

- Registry consistency and the percentage range guard.
- Bucketing: stability, range, independence across flags, and an even-ish spread over 1000 synthetic users (with generous slack so it cannot flake).
- Every strategy, including the anonymous-caller cases and the allowlist bypass at 0%.
- **Monotonicity** — the set enabled at 25% is a subset of the set enabled at 75%.
- Overrides: forcing on, killing, partial override falling through to the definition, and that applying one does not mutate the definition.
- Service-level: no cache, unknown key, string vs dict override shapes, malformed JSON, unknown strategy, bad percentage type, cache raising, and override write failure.
- `require_flag` returning 404 when off, passing when on, and evaluating against the calling user.

Endpoint smoke test against the real app:

```
GET /api/feature-flags        -> 200 {"flags": {"graph_view": false, ...}}
GET /api/admin/feature-flags  -> 401 (unauthenticated)
```

No existing tests touch this code.

---

## Checklist

---

## Additional Notes

Nothing existing is placed behind a flag in this PR. The five seeded definitions describe surfaces that *should* be gated, but actually gating them changes their behaviour for users and belongs in separate, reviewable PRs — one per surface. This one is just the mechanism.

The docs end with a note on retiring flags. A registry full of permanently-on flags is dead branching, so removing them once a feature has fully landed should be part of the routine.


MOHITKOURAV01:

| :--- | :--- | :--- | :--- |

nensii21:
@MOHITKOURAV01 resolve conflicts so i can merge.

PR #862, opened 2026-08-03
Add PWA support: manifest, service worker, and offline fallback

### LINKED ISSUE #857: Add PWA support: web app manifest, service worker, and offline fallback
## Description

The frontend has no offline story at all. `frontend/public/` is empty, there is no web app manifest, and there is no service worker. Consequences:

- The app cannot be installed to a phone home screen or a desktop dock.
- A dropped connection produces the browser's default "no internet" error page rather than anything DevLink-branded.
- Static assets are re-fetched on every cold load even though they are content-hashed and immutable.
- There is no in-app signal that the user has gone offline, so failed mutations look like random errors.

For a collaboration tool that developers keep open in a background tab all day, losing the tab's state to a flaky network is a real annoyance.

## Proposed solution

Add a minimal PWA layer — no new runtime dependency, hand-rolled service worker:

- `public/manifest.webmanifest` with name, theme colours matching the existing dark/light tokens, and icons.
- A service worker with two strategies: cache-first for hashed build assets, network-first with cache fallback for navigations.
- An offline fallback page served when a navigation fails and nothing is cached.
- Registration helper that only registers in production builds and handles the `updatefound` lifecycle so a new deploy does not leave users on a stale shell.
- A `useOnlineStatus` hook plus a small banner so the UI can say "You're offline" instead of failing silently.

Explicitly out of scope: background sync and push notifications.

## Acceptance criteria

- [ ] Manifest linked from the document head; app is installable (Lighthouse PWA check).
- [ ] Service worker registers in production only, never in dev or test.
- [ ] Old caches purged on `activate` so deploys do not leak storage.
- [ ] Navigation falls back to an offline page when the network is unavailable.
- [ ] API requests are never cached by the service worker.
- [ ] Offline banner appears/disappears on `online`/`offline` events.
- [ ] Tests for the hook and the registration helper.

MOHITKOURAV01:
# Pull Request

## Summary

`frontend/public/` was empty — no web app manifest, no service worker, no offline story at all. That meant:

- The app could not be installed to a home screen or dock.
- A dropped connection produced the browser's default "no internet" page.
- Failed mutations looked like random errors, with nothing telling the user their network was down.

This adds a small PWA layer that fixes those three things. It is **not** an offline-first rewrite, and deliberately so.

---

## Related Issue

Closes #857

---

## Type of Change

- [ ] Bug fix
- [x] New feature
- [x] Documentation update
- [x] Performance improvement
- [ ] Refactoring
- [x] UI/UX enhancement
- [x] Tests
- [ ] Other

---

## Changes Made

| File | Purpose |
| :--- | :--- |
| `public/manifest.webmanifest` | Name, theme colours matching the existing tokens, icons, shortcuts |
| `public/sw.js` | The service worker |
| `public/offline.html` | Fallback when a navigation fails |
| `public/icons/*.svg` | App icon + maskable variant |
| `src/lib/pwa.ts` | Registration and update lifecycle |
| `src/hooks/useOnlineStatus.ts` | Connectivity hook |
| `src/components/OfflineBanner.tsx` | The banner |
| `src/routes/__root.tsx` | Manifest link, theme-color, registration, banner |
| `docs/pwa.md` | How it works and how to test it |

### Why hand-rolled

No new runtime dependency. What we need is ~160 lines; the generator plugins bring a build plugin, a config file, and a set of defaults nobody on the team can explain when something caches wrong at 2am.

### Caching

Same-origin `GET` only. Everything else goes straight to the network.

- **Cache-first** for `/_build/`, `/assets/`, `/icons/` — content-hashed, so a hit can never be stale.
- **Network-first** for navigations, falling back to the cached shell then `offline.html`. Not cache-first: a stale shell paired with a newer API is a worse failure than a slightly slower navigation.
- **Never** for `/api/` and `/uploads/` — user-scoped and constantly changing, and caching an authenticated response risks handing it to the wrong user on a shared device.

Page cache capped at 40 entries, oldest evicted first. `activate` deletes every non-current cache so deploys don't leak storage.

### Two guards worth flagging in review

1. `watchForUpdates` checks `navigator.serviceWorker.controller` before announcing an update. `controller` is null on a first install, so without this **every first-time visitor gets told to reload**.
2. `applyServiceWorkerUpdate` guards its `controllerchange` handler with a flag. Chrome can fire it more than once, and an unguarded reload there is an infinite reload loop.

### A bug the tests caught

I originally used `"serviceWorker" in navigator` for feature detection. That passes when the key *exists holding `undefined`* — jsdom, some embedded webviews, anything that has shimmed it — and then throws on first property access. Switched to a truthiness check via a `getServiceWorkerContainer()` helper. The test asserting a no-support environment is what surfaced it.

---

## Testing

- [x] Tested locally
- [x] Existing tests pass
- [x] Added new tests
- [x] UI verified

Additional testing notes:

**New tests — 25 passed** (`src/lib/__tests__/pwa.test.ts`, `src/hooks/__tests__/useOnlineStatus.test.ts`):

- Registration is skipped outside production, forced when asked, scoped to `/`, and returns `null` rather than throwing on failure.
- Update detection: already-waiting worker, worker installing while the tab is open, **no notification on first install**, no notification while still installing, and an `updatefound` with no installing worker.
- `applyServiceWorkerUpdate` posts `SKIP_WAITING`, no-ops with nothing waiting, and **reloads exactly once across repeated `controllerchange` events**.
- `unregisterServiceWorkers` handles no support and clears registrations plus caches.
- `useOnlineStatus`: both states, `undefined` treated as online, both event transitions, listener cleanup, and no update after unmount.

**Full suite:** 296 → 321 passed. The delta is exactly the 25 new tests; nothing else moved. (Four test *files* fail on both sides — Playwright e2e specs that vitest picks up. Pre-existing and unrelated.)

**Typecheck:** 37 errors before, 37 after — all pre-existing, none in the files this PR touches.

**Lint:** clean (`eslint` + `prettier`) on every new and modified file.

**Manual, against `npm run build && npm run preview`:** manifest reports installable, worker registers and activates, DevTools offline reload serves the cached shell, and an unvisited route serves `offline.html`.

---

## Checklist

---

## Additional Notes

**Out of scope on purpose:** background sync and push notifications. Both need a server-side story — VAPID keys, a subscription store, a delivery pipeline — and belong with the notification work rather than bolted on here.

Icons are SVG. Chrome accepts `sizes: "any"` SVG icons for installability, and a vector avoids checking a pile of PNG raster sizes into the repo. If we later want the widest possible install-prompt support, generating 192/512 PNGs from these SVGs is a small follow-up.

`CACHE_VERSION` in `sw.js` is the kill switch — bumping it invalidates every cache on the next deploy.

### DIFF
--- a/docs/pwa.md
+++ b/docs/pwa.md
@@ -1,0 +1,104 @@
+# Progressive Web App
+
+DevLink ships a small PWA layer: a web app manifest, a service worker, an
+offline fallback page, and an in-app offline indicator.
+
+The goal is modest and deliberately so. This is not an offline-first rewrite.
+It makes the app installable, stops a dropped connection producing the
+browser's default error page, and tells the user when they are offline instead
+of letting mutations fail silently.
+
+## What's included
+
+| File                              | Purpose                                          |
+| :-------------------------------- | :----------------------------------------------- |
+| `public/manifest.webmanifest`     | Name, colours, icons, app shortcuts               |
+| `public/sw.js`                    | The service worker                                |
+| `public/offline.html`             | Fallback page when a navigation fails             |
+| `public/icons/icon.svg`           | App icon                                          |
+| `public/icons/icon-maskable.svg`  | Maskable variant, scaled for the 80% safe zone    |
+| `src/lib/pwa.ts`                  | Registration and update handling                  |
+| `src/hooks/useOnlineStatus.ts`    | Connectivity hook                                 |
+| `src/components/OfflineBanner.tsx`| The banner                                        |
+
+## Caching strategy
+
+The worker only touches same-origin `GET` requests. Everything else goes
+straight to the network.
+
+**Cache-first — `/_build/`, `/assets/`, `/icons/`**
+These URLs are content-hashed, so a cache hit can never be stale. A new deploy
+produces new filenames.
+
+**Network-first — navigations**
+Try the network, fall back to the cached shell, then to `offline.html`. Not
+cache-first: a stale app shell paired with a newer API is a worse failure than
+a slightly slower navigation.
+
+**Never cached — `/api/`, `/uploads/`**
+API responses are user-scoped and change constantly; serving a stale one is
+worse than failing the request. Caching an authenticated response also risks
+handing it to the wrong user on a shared device.
+
+The page cache is capped at 40 entries, evicting oldest-first.
+
+## Updates
+
+When a deploy lands while a tab is open, the new worker installs and waits.
+`src/lib/pwa.ts` detects that and the app shows a toast with a **Reload**
+action, which posts `SKIP_WAITING` to the waiting worker and reloads once it
+takes over.
+
+Users are never force-reloaded mid-task. The `controllerchange` handler is
+guarded by a flag, because Chrome can fire it more than once and a reload loop
+is a spectacularly bad failure mode.
+
+`activate` deletes every cache that is not the current version, so deploys do
+not leak storage. Bumping `CACHE_VERSION` in `sw.js` invalidates everything.
+
+## Registration
+
+Only in **production browser builds**:
+
+```ts
+registerServiceWorker({ onUpdateAvailable: (registration) => { ... } });
+```
+
+Not in dev — a service worker there is a reliable way to spend an afternoon
+debugging a stale bundle — and not under Vitest. Pass `force: true` when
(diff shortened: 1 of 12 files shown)


MOHITKOURAV01:

| :--- | :--- | :--- | :--- |

Issue #574, opened 2026-07-27
Build User Settings Page


nensii21:
Description

Create a centralized settings page with tabs for:

Account
Privacy
Notifications
Appearance
Security

Babin123456:
@nensii21 I want to try this.

PR #863, opened 2026-08-03
Add an internationalization (i18n) foundation to the frontend

### LINKED ISSUE #858: Add an internationalization (i18n) foundation to the frontend
## Description

Every user-facing string in the frontend is a hard-coded English literal. There is no i18n library, no locale files, and no way for a non-English speaker to use DevLink. Dates are formatted with ad-hoc `date-fns` calls and numbers with raw template strings, so even a user whose browser is set to `de-DE` sees US formatting throughout.

Retro-fitting i18n once there are thousands of literals is significantly more painful than putting the foundation in now, and a large share of this project's contributors are not native English speakers.

## Proposed solution

A lightweight, dependency-free i18n foundation rather than pulling in `react-i18next` (which is a sizeable addition for what we need today):

- A typed translation catalogue keyed by dot-path, with English as the source of truth.
- `I18nProvider` + `useTranslation()` returning a `t()` that supports `{placeholder}` interpolation and simple pluralisation via `Intl.PluralRules`.
- Locale detection order: explicit user choice in `localStorage` → `navigator.language` → `en`.
- `Intl`-backed helpers for dates, relative times, numbers, and percentages so formatting follows the active locale.
- Missing keys fall back to English and warn in dev, never crash and never render a raw key in production.
- A language switcher component, and one real screen migrated end-to-end to prove the pattern.

Adding further locales should then be a matter of dropping in one file.

## Acceptance criteria

- [ ] `t()` resolves nested keys and interpolates named placeholders.
- [ ] Pluralisation follows the active locale's rules.
- [ ] Unknown key falls back to English, then to a readable last segment.
- [ ] Locale persists across reloads.
- [ ] `<html lang>` updates when the locale changes.
- [ ] Date/number/relative-time helpers respect the active locale.
- [ ] At least one non-English catalogue included as a reference.
- [ ] Unit tests for resolution, interpolation, pluralisation, and fallback.

MOHITKOURAV01:
# Pull Request

## Summary

Every user-facing string in the frontend is a hard-coded English literal. There is no i18n library, no locale files, and dates and numbers are formatted with US conventions regardless of the browser's locale — a user on `de-DE` sees `1,234.5` where they expect `1.234,5`.

This adds the foundation, plus one screen migrated end-to-end to prove the pattern.

Retro-fitting i18n once there are thousands of literals is dramatically more painful than putting the plumbing in now, and a large share of this project's contributors are not native English speakers.

---

## Related Issue

Closes #858

---

## Type of Change

- [ ] Bug fix
- [x] New feature
- [x] Documentation update
- [ ] Performance improvement
- [x] Refactoring
- [x] UI/UX enhancement
- [x] Tests
- [ ] Other

---

## Changes Made

| Path | Purpose |
| :--- | :--- |
| `src/i18n/types.ts` | `Catalogue`, `PluralForms`, option types |
| `src/i18n/translate.ts` | Resolution, interpolation, pluralisation (pure) |
| `src/i18n/format.ts` | `Intl` wrappers for dates, relative times, numbers |
| `src/i18n/locales/en.ts` | English — the source of truth |
| `src/i18n/locales/es.ts` | Spanish — worked reference |
| `src/i18n/locales/index.ts` | Registry and BCP-47 tag resolution |
| `src/context/I18nContext.tsx` | `I18nProvider`, `useTranslation()` |
| `src/components/LanguageSwitcher.tsx` | The picker |
| `src/components/errors/*` | Migrated end-to-end |
| `docs/i18n.md` | How to add strings and languages |

### Why not `react-i18next`

What we need today is key lookup, `{placeholder}` interpolation, plural selection and `Intl` formatting — a few hundred lines. `react-i18next` is a sizeable dependency with a plugin system, backend adapters and an initialisation dance we would not use.

If requirements grow past this (ICU syntax, lazy namespaces, a TMS integration), swapping a library in behind the same `useTranslation()` surface is contained. Starting *with* the library is the harder thing to reverse.

### Flat keys, not nested

Nesting reads nicely in the file but makes the key type a recursive conditional — slow to check and unreadable in an editor tooltip. `keyof typeof en` on a flat object is exact and instant, so `t("typo.key")` is a compile error and autocomplete works.

---

## Two things the tests caught

**1. The `zero` plural form silently never fired.**

English has no `zero` CLDR category — `Intl.PluralRules("en").select(0)` returns `other`. Following that strictly meant a translator who wrote `zero: "No unread notifications"` never saw it and got "0 unread notifications" instead.

Writing a `zero` form is an unambiguous statement of intent, so it now wins at `count === 0` in every locale. This is a deliberate, documented deviation from CLDR — everything else follows `Intl.PluralRules` exactly. Flagging it explicitly since it's the one place this diverges from the spec.

**2. The error pages render outside the providers.**

The root route's `errorComponent` and `notFoundComponent` **replace** `RootComponent` rather than rendering inside it. Since the migrated error pages call `useTranslation()`, an API 401 would have surfaced as *"useTranslation must be used within an I18nProvider"* instead of the sign-in page. `__root.tsx` now wraps them via `OutsideRootProviders`.

This is exactly why the PR migrates a real screen rather than only shipping the mechanism — the wiring problem was invisible until something actually consumed it.

---

## Design notes

- **Fallback chain:** active catalogue → English → a humanised key (`projects.emptyState.title` → "Title"). `t()` never throws and never returns `""` — a translation problem should degrade the copy, not blank the UI.
- **Unknown placeholders stay visible.** A stray `{userName}` on screen is an obvious bug report; the string "undefined" is a confusing one.
- **`useTranslation()` throws outside a provider** rather than quietly defaulting to English, which would hide the wiring bug until a non-English user reported it.
- **`formatPercent` takes a 0–1 ratio**, not an already-multiplied number, so callers cannot disagree about whether `50` means 50% or 5000%.
- **Formatters are cached** — `Intl` construction is expensive relative to how often we format — and every one degrades to `en` on a malformed locale tag, so a bad `localStorage` value cannot blank a page.
- `<html lang>` follows the active locale, which screen readers use to pick a voice.

---

## Testing

- [x] Tested locally
- [x] Existing tests pass
- [x] Added new tests
- [x] UI verified

Additional testing notes:

**New tests — 73 passed** across three files:

- `translate.test.ts` — interpolation (repeats, whitespace, numbers, unknown and `undefined` values), plural selection including the Japanese `other`-only case and a malformed tag, key humanisation, the full fallback chain, and the plural-without-count warning.
- `format.test.ts` — locale-dependent grouping/decimals, compact notation, relative time in English and Spanish, the `numeric: "auto"` "yesterday" case, unit selection, and the non-finite / unparseable guards.
- `I18nContext.test.tsx` — detection precedence, walking `navigator.languages` past unsupported entries, rejecting an unsupported stored value, persistence, surviving a throwing `localStorage` (Safari private mode), `<html lang>` updates, live locale switching, English fallback for a key Spanish lacks, and the throw-outside-provider case.

Three of these are **catalogue invariants** rather than unit tests, and will keep future locales honest: every non-English key must exist in English, every plural entry must define `other`, and a translated string's `{placeholders}` must match the English original.

**Full suite:** 296 → 369 passed. The delta is exactly the 73 new tests. (Four test *files* fail on both sides — Playwright e2e specs vitest picks up. Pre-existing and unrelated.)

**Typecheck:** 37 errors before, 37 after — all pre-existing, none in files this PR touches.

**Lint:** clean. `I18nContext.tsx` emits the same `react-refresh/only-export-components` warning that the existing `ThemeContext.tsx` does, which is inherent to a context file exporting both a provider and its hook.

---

## Checklist

---

## Merge order

> **This branch has #862 merged into it.** Both PRs wire themselves into `src/routes/__root.tsx` — the import block and the provider tree — so they collided. Both hunks were purely additive and the resolution keeps everything from each side:
>
> ```tsx
> <I18nProvider>
>   <ThemeProvider defaultTheme="system">
>     <OfflineBanner />
>     ...
> ```
>
> `I18nProvider` sits outside `ThemeProvider` so anything under the theme can translate. `OfflineBanner` does not consume i18n today, but that is where it will need to be once its copy is migrated.
>
> **Please merge #862 first.** Once it lands, this PR's diff collapses to only the i18n changes. Verified in a combined worktree with all five PRs merged: 394 tests pass, typecheck and lint unchanged from baseline.

---

## Additional Notes

**No visible change for English users.** The migrated error pages render exactly the copy they rendered before; the strings just come from a catalogue now.

Spanish deliberately omits `notifications.unread` so the English fallback is covered by a test rather than assumed. Machine-assisted and worth a native review before anyone relies on it.

The rest of the app still has hard-coded literals. Migrating a screen is: add its strings to `en.ts`, swap the literals for `t()`. That can happen incrementally, one screen per PR, and does not need to block this.

`LanguageSwitcher` is not mounted in the settings page yet — that touches a screen this PR otherwise leaves alone, and belongs with the settings work in #574.

### DIFF
--- a/docs/i18n.md
+++ b/docs/i18n.md
@@ -1,0 +1,159 @@
+# Internationalization
+
+Every user-facing string in the frontend used to be a hard-coded English
+literal. There was no i18n library, no locale files, and dates and numbers were
+formatted with US conventions regardless of the user's browser settings.
+
+This document covers the foundation added in #858.
+
+## Why not `react-i18next`
+
+What the app needs today is key lookup, `{placeholder}` interpolation, plural
+selection and `Intl` formatting. That is a few hundred lines. `react-i18next`
+is a sizeable dependency with its own plugin system, backend adapters and
+initialisation dance, most of which we would not use.
+
+If requirements grow past this — ICU message syntax, lazy-loaded namespaces,
+integration with a translation-management platform — swapping a library in
+behind the same `useTranslation()` surface is a contained change. Starting with
+the library would have been the harder thing to reverse.
+
+## Layout
+
+| Path                          | Purpose                                    |
+| :---------------------------- | :----------------------------------------- |
+| `src/i18n/types.ts`           | `Catalogue`, `PluralForms`, option types    |
+| `src/i18n/translate.ts`       | Resolution, interpolation, pluralisation    |
+| `src/i18n/format.ts`          | `Intl` wrappers for dates and numbers       |
+| `src/i18n/locales/en.ts`      | English — the source of truth               |
+| `src/i18n/locales/es.ts`      | Spanish — worked reference                  |
+| `src/i18n/locales/index.ts`   | Locale registry and tag resolution          |
+| `src/context/I18nContext.tsx` | `I18nProvider`, `useTranslation()`          |
+| `src/components/LanguageSwitcher.tsx` | The picker                          |
+
+## Using it
+
+```tsx
+import { useTranslation } from "@/context/I18nContext";
+
+function ProjectHeader({ memberCount, createdAt }) {
+  const { t, formatRelativeTime } = useTranslation();
+
+  return (
+    <header>
+      <p>{t("members.count", { count: memberCount })}</p>
+      <time>{formatRelativeTime(createdAt)}</time>
+    </header>
+  );
+}
+```
+
+`t()` is typed against `keyof typeof en`, so an unknown key is a compile error
+and your editor autocompletes the real ones.
+
+## Adding a string
+
+Add it to `src/i18n/locales/en.ts`. That's the whole step — other locales are
+allowed to lag behind and fall back to English.
+
+```ts
+"projects.emptyState.title": "No projects yet",
+"projects.count": { one: "{count} project", other: "{count} projects" },
+```
+
+Keys are **flat and dot-separated**, not nested. Nesting reads nicely in the
+file but turns the key type into a recursive conditional that is slow to check
+and unreadable in an editor tooltip. `keyof` on a flat object is exact and
+instant.
+
+The `as const` on the catalogue is what produces the key type. It must stay.
+
+## Pluralisation
+
+Provide a variant per CLDR category. Only `other` is required:
+
+```ts
+"notifications.unread": {
+  zero: "No unread notifications",
+  one: "{count} unread notification",
+  other: "{count} unread notifications",
+},
+```
+
+`count` selects the form *and* is available as `{count}`, so you don't pass it
+twice.
+
+**One deliberate deviation from CLDR.** English has no `zero` plural category —
+`Intl.PluralRules("en").select(0)` returns `other`. Strictly following that
+would mean a translator who wrote `zero: "No unread notifications"` never saw
+it and got "0 unread notifications" instead. Writing a `zero` form is an
+unambiguous statement of intent, so it wins at `count === 0` in every locale.
+Everything else follows `Intl.PluralRules` exactly.
+
+## Fallback chain
+
+1. Active catalogue
+2. English
+3. A humanised form of the key — `projects.emptyState.title` renders as "Title"
+
+Step 3 is not a translation, but it beats showing a raw dotted key. Missing
+keys warn once each in development and are silent in production. `t()` never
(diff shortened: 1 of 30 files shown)


MOHITKOURAV01:

| :--- | :--- | :--- | :--- |

PR #880, opened 2026-08-04
fix: resolve conflict markers committed on main and unbreak both builds

### LINKED ISSUE #875: Committed merge conflict markers on main break the backend import and the frontend build
### Summary

`main` currently contains raw Git conflict markers that were committed instead of resolved. The backend cannot be imported and the frontend cannot be type-checked or built from a clean checkout.

### Where the markers are

| File | Lines |
| --- | --- |
| `backend/app/main.py` | 610 – 649 |
| `backend/app/models/__init__.py` | 73 – 86 |
| `frontend/src/routeTree.gen.ts` | 18 separate conflicted regions |

All of them carry the same right-hand label: `>>>>>>> d54dec75 (feat(jobs): implement background job monitoring and retry support)`.

### How to reproduce

```bash
git checkout main
cd backend && python -c "import app.main"
```

```
File "backend/app/main.py", line 610
    <<<<<<< HEAD
    ^
SyntaxError: invalid syntax
```

```bash
cd frontend && npm run typecheck
```

fails the same way on `src/routeTree.gen.ts`.

### Impact

* Backend CI (`Backend CI / backend`) is red on `main` and on every PR branched from it.
* Frontend CI (`frontend / build`) is red for the same reason.
* Because both sides are already failing, a genuinely broken PR is indistinguishable from a healthy one — every contributor now has to reason about which red check is "the usual one".
* The conflict in `models/__init__.py` also means the model registry silently loses whichever side was dropped, so `Base.metadata` is incomplete at runtime.

### What the correct resolution looks like

The two sides are additive, not competing — neither branch deleted anything the other needed:

* `main.py`: keep the `custom_openapi()` block **and** the `background_jobs` router registration.
* `models/__init__.py`: keep the existing model imports **and** `BackgroundJob`/`JobStatus`.
* `routeTree.gen.ts`: this file is generated by TanStack Router. It should be regenerated so it lists the union of both branches' routes, including `/admin/jobs`.

### Acceptance criteria

- [ ] `grep -rn '^<<<<<<< \|^>>>>>>> ' backend frontend` returns nothing.
- [ ] `python -c "import app.main"` succeeds from `backend/`.
- [ ] `npm run typecheck` and `npm run build` succeed from `frontend/`.
- [ ] The `/admin/jobs` route and the background-job endpoints both still resolve.
- [ ] A CI guard fails the build if conflict markers are ever committed again.

MOHITKOURAV01:
Closes #875

## What was wrong

`main` did not build on either side. Three files were committed with their Git conflict markers still in place, and a fourth had an import left unterminated by a bad resolution:

| File | Damage |
| --- | --- |
| `backend/app/main.py` | markers at 610–649 |
| `backend/app/models/__init__.py` | markers at 73–86 |
| `backend/app/routers/projects.py` | unclosed `from … import (` |
| `frontend/src/routes/_app.admin.tsx` | markers at 17–30 |
| `frontend/src/routeTree.gen.ts` | 18 conflicted regions |

`backend/app/routers/projects.py` was not in the original issue because `grep` for markers does not find it — it is a *resolved* conflict that dropped a closing paren. `black` was the thing that spotted it: `Cannot parse: 18:0`.

## Resolutions

Both sides of every conflict were additive, so none of them needed a winner:

* **`main.py`** — keeps the OpenAPI shared-error-response pass **and** the `background_jobs` router registration. While in there, the response loop now skips non-operation path-item keys (a path item can hold a `parameters` list, and `method.setdefault` on a list would have thrown).
* **`models/__init__.py`** — keeps the existing model registry **and** `BackgroundJob`/`JobStatus`. Losing either side leaves `Base.metadata` incomplete.
* **`_app.admin.tsx`** — the two sides were two competing nav strips. Collapsed into one, driven by a list, so the seven admin pages that exist all have a tab. Three of them previously had no way in but typing the URL.
* **`routeTree.gen.ts`** — regenerated by the TanStack plugin rather than hand-merged.

`_app.builders.tsx` had the same class of damage in a subtler form: both sides' `lucide-react` imports were kept verbatim, so `Search` was declared twice and the module would not parse.

## The 40 type errors underneath

With the syntax errors gone, `tsc --noEmit` could run for the first time in a while. Most of what it found is **one bug wearing thirteen hats**.

Several pages call `api.get(...)` and then read `.data` off the result. That is an axios idiom, and `src/api/client.ts` is not axios — `request<T>` resolves to the parsed body. So those reads were `undefined` at runtime, not just untyped:

* the notification centre rendered an empty list regardless of what the server returned,
* the notification-preferences form always showed its hard-coded defaults instead of the user's saved settings,
* the admin stat tiles always rendered their `|| 0` fallback,
* `/admin/jobs` passed axios' `params` where the client wants `query`, so every request was unfiltered and unpaginated.

Each call site now names the response type it expects. **These are behaviour fixes, not annotations.**

The remainder:

* `filter-drawer` typed its values as `unknown` and handed them straight to DOM inputs. It now has a `FilterValue` union and three narrowing helpers.
* The projects page's filter drawer was still wired to a filter API that no longer exists — comma-joined language strings, `handleSetFilters`, `techStack`. Rewired to the current array-based search schema. The string↔boolean translation keeps "unset" distinct from "explicitly false", so an untouched **Paid** filter no longer reads as **Unpaid**.
* `ToastOptions` intersected the whole of sonner's `ExternalToast`, which made `action` an impossible `{ label; onClick } & ReactNode`. Split into an explicit passthrough of the keys we do not model ourselves.
* `OAuthAccountsSection` declared a field as `bool` — the Python spelling.
* The applications page called `useWithdrawApplication` with its import commented out.
* `useProjectFilters`' two `any`s are now the record type the router actually passes.

`eslint --fix` mops up the prettier drift that accumulated behind the broken build.

## Preventing the recurrence

* `.github/workflows/conflict-markers.yml` — a `git grep` over tracked files. Deliberately not a pre-commit-only check, since contributors are not required to install the hooks and the point is that it cannot be skipped. The pattern is anchored and requires the trailing space or EOL that Git actually writes, so a Markdown diff block does not trip it.
* `check-merge-conflict` added to `.pre-commit-config.yaml` for the faster local signal.

## Verification

```
$ git grep -nIE '^(<{7}|={7}|>{7})( |$)'      # nothing
$ python -c "import app.main"                 # backend imports OK
$ npm run typecheck                           # clean, was 40 errors
$ npm run lint                                # 0 errors (12 pre-existing warnings)
$ npm run build                               # ✓
$ npm run test                                # 394 passed
```

## Left out on purpose

`black --check .` still fails, on **234 files** of pre-existing formatting drift that has nothing to do with this conflict. Reformatting the tree is a large mechanical diff that would conflict with every open PR, so it belongs in its own change. This PR takes the backend from *"cannot parse"* to *"formatted differently"*, which is the part that was actually blocking.


nensii21:
@MOHITKOURAV01 Resolve conflicts.

PR #881, opened 2026-08-04
feat: add a link preview service with SSRF protection

### LINKED ISSUE #876: Add a link preview service for URLs shared across the platform
### Problem

Links are everywhere in DevLink — a repo URL in a project description, a blog post pasted into a conversation, a docs link dropped into an issue comment. Right now every one of them renders as bare text. The reader has no idea whether `https://example.com/a/b/c` is a GitHub repo, a YouTube video, or a phishing page until they click it.

Every product that handles user-pasted links eventually grows a preview card. We should build ours deliberately rather than bolting it on later, because the naive version of this feature is a **server-side request forgery hole**: an endpoint that fetches an arbitrary user-supplied URL from inside our network is an open proxy to `169.254.169.254`, to `localhost:5432`, and to anything else on the private network.

### Proposal

A `link_preview_service` that, given a URL, returns the title, description, site name, canonical URL, and image for it — sourced from Open Graph tags, falling back to Twitter card tags, falling back to plain `<title>`/`<meta name="description">`.

### The safety requirements are the point

This is not an optional hardening pass, it is the feature:

1. **Scheme allowlist.** `http` and `https` only. No `file:`, `gopher:`, `ftp:`, `data:`.
2. **DNS resolution before connecting**, then reject every resolved address that is loopback, link-local, private (RFC 1918), unique-local, multicast, or reserved — for both IPv4 and IPv6. Checking the hostname string is not enough; `127.0.0.1.nip.io` resolves to loopback.
3. **Redirects re-validated at every hop.** A public URL that 302s to `http://169.254.169.254/latest/meta-data/` must be rejected at the second hop. This means redirects have to be followed manually, not with `follow_redirects=True`.
4. **Byte cap on the response body.** Stream it and stop reading after ~512 KB so a hostile server cannot stream us gigabytes.
5. **Content-type check.** Only parse `text/html` and `application/xhtml+xml`.
6. **Short connect/read timeout,** and a bounded number of redirect hops.

### Also needed

* Cache successful previews (and negative results, for a shorter window) so a link pasted into a busy conversation is fetched once, not once per reader.
* Rate limit the endpoint per user — it is an outbound-request amplifier by nature.
* Resolve relative `og:image` paths against the final URL, and decode HTML entities in the extracted text.
* Truncate title/description to sane lengths before returning them.

### API sketch

```
GET  /api/link-previews?url=https://example.com/post   -> single preview
POST /api/link-previews/batch  { "urls": [...] }       -> up to 10 at once
```

Both authenticated. The batch form returns a per-URL status so one bad link does not fail the whole request.

### Acceptance criteria

- [ ] Open Graph, Twitter card, and bare-HTML extraction, in that precedence order.
- [ ] All six safety requirements above are implemented and unit-tested, including the redirect-to-private-IP case.
- [ ] Previews are cached; repeat requests do not re-fetch.
- [ ] The endpoint never returns a raw upstream error body to the caller.

MOHITKOURAV01:
Closes #876

## What this adds

```
GET  /api/link-previews?url=https://example.com/post   → one preview
POST /api/link-previews/batch  { "urls": [...] }       → up to 10
```

Metadata comes from Open Graph, falling back to Twitter card tags, falling back to the plain `<title>` / `<meta name="description">`. `og:url` wins for `final_url` because it is the page's own canonical spelling of itself, which is worth more than whatever redirect chain we happened to follow.

The batch endpoint always returns `200`. Each entry carries either a `preview` or an `error`, so one dead link in a message does not cost the reader every other card in it.

## The safety work is the feature

This is an endpoint whose entire job is to make our server fetch a URL the caller chose, from inside our network. Written naively it is an open proxy to `169.254.169.254`, to `localhost:5432`, and to everything else on the private subnet — with our source address and our credentials.

### `app/utils/url_safety.py`

Kept as a separate module on purpose. Webhook delivery, avatar import and repository metadata all have the same problem, and none of them should grow their own half-version of this.

* **scheme allowlist** — `http`/`https`. No `file:`, no `gopher:`, no `data:`.
* **port allowlist** — 80, 443, 8080, 8443.
* **no credentials in the URL** — the browser shows the userinfo and the server sees something else, which is a phishing vector on its own.
* **resolved-address checks** — the hostname is resolved and *every* returned address must be public.

That last point is the one that actually matters, and it is why this resolves instead of pattern-matching the hostname:

> `127.0.0.1.nip.io` is a perfectly ordinary public hostname. It resolves to loopback. There is a whole family of services that exist to do this.

Checking *all* the addresses matters for the same reason. A name with one public A record and one private AAAA record is a working attack if only the first answer is inspected — nothing in this code chooses which one the HTTP client ends up connecting to.

Refused: loopback, link-local (which is where cloud metadata lives), RFC 1918, unique-local, reserved, multicast, unspecified — for IPv4, IPv6, and IPv4-mapped IPv6, because `::ffff:127.0.0.1` is loopback wearing a hat.

### `app/services/link_preview_service.py`

* **Redirects are followed by hand**, and every hop is re-validated. This is the whole reason the fetch loop is written out longhand instead of using `follow_redirects=True` — a public URL that 302s to `http://169.254.169.254/latest/meta-data/` has to fail at the *second* hop, and the convenient version walks straight onto it.
* **The body is streamed and abandoned past 512 KiB.** Everything we want is in `<head>`, so the cap costs nothing on a real page and stops a hostile server making us buffer a gigabyte.
* Only `text/html` and `application/xhtml+xml` are parsed.
* Timeouts are short (5s) and the hop count is bounded (3).

## Caching

24 hours for a success, 5 minutes for a failure. Without it, a link pasted into a busy conversation is fetched once *per reader*. The short failure TTL means a site that was down for five minutes is not written off for a day.

The cache key is a normalised URL — lowercased scheme and host, redundant `:443` dropped, fragment dropped — so two spellings of the same page share one entry.

## Tests

47 tests, and the resolver is injected rather than real. A test that depends on what the CI runner's DNS says about `localhost` today is a test that will eventually lie.

Covered: every rejected address family individually, the mixed public/private records case, disallowed schemes and ports, credentials in the URL, cache-key normalisation, and on the parsing side — Open Graph precedence, Twitter fallback, `<title>` fallback, **attribute order** (`content="…" property="…"` is extremely common in the wild and a single fixed-order regex silently misses all of it), single-quoted attributes, duplicate tags, relative image resolution against the final URL, entity decoding, whitespace collapsing, and word-boundary truncation.

```
$ pytest tests/test_link_previews.py -q
47 passed
```

Endpoint smoke-checked against the running app: unauthenticated `GET /api/link-previews` returns `401`, so the route is wired and the auth dependency is doing its job.

## Notes for review

* Rate limits are deliberately tight — 30/min single, 10/min batch. This endpoint is an outbound-request amplifier by construction.
* No new model, no migration.
* `main.py` gains three lines, at line 516, well away from anything else in flight.
* Docs in `docs/link_previews.md`, including a short section on reusing the guard for the next feature that needs to fetch a user-supplied URL.

### DIFF
--- a/backend/app/core/config.py
+++ b/backend/app/core/config.py
@@ -287,6 +287,37 @@ class Settings(BaseSettings):
     ENABLE_APPLICATIONS: bool = True
 
     # ==========================================================
+    # Link Previews
+    # ==========================================================
+
+    # Short, because a preview is a nice-to-have sitting in front of a user
+    # waiting for a message to send. A slow site gets no card rather than a
+    # slow card.
+    LINK_PREVIEW_TIMEOUT_SECONDS: float = 5.0
+
+    # Everything we want lives in <head>. 512 KiB is generous for that and
+    # bounds what a hostile server can make us buffer.
+    LINK_PREVIEW_MAX_BYTES: int = 512 * 1024
+
+    # Each hop is re-validated against the SSRF rules, so this is a cost
+    # ceiling rather than a safety one.
+    LINK_PREVIEW_MAX_REDIRECTS: int = 3
+
+    # A day: Open Graph tags change rarely, and a stale title is a much smaller
+    # problem than re-fetching a popular link for every reader.
+    LINK_PREVIEW_CACHE_TTL_SECONDS: int = 86400
+
+    # Failures are cached far more briefly, so a site that was down for five
+    # minutes is not written off for a day.
+    LINK_PREVIEW_FAILURE_CACHE_TTL_SECONDS: int = 300
+
+    # Identifying ourselves honestly, with a contact URL, is what lets a site
+    # owner rate-limit us instead of silently blackholing us.
+    LINK_PREVIEW_USER_AGENT: str = (
+        "DevLinkBot/1.0 (+https://github.com/nensii21/devlink)"
+    )
+
+    # ==========================================================
     # Helper Properties
     # ==========================================================
 
--- a/backend/app/main.py
+++ b/backend/app/main.py
@@ -516,6 +516,9 @@ from app.routers import (
 
 
 app.include_router(media.router, prefix="/api", tags=["Media"])
+from app.routers import link_previews
+
+app.include_router(link_previews.router, prefix="/api", tags=["Link Previews"])
 app.include_router(auth.router, prefix="/api/auth", tags=["Authentication"])
 from app.routers import mfa
 app.include_router(mfa.router, prefix="/api")
--- a/backend/app/routers/link_previews.py
+++ b/backend/app/routers/link_previews.py
@@ -1,0 +1,116 @@
+"""
+Link preview endpoints.
+
+Both are authenticated and rate limited. That is not incidental: this is an
+endpoint whose whole job is to make our server fetch a URL somebody else chose,
+which makes it an outbound-request amplifier. Anonymous access would turn it
+into an open proxy, and an unlimited rate would turn it into a way to point our
+egress bandwidth at a third party.
+"""
+
+from __future__ import annotations
+
+from fastapi import APIRouter, Depends, HTTPException, Query, Request, status
+
+from app.dependencies import get_current_user
+from app.middleware.rate_limit import limiter
+from app.models.user import User
+from app.schemas.link_preview import (
+    LinkPreviewBatchItem,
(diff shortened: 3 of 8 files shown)


MOHITKOURAV01:

PR #882, opened 2026-08-04
feat: publish hackathons and milestones as a subscribable iCalendar feed

### LINKED ISSUE #877: Publish hackathons and project milestones as a subscribable iCalendar feed
### Problem

DevLink already knows about two kinds of dated commitments:

* **Hackathons** — with `start_date`, `end_date`, and a registration deadline.
* **Project milestones** — with a `due_date` and a status.

Both of them live only inside the app. If a hackathon starts at 09:00 on a Saturday, nothing puts that in the user's actual calendar, so they find out when someone messages them about it. "Add to calendar" is the single most requested affordance for any dated entity, and right now the answer is "type it in yourself".

### Proposal

Expose the user's dated items as **iCalendar (RFC 5545)** — the format Google Calendar, Apple Calendar, and Outlook all subscribe to.

Two shapes, because they solve different problems:

**1. A single-event download.** A small "Add to calendar" button next to a hackathon or milestone that returns one `VEVENT` as a `.ics` attachment. Works offline, works when the user is not signed in on their phone, no ongoing relationship with our server.

**2. A subscribable feed.** A stable, token-authenticated URL the user pastes into their calendar client once. It then polls us, and the user's hackathons and milestones stay in sync forever. This is the one people actually want.

```
GET /api/calendar/events.ics?token=<feed-token>   -> everything for that user
GET /api/calendar/hackathons/{id}.ics             -> single event
GET /api/calendar/milestones/{id}.ics             -> single event
```

The feed URL cannot use a normal session — calendar clients do not do OAuth, they do "GET this URL forever". So it needs its own opaque, revocable, per-user feed token, and it must be scoped to *only* the calendar feed. It should be regenerable, so a user who has leaked the URL can rotate it.

### The format has sharp edges

Writing iCalendar by hand is where this usually goes wrong, so the implementation needs to actually respect the spec:

* **Line folding at 75 octets**, folding on octet boundaries and not in the middle of a UTF-8 sequence.
* **Escaping** of `\`, `;`, `,` and newlines in `SUMMARY`/`DESCRIPTION` — an unescaped comma in a project name silently truncates the field in Outlook.
* **CRLF line endings**, not `\n`.
* **UTC timestamps** in `YYYYMMDDTHHMMSSZ` form, and `VALUE=DATE` for all-day entries like a milestone due date.
* **Globally unique, stable `UID`s.** If the UID changes between polls, the client creates a duplicate event instead of updating the existing one. It has to be derived from the entity type and id.
* **`SEQUENCE` incremented on edit**, or clients will ignore updates.
* A `VALARM` so people actually get reminded.

### Acceptance criteria

- [ ] `.ics` output validates and imports cleanly into Google Calendar and Apple Calendar.
- [ ] Folding, escaping, and CRLF behaviour are unit-tested against the awkward cases (long unicode summaries, commas, embedded newlines).
- [ ] UIDs are stable across regeneration of the same feed.
- [ ] The feed token is revocable and grants access to nothing but the feed.
- [ ] Correct `Content-Type: text/calendar; charset=utf-8` and `Content-Disposition` headers.

MOHITKOURAV01:
Closes #877

## What this adds

```
GET /api/calendar/feed-token                 → mint a token, get the subscribe URL
GET /api/calendar/events.ics?token=…         → the subscribable feed
GET /api/calendar/hackathons/{id}.ics        → one hackathon
GET /api/calendar/milestones/{id}.ics        → one milestone
```

The single-event endpoints send `Content-Disposition: attachment`, which is what makes a browser hand the file to the calendar application instead of rendering it as text. The feed endpoint deliberately does **not** — it is polled by a subscription, and an attachment header turns every manual visit into a download prompt.

## Why the feed needs its own credential

Calendar clients do not do OAuth. They do *"GET this URL, forever, with no headers I control"*. So the feed cannot use a session cookie or a bearer token — the secret has to live in the URL.

The token is an HMAC-SHA256 over version, user id and issue time, keyed on a value derived from `SECRET_KEY` and `CALENDAR_FEED_TOKEN_SALT`. Deriving rather than using `SECRET_KEY` directly means a feed token can never be confused with a session token even if one of the formats changes later.

It is **scoped** — a leaked feed URL exposes hackathon and milestone titles and nothing else. No messages, no profile, no session. It **expires** after a year. And it is **stateless**, so no new table and no fourteenth alembic head.

A bad token gets **404, not 401**. A 401 that distinguishes "wrong token" from "no such feed" is an oracle for guessing tokens, and a calendar client cannot act on a 401 anyway.

### The limitation, stated plainly

Because the tokens are stateless, **a single leaked token cannot be revoked on its own**. The levers are: wait for expiry, or change `CALENDAR_FEED_TOKEN_SALT` and restart, which kills every issued feed URL at once. Calling `feed-token` again issues a new token but does not invalidate the old one.

Proper per-user revocation needs something to revoke against — a `calendar_feed_tokens` table, or a nonce column on `users`. Small change, good follow-up. It is not in this PR because the repo already carries **13 alembic heads**, and adding a 14th inside a feature PR is not a trade worth making. `docs/calendar_feeds.md` says all of this in the same words, so nobody has to rediscover it from the code.

## Where the care went

Most of the work is in `app/utils/ics.py`, because iCalendar failures are not exceptions. They are:

* Outlook silently truncating a project called `Fast, cheap, good` at the first comma.
* Google creating a *duplicate* event on every poll, forever, because the UID moved.

Both only appear in a real client, days later, and neither produces an error anywhere. So the writer is fussy about four things and the tests pin every one:

**Folding at 75 octets.** The limit is octets, not characters, and a fold landing mid-UTF-8-sequence corrupts the value for every parser. The folder walks encoded lengths. It also accounts for the continuation line's leading space eating one octet of its own budget — an easy off-by-one that produces 76-octet lines. Tested with emoji straddling the boundary, plus an unfold round-trip.

**Escaping.** Backslash first, or we escape the backslashes we just introduced. `;` `,` and newlines after. URI values are **not** TEXT and are left alone — a comma in a query string is legal and escaping it breaks the link.

**CRLF.** Some parsers accept a bare `\n` and some do not; the ones that do not fail obscurely.

**Stable UIDs.** `hackathon-{id}@devlink` / `milestone-{id}@devlink`, derived from the entity, never regenerated. `SEQUENCE` is emitted when non-zero, because clients ignore an update whose sequence has not moved.

## Modelling decisions

* **Milestones are all-day events.** A due date has no meaningful time of day; rendering it at midnight puts every deadline in the small hours of the client's display. All-day uses `VALUE=DATE`, and a single-day event correctly omits `DTEND`.
* **Only registered hackathons.** A feed of every hackathon on the platform is a newsletter, not a calendar.
* **Only active, incomplete, unarchived milestones.** A done deadline in a calendar is noise.
* **A rolling 90-back / 365-forward window.** A hackathon that finished two years ago is not something anybody wants filling their calendar.
* **Alarms are asymmetric** — a day before a deadline, an hour before an event. A deadline needs warning; an event needs a nudge.
* Access checks 404 rather than 403, so an unpublished hackathon does not confirm its own existence to somebody who cannot see it.

## Tests

```
$ pytest tests/test_calendar_feed.py -q
41 passed
```

Endpoints smoke-checked against the running app: bad feed token → `404`, `feed-token` unauthenticated → `401`, single-event unauthenticated → `401`.

`black --check` clean on all new files.

## Notes for review

* No new model, no migration.
* `main.py` gains two lines at line 530.
* Docs in `docs/calendar_feeds.md`, including subscribe instructions for the three major clients and a note that they cache aggressively and mostly ignore `REFRESH-INTERVAL` — an hour or two of lag on a new event is normal, not a bug in the feed.

### DIFF
--- a/backend/app/core/config.py
+++ b/backend/app/core/config.py
@@ -258,6 +258,24 @@ class Settings(BaseSettings):
     ETAG_CACHE_CONTROL: str = "private, no-cache"
 
     # ==========================================================
+    # Calendar Feeds
+    # ==========================================================
+
+    # Mixed into the feed-token signing key. Changing it invalidates every
+    # issued feed URL at once, which is the blunt revocation lever available
+    # while tokens are stateless.
+    CALENDAR_FEED_TOKEN_SALT: str = "devlink-calendar-feed"
+
+    # A year. Long enough that a subscription is genuinely set-and-forget,
+    # short enough that a URL abandoned in a browser history stops working.
+    CALENDAR_FEED_TOKEN_MAX_AGE_DAYS: int = 365
+
+    # Advertised to clients via REFRESH-INTERVAL and X-PUBLISHED-TTL. Without
+    # a hint, clients pick their own interval, often an hour for a feed that
+    # changes daily.
+    CALENDAR_FEED_REFRESH_MINUTES: int = 360
+
+    # ==========================================================
     # Celery
     # ==========================================================
 
--- a/backend/app/main.py
+++ b/backend/app/main.py
@@ -530,6 +530,8 @@ app.include_router(project_members.router, prefix="/api", tags=["Project Members
 app.include_router(project_dashboards.router, prefix="/api", tags=["Project Dashboards"])
 from app.routers import project_milestones
 app.include_router(project_milestones.router, prefix="/api", tags=["Project Milestones"])
+from app.routers import calendar as calendar_router
+app.include_router(calendar_router.router, prefix="/api", tags=["Calendar"])
 app.include_router(analytics.router, prefix="/api", tags=["Analytics"])
 app.include_router(builder_flares.router, prefix="/api/flare", tags=["Builder's Flare"])
 app.include_router(messages.router, prefix="/api/messages", tags=["Messages"])
--- a/backend/app/routers/calendar.py
+++ b/backend/app/routers/calendar.py
@@ -1,0 +1,189 @@
+"""
+Calendar endpoints: single-event downloads and the subscribable feed.
+
+The feed endpoint is the odd one out. It authenticates with a token in the
+query string rather than a header, because a calendar client will not send one
+-- it does "GET this URL, forever" and nothing more. Everything else here is
+ordinary session-authenticated API.
+"""
+
+from __future__ import annotations
+
+import uuid
+
+from fastapi import APIRouter, Depends, HTTPException, Query, Request, status
+from fastapi.responses import Response
+from sqlalchemy.orm import Session
+
+from app.dependencies import get_current_user, get_database
+from app.middleware.rate_limit import limiter
+from app.models.hackathon import Hackathon
+from app.models.milestone import Milestone
+from app.models.user import User
+from app.schemas.calendar import CalendarFeedTokenResponse
+from app.services.calendar_feed_service import (
+    CalendarFeedService,
+    InvalidFeedToken,
+    generate_feed_token,
+    parse_feed_token,
+)
+
+router = APIRouter(
+    prefix="/calendar",
+    tags=["Calendar"],
+)
+
+CALENDAR_LIMIT = "60/minute"
+
+CALENDAR_MEDIA_TYPE = "text/calendar; charset=utf-8"
+
+
+def _ics_response(body: str, filename: str) -> Response:
+    """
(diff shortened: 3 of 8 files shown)


MOHITKOURAV01:

PR #883, opened 2026-08-04
feat: add content moderation and spam scoring for user-submitted text

### LINKED ISSUE #878: Add a content moderation and spam scoring pass for user-submitted text
### Problem

DevLink accepts free text from users in a lot of places — messages, project descriptions, conversation starters, feedback, profile bios, issue comments. None of it is screened. There is a `UserReport` model, so the current design is entirely **reactive**: something abusive gets posted, a human sees it, a human reports it, a moderator eventually acts.

That works for a small community and stops working the moment the platform is worth spamming. The two failure modes are different and both are already possible today:

* **Abuse** — slurs and harassment aimed at a specific person, which the recipient has to read before anyone can act on it.
* **Spam** — link farms, crypto pitches, "DM me for cheap followers". Repetitive, high-volume, and completely mechanical to detect.

### Proposal

A `moderation_service` that scores a piece of text and returns a **recommendation**, not a verdict:

```
allow   -> nothing interesting
flag    -> publish it, but queue it for review
review  -> hold it, a moderator decides
block   -> refuse it outright, tell the author why
```

The caller decides what to do with the recommendation. A profile bio and a direct message should not have the same threshold, so the thresholds belong to the call site, not to the service.

### Signals worth having

**Term matching that survives obfuscation.** A plain `word in text` check is defeated by `f  u  c  k`, `f.u.c.k`, `fυck` (Greek upsilon), and `𝐟𝐮𝐜𝐤` (mathematical bold). So the matcher needs to normalise first: NFKC, confusable/homoglyph folding, leetspeak folding, then collapse runs of repeated characters and strip separators. And it has to match on **word boundaries after normalisation**, or "Scunthorpe" gets blocked and "class" gets flagged for containing a slur substring.

**Spam heuristics**, each contributing to a score rather than deciding alone:

* link density relative to text length, and count of distinct domains
* URL shorteners, which have no legitimate use in a project description
* repeated character runs (`sooooo goooood`) and repeated words
* shouting ratio (uppercase letters over total letters), ignoring short text
* contact-details harvesting — phone numbers, "telegram me", email addresses in a bio
* known spam phrasings

**Context that matters more than any keyword.** A brand-new account posting five links is not the same as a two-year-old account doing it. The service should accept optional signals (account age, prior flags) and weight accordingly.

### Deliberately out of scope

No ML model, no third-party moderation API. This is the deterministic, explainable, testable layer that runs on every submission in under a millisecond. A model can sit behind it later, but a model cannot be unit-tested against "does Scunthorpe still work".

### Requirements

- [ ] Every result explains itself — which rules fired, with what weight. An unexplainable moderation decision is unappealable, and someone will have to answer a "why was my post blocked" ticket.
- [ ] Normalisation is tested against homoglyph, leetspeak, spacing, and repetition evasion.
- [ ] The false-positive cases are tested explicitly: Scunthorpe, "assignment", "class", "analysis", legitimate multi-link posts.
- [ ] Pure functions, no database access, no network. Cheap enough to call inline on every write.
- [ ] Terms live in a data module that a maintainer can edit without touching the scoring logic.

MOHITKOURAV01:
Closes #878

## What this adds

`moderation_service.check(text)` → a **recommendation**, not a verdict:

```python
result = moderation_service.check(message_body)

if result.action == Action.BLOCK:
    raise HTTPException(400, detail=result.explain())
```

`allow` / `flag` / `review` / `block`. The **caller** decides what to do with it, because a profile bio and a direct message should not share a threshold — that is a property of the call site, not of the text. Pass your own `thresholds` to move the line.

Plus `POST /api/moderation/check`, so the frontend can warn somebody *before* they hit send (much better than a rejection afterwards) and moderators can ask why a piece of text scored the way it did without reading the source.

## The hard part is not catching evasion

Catching evasion is easy. Not catching everything *else* is the part that goes wrong.

A check written as `term in text.lower()` loses to every one of these:

| | |
| --- | --- |
| `f u c k` | spacing |
| `f.u.c.k` | separators |
| `fvck` `sh1t` `a$$hole` | substitution |
| `fυck` | Greek upsilon rendering as a "u" |
| `𝐟𝐮𝐜𝐤` | mathematical bold — a different codepoint per letter |
| `fuuuuuck` | repetition |

The obvious fix is to strip everything into one blob and substring-match that. **Then:**

* `Scunthorpe` contains a swear word
* `class` contains a slur
* `assignment` contains a swear word

That is the Scunthorpe problem. Every filter that ships without a test for it eventually ships the bug — **including my first draft of this one**, which is exactly what `test_scunthorpe_is_fine` caught before this PR existed.

### The fix

`normalise()` is the **safe** fold. Accents, script homoglyphs (Cyrillic, Greek) and repeated characters are handled; separators become single spaces so **word boundaries survive**. Digits and symbols are deliberately left alone, because `123` is a number and `!` is punctuation — folding them destroys the boundaries every whole-word check depends on.

`evasion_pattern(term)` builds a per-term regex that tolerates separators and substitutions **inside** the term, while still requiring a word boundary around the **whole match**:

```
(?<![a-z0-9]) c+ [sep]{0,2} [uv]+ [sep]{0,2} n+ [sep]{0,2} t+ (?![a-z0-9])
```

`c u n t` matches. `Scunthorpe` does not — the lookbehind sees the `s`. That asymmetry is the entire design. Lookarounds rather than `\b`, because `\b` is defined against `\w`, which includes the digits and underscores evasions are built from.

`squash()` still exists for comparing two strings that should be the same word, and its docstring says in as many words that it is not safe to substring-match a term list against.

**Known limit, stated in the code and the tests:** substitutions come from a table, so an unusual one — `f4ck`, where the `4` stands for a `u` rather than the `a` it normally spells — is missed. The fix is to add the mapping, not to loosen the pattern until it starts eating ordinary words.

## The weights encode judgement, not a keyword count

* **Profanity flags, it does not block.** A developer platform that refuses *"this API is a shitshow"* is a platform nobody uses.
* **Slurs and harassment block** at 1.0.
* **Link density is relative to length**, and `ALLOWED_DOMAINS` (GitHub, MDN, PyPI, Stack Overflow…) never counts. Three links in a paragraph is a well-referenced paragraph; three links in fifteen words is a drop.
* **Contact details score higher when *solicited*.** An email in a bug report is an email. "Reach me at" plus an email is somebody routing around the platform.
* **Formatting signals are weak on purpose** — somebody is allowed to be excited. They earn their place as tiebreakers because they correlate with everything else.

## Author context

A day-old account posting five links is not the same event as a two-year-old account doing it, and no amount of keyword matching can tell those apart.

Two deliberate details:

1. **The multiplier only applies when something already fired.** Multiplying zero is pointless, and a new account posting something clean must still score zero. (`test_context_does_not_manufacture_a_score`)
2. **The account-age discount does not apply to slurs or harassment.** Seniority is not a licence. (`test_seniority_is_not_a_licence_for_slurs`)

## Explainability

```python
result.explain()
# "review @ 0.70: url_shortener (+0.40), spam_phrase (+0.30)"
```

Not decoration. A moderation decision nobody can explain is a decision nobody can appeal.

## Deliberately not a model

No ML, no third-party moderation API. This is the layer that runs inline on every write in under a millisecond, explains itself, and can be unit-tested against *"does Scunthorpe still work"* — a question no model can answer. A model can sit behind this later for what rules cannot reach; it is not the right **first** layer.

## Tests

```
$ pytest tests/test_moderation.py -q
52 passed
```

Roughly half are dedicated to *not* firing: Scunthorpe, Penistone, assignment, class, analysis, specification, constitution, a bare email in a bug report, and a technical post carrying three documentation links.

Separately spot-checked against a corpus of ordinary developer prose — *cockpit, peacock, shuttlecock, Hitchcock, Sussex, Essex, bass, analyst, cocktail* — with zero false positives.

Endpoint smoke-checked: unauthenticated `POST /api/moderation/check` → `401`. `black --check` clean on all new files.

## Notes for review

* Pure functions. No database, no network, no state, no migration.
* Terms live in `app/core/moderation_terms.py`, which is pure data — a maintainer can add one without touching scoring logic, in ordinary spelling.
* The `SLURS` list here is deliberately short and non-exhaustive; deployments extend it and the code does not change.
* `main.py` gains two lines at line 544.
* Nothing is wired into a write path yet — that is a separate change per call site, since each one needs its own threshold decision.

### DIFF
--- a/backend/app/core/moderation_terms.py
+++ b/backend/app/core/moderation_terms.py
@@ -1,0 +1,192 @@
+"""
+The word lists and patterns the moderation service scores against.
+
+Kept apart from the scoring logic on purpose: a maintainer adding a term that
+keeps showing up should not have to read, or risk breaking, the scoring code.
+Everything here is data.
+
+Terms are written in ordinary spelling. The service normalises both the term
+and the text before comparing, so there is no need to enumerate ``f4ck``,
+``fυck`` and ``f.u.c.k`` -- see ``app/utils/text_normalize.py``.
+"""
+
+from __future__ import annotations
+
+import re
+
+# ----------------------------------------------------------------------
+# Profanity
+# ----------------------------------------------------------------------
+
+# Coarse language. On its own this is a *flag*, not a block: plenty of
+# perfectly good technical discussion is sweary, and a platform for developers
+# that blocks "this API is a shitshow" is a platform nobody uses.
+PROFANITY = frozenset(
+    {
+        "arse",
+        "arsehole",
+        "asshole",
+        "bastard",
+        "bitch",
+        "bollocks",
+        "bullshit",
+        "cock",
+        "crap",
+        "cunt",
+        "damn",
+        "dick",
+        "dickhead",
+        "douchebag",
+        "fuck",
+        "fucker",
+        "fucking",
+        "motherfucker",
+        "piss",
+        "prick",
+        "shit",
+        "shite",
+        "shithead",
+        "slut",
+        "twat",
+        "wanker",
+        "whore",
+    }
+)
+
+# Language aimed at a person rather than a situation. Weighted much harder
+# than profanity, because "this is shit" and "you are worthless" are not the
+# same act even though both contain a word from a list.
+HARASSMENT = frozenset(
+    {
+        "kill yourself",
+        "kys",
+        "go die",
+        "hope you die",
+        "nobody likes you",
+        "you are worthless",
+        "youre worthless",
+        "you should die",
+        "end yourself",
+        "neck yourself",
+    }
+)
+
+# Slurs are held separately and scored at the maximum. This list is
+# deliberately short and non-exhaustive in the repository; deployments are
+# expected to extend it, and the scoring code does not need to change when
+# they do.
+SLURS = frozenset(
+    {
+        "retard",
+        "retarded",
+        "tranny",
+        "faggot",
+        "fag",
+    }
+)
+
+# ----------------------------------------------------------------------
+# Spam
+# ----------------------------------------------------------------------
+
+# Phrasings that essentially never appear in a genuine project description or
+# a message between collaborators.
+SPAM_PHRASES = frozenset(
+    {
+        "act now",
+        "buy followers",
+        "cheap followers",
+        "click here now",
+        "crypto giveaway",
+        "double your money",
+        "dm me for",
+        "earn money fast",
+        "financial freedom",
+        "forex signals",
+        "free bitcoin",
+        "free gift card",
+        "guaranteed returns",
+        "investment opportunity",
+        "limited time offer",
+        "make money online",
+        "no experience needed",
+        "risk free",
+        "telegram me",
+        "text me on whatsapp",
+        "work from home",
+        "100% free",
+    }
+)
+
+# Shorteners hide their destination, which is the entire reason they get used
+# in spam. There is no legitimate reason for one in a project description.
+URL_SHORTENERS = frozenset(
+    {
+        "bit.ly",
+        "buff.ly",
+        "cutt.ly",
+        "goo.gl",
+        "is.gd",
+        "ow.ly",
+        "rb.gy",
+        "rebrand.ly",
+        "shorturl.at",
+        "t.co",
+        "tinyurl.com",
+        "t.me",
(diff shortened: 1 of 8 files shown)


MOHITKOURAV01:

PR #885, opened 2026-08-04
feat: add shared accessibility primitives for focus management and announcements

### LINKED ISSUE #879: Build shared accessibility primitives for focus management and screen reader announcements
### Problem

The frontend has a lot of interactive surfaces — the command palette, dialogs, dropdowns, the notification centre, the sidebar — and no shared accessibility primitives. `grep -r "focus-trap\|focusTrap\|skip-to-content"` in `frontend/src` returns nothing.

Concretely, today:

* **There is no skip link.** A keyboard user landing on any page has to tab through the entire sidebar and header before reaching the content. On the dashboard that is roughly forty stops, on every single navigation.
* **Nothing traps focus in a modal.** Open a dialog, press Tab enough times, and focus walks out of the dialog and into the page behind it — which is still there, still focusable, and visually covered. The user is now typing into something they cannot see.
* **Focus is not restored.** Close a dialog and focus falls back to `<body>`, so the next Tab starts from the top of the document instead of the button that was just used.
* **Nothing is announced.** Toasts, "saved", "3 new notifications", search-result counts — all of it is invisible to a screen reader because there is no live region. `aria-live` appears in exactly three files, each hand-rolled.
* **Composite widgets use one tab stop per item.** A toolbar or tab list should be a single tab stop with arrow-key navigation between items. Ours are N tab stops.

Each of these gets re-solved, slightly differently and slightly wrong, in each component that needs it. Radix handles some of it inside its own primitives, but everything we build by hand — the command palette, the notification panel, custom menus — gets nothing.

### Proposal

A small `src/lib/a11y` layer plus the hooks and components that sit on it. No new dependency; this is a few hundred lines of well-tested DOM code.

**Tabbable element discovery.** The thing everything else needs, and the part that is genuinely fiddly. Find focusable descendants in DOM order, correctly excluding: `disabled`, `hidden`, `display:none`, `visibility:hidden`, `inert` subtrees, `tabindex="-1"`, radio buttons that are not the checked member of their group, and `<details>` content that is collapsed. Then order by positive `tabindex` first (in value order), then document order — which is what the browser actually does and what almost every hand-rolled implementation gets wrong.

**`useFocusTrap`** — confines Tab/Shift+Tab to a container, wrapping at both ends, restoring focus to the previously active element on release, with an opt-out for Escape.

**`useRovingTabIndex`** — one tab stop for a group, arrow keys to move within it, Home/End, optional wrap, optional typeahead, horizontal or vertical orientation.

**An announcer** — one polite and one assertive live region mounted once at the root, with an imperative `announce(message, urgency)`. Re-announcing identical text needs the zero-width-space trick, or screen readers stay silent on a repeat.

**`<SkipLink>`** — visually hidden until focused, jumps to `#main-content`.

**`useFocusVisible`** — so focus rings appear for keyboard users and not on mouse click, matching `:focus-visible` semantics where the CSS pseudo-class is not enough.

### Requirements

- [ ] Everything is SSR-safe. This app server-renders; touching `document` at module scope breaks the build.
- [ ] Tested with Vitest + Testing Library, including the exclusion rules above.
- [ ] Skip link wired into the app shell with a real `#main-content` target.
- [ ] No new runtime dependency.
- [ ] Documented, so the next person building a menu uses these instead of writing a sixth focus trap.

MOHITKOURAV01:
Closes #879

## What was broken

`grep -r "focus-trap\|focusTrap\|skip-to-content" frontend/src` returns nothing, and it shows:

* **No skip link.** A keyboard user landing on the dashboard tabs through roughly **forty stops** before reaching the content — on every navigation.
* **Nothing traps focus in a modal.** Tab enough times in a dialog and focus walks into the page behind it: still there, still focusable, covered by an overlay. The user is now typing into something they cannot see.
* **Focus is never restored.** Closing a dialog drops focus to `<body>`, so the next Tab starts from the top of the document instead of the button just used.
* **Nothing is announced.** Toasts, "saved", "3 new notifications", search result counts — all silent to a screen reader. `aria-live` appears in exactly three files, each hand-rolled.
* **Composite widgets are N tab stops.** A six-item toolbar costs six Tab presses to get past.

Radix handles this inside its own primitives. Everything we build by hand — the command palette, the notification panel, custom menus — gets nothing, and each one re-solves the problem slightly differently and slightly wrong.

## `lib/a11y/tabbable.ts`

What everything else stands on, and the genuinely fiddly part. `querySelectorAll("button, a[href], input, …")` gets the common case right and then goes wrong on:

| | |
| --- | --- |
| `disabled` | including inside a `disabled` fieldset — **except** anything in that fieldset's first `<legend>`, which stays interactive. That exception is real and people use it for a "customise" toggle above a disabled block. |
| `inert` subtrees | how a correctly built modal turns off the page behind it |
| `display:none` / `visibility:hidden` / `hidden` | |
| collapsed `<details>` | content excluded, `<summary>` kept |
| `tabindex="-1"` | focusable by script, skipped by Tab — the whole difference between `isFocusable` and `isTabbable` |
| **radio groups** | **one** tab stop: the checked member, or the first if none is checked. Treating each radio as its own stop makes a focus trap loop through a form's options one at a time. |
| **positive `tabindex`** | visited **before** everything in document order |

That last one is the most commonly missed. A trap that ignores it wraps in a different order than Tab does, and the user ends up somewhere they did not expect.

## `useFocusTrap`

Two things a minimal version gets wrong, both covered by tests:

**Edges are recomputed on every keypress**, not cached when the trap opens. Dialog content changes — a validation error appears, a disclosure expands, a button becomes enabled — and a cached "last element" sends focus to something no longer there.

**Focus is pulled back when it has escaped.** The user can click something behind the overlay, or a script can move focus. If the trap only handles the wrap case, that state is unrecoverable by keyboard.

Focus is restored on release, after checking the element is still in the document — it may have been removed while the dialog was open.

## `useRovingTabIndex`

One tab stop for a group, arrow keys inside it, Home/End, optional wrap and typeahead, horizontal / vertical / both.

`getItemProps` includes an `onFocus` handler, and it matters as much as the key handling: somebody can click or shift-tab straight into the middle of the group, and if the index does not follow, the next arrow press jumps back to wherever the index was left.

A horizontal group deliberately ignores `ArrowUp`/`ArrowDown` — the page still needs to scroll.

## The announcer

**Two live regions, not one.** `aria-live` is read when the region is first seen; flipping the attribute between polite and assertive on an existing region does nothing on most screen readers.

**The zero-width space.** Screen readers only re-announce when the text *changes*. Announcing "3 results", then "3 results" again after a different search, produces **silence** the second time — and the user concludes nothing happened. Appending U+200B makes it technically different while being inaudible and invisible. It is the standard workaround and the kind of thing every hand-rolled live region rediscovers the hard way.

The regions are clipped, not `hidden` and not `display: none` — either removes them from the accessibility tree, which is the exact opposite of the point.

`useAnnouncer` no-ops outside a provider, so a component using it still works in isolation and in tests that do not care.

## `SkipLink`

A bare `href="#main-content"` moves the **scroll position but not focus** unless the target is focusable — so the next Tab starts from the top of the document again and the user is back where they began. The link sets `tabindex="-1"` on the target and focuses it explicitly.

Wired into `__root.tsx` as the first thing in the tab order, targeting the `<main>` in `DashboardLayout`.

## Tests

```
$ vitest run src/lib/a11y src/hooks/__tests__/useFocusTrap* src/hooks/__tests__/useRoving* src/components/a11y
61 passed
```

Most of them pin `tabbable`'s exclusion rules **one at a time** — that is where the bugs live. Plus a full eight-press Tab cycle asserting focus never leaves an open dialog, and a test that the trap notices elements added after it opened.

Two real fixes came out of writing them:

1. `contenteditable` hosts report `tabIndex === 0` in browsers but `-1` in jsdom, whose `tabIndex` getter only knows its focusable-areas list. Reading the attribute is the portable answer, with an explicit `tabindex` still winning.
2. `scrollIntoView` is not implemented everywhere, so the skip link's call to it is optional — it is a refinement, not the point of the link.

## Notes for review

* **No new runtime dependency.**
* **SSR-safe** — nothing touches `document` at module scope, and `useFocusVisible` guards for it. This app server-renders.
* Full suite on top of #880: `typecheck` clean, `build` ✓, `455 tests passed`. On `main` alone the typecheck still fails on the committed conflict markers that #880 fixes — nothing in this PR touches those files.
* One remaining lint warning, `react-refresh/only-export-components` on `Announcer.tsx`, matching the existing pattern in `I18nContext.tsx` and `ThemeContext.tsx`: a provider and its hook in one file.
* Docs in `docs/accessibility.md`.
* Nothing existing was migrated onto these yet — that is a per-component change with its own review surface. This PR is the layer plus the two wirings that have no downside (skip link, announcer provider).

### DIFF
--- a/docs/accessibility.md
+++ b/docs/accessibility.md
@@ -1,0 +1,179 @@
+# Accessibility Primitives
+
+The frontend has a lot of interactive surfaces — the command palette, dialogs,
+dropdowns, the notification centre, the sidebar — and no shared accessibility
+building blocks. So each one re-solves the same problems, slightly differently
+and slightly wrong.
+
+This is the shared layer. No new dependency; a few hundred lines of DOM code
+with tests.
+
+## What was broken
+
+* **No skip link.** A keyboard user landing on any page had to tab through the
+  entire sidebar and header before reaching the content. On the dashboard that
+  is roughly forty stops, on *every single navigation*.
+* **Nothing trapped focus in a modal.** Tab enough times in a dialog and focus
+  walks into the page behind it — still there, still focusable, covered by an
+  overlay. The user is now typing into something they cannot see.
+* **Focus was never restored.** Closing a dialog dropped focus to `<body>`, so
+  the next Tab started from the top of the document instead of the button just
+  used.
+* **Nothing was announced.** Toasts, "saved", "3 new notifications", search
+  result counts — invisible to a screen reader, because there was no live
+  region. `aria-live` appeared in three files, each hand-rolled.
+* **Composite widgets were N tab stops.** A toolbar should be one stop with
+  arrow-key navigation between items.
+
+## `lib/a11y/tabbable.ts`
+
+The primitive everything else is built on, and the fiddly part.
+
+`container.querySelectorAll("button, a[href], input, …")` gets the common case
+right and then goes wrong on:
+
+* a `disabled` button — **and** one inside a `disabled` fieldset, except
+  anything inside that fieldset's first `<legend>`, which stays interactive
+* anything inside an `inert` subtree
+* `display: none` / `visibility: hidden` / `hidden`
+* collapsed `<details>` content, while keeping its `<summary>`
+* `tabindex="-1"` — focusable by script, skipped by Tab
+* **radio groups**, which are *one* tab stop: the checked member, or the first
+  if none is checked. Treating each as its own stop makes a focus trap loop
+  through a form's radio options one at a time.
+* **positive `tabindex`**, which the browser visits *before* everything in
+  document order
+
+That last one is the most commonly missed. A trap that ignores it wraps in a
+different order than Tab does, and the user ends up somewhere they did not
+expect.
+
+```ts
+import { getTabbableElements, getTabbableEdges, focusFirst } from "@/lib/a11y";
+```
+
+`focusFirst` falls back to focusing the container itself when nothing inside is
+tabbable — focus has to land *somewhere* inside a dialog, or a screen reader
+stays on the page behind it.
+
+## `useFocusTrap`
+
+```tsx
+const ref = useFocusTrap<HTMLDivElement>({
+  active: open,
+  onEscape: () => setOpen(false),
+});
+
+return <div ref={ref} role="dialog">…</div>;
+```
+
+Two details that a minimal version gets wrong:
+
+**Edges are recomputed on every keypress**, not cached when the trap opens.
+Dialog content changes — a validation error appears, a disclosure expands, a
+button becomes enabled — and a cached "last element" sends focus to something
+that is no longer there.
+
+**Focus is pulled back when it has escaped.** The user can click something
+behind the overlay, or a script can move focus. If the trap only handles the
+wrap case, that situation is unrecoverable by keyboard.
+
+Focus is restored to the previously focused element on release, and the hook
+checks the element is still in the document first — it may have been removed
+while the dialog was open.
+
+## `useRovingTabIndex`
+
+Makes a group of controls a single tab stop, with arrow keys moving inside it.
+The WAI-ARIA authoring-practices pattern: exactly one item carries
+`tabIndex={0}`, the rest carry `-1`, and the index moves as the user arrows
+around.
+
+```tsx
+const { containerRef, getItemProps, containerProps } = useRovingTabIndex({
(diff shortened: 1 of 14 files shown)


MOHITKOURAV01:

PR #931, opened 2026-08-06
fix(db): make the migration chain runnable on a fresh database (#926)

### LINKED ISSUE #926: alembic upgrade head fails: the migration graph has 11 unmerged heads
## Summary

`alembic upgrade head` cannot run. The revision graph in `backend/alembic/versions/` has **two heads**, and Alembic refuses to resolve `head` when more than one exists. A second, quieter defect in the same graph: there are **two bases**, and the orphan base alters a table it never declares a dependency on.

## Reproduction

```bash
cd backend
alembic upgrade head
```

```
ERROR [alembic.util.messaging] Multiple head revisions are present for given
argument 'head'; please specify a specific target revision, '<branchname>@head'
to narrow to a specific head, or 'heads' for all heads
FAILED: Multiple head revisions are present for given argument 'head'
```

Reproducible offline with `alembic upgrade head --sql`, so it does not need a live database to see.

## Defect 1 — two heads

```
$ alembic heads
b3b8e5c2e123 (head)
ffff00000002 (head)
```

| Revision | File | Parent |
| --- | --- | --- |
| `b3b8e5c2e123` | `b3b8e5c2e123_add_background_jobs.py` | `('a1b2c3d4e5f7', 'f6a7b8c9d0e1')` |
| `ffff00000002` | `ffff00000002_add_microsoft_id_to_user.py` | `ffff00000001` |

The graph was already merged once — `ffff00000001` is a ten-way merge revision — and then forked again immediately afterwards. `ffff00000002` was written on top of the merge point while `b3b8e5c2e123` was written on top of two of the branches that fed into it. Neither author saw the other's revision.

## Defect 2 — two bases, one of them orphaned

```
$ alembic history | grep -c "<base>"
2
```

`523764cd8096_add_advanced_project_filters.py` declares `down_revision = None`, which makes it a second root of the graph rather than a link in the chain. Its `upgrade()` is:

```python
op.add_column("projects", sa.Column("language", sa.String(length=100), nullable=True))
op.add_column("projects", sa.Column("experience", sa.String(length=50), nullable=True))
op.add_column("projects", sa.Column("is_remote", sa.Boolean(), ...))
op.add_column("projects", sa.Column("is_paid", sa.Boolean(), ...))
op.add_column("projects", sa.Column("is_open_source", sa.Boolean(), ...))
```

`projects` is created by `3d3fd43150c1`, but nothing records that `523764cd8096` must run after it. Alembic topologically sorts disconnected roots, and today's sort happens to place it late enough to work:

```
-- Running upgrade 51602fc145b0 -> c14aa06f723a
CREATE TABLE audit_logs (...)
-- Running upgrade  -> 523764cd8096
ALTER TABLE projects ADD COLUMN language VARCHAR(100);
```

That ordering is incidental, not guaranteed. Nothing in the graph forces it, so a future revision that perturbs the sort turns this into `relation "projects" does not exist` on a fresh database.

## Impact

- A new contributor cannot bring up a local database — the documented setup step fails outright.
- Any deployment running `alembic upgrade head` in an entrypoint or job fails the same way.
- `alembic upgrade heads` (plural) is the only thing that works today, and it silently depends on a fragile ordering.
- Nothing catches either problem in CI, so the next parallel branch forks the graph again.

## Proposed fix

1. Add an Alembic merge revision joining `b3b8e5c2e123` and `ffff00000002` back to a single head. Merge revisions carry no `op.*` calls, so they are safe against databases that are already partly migrated.
2. Reparent `523764cd8096` onto the revision that creates the table it alters, removing the second base and making the ordering explicit.
3. Add a CI check that fails when the graph has more than one head or more than one base, so neither can regress.
4. Document the "rebase your migration onto current head" rule for contributors.

## Acceptance criteria

- [ ] `alembic heads` prints exactly one revision.
- [ ] `alembic history | grep -c "<base>"` returns 1.
- [ ] `alembic upgrade head` resolves and runs.
- [ ] `alembic downgrade base` still works.
- [ ] CI fails a PR that introduces a second head or a second base.
- [ ] Contributor docs explain how to avoid creating a new head.

MOHITKOURAV01:
# Pull Request

## Summary

`alembic upgrade head` did not work, and on investigation the chain had never run to completion on a fresh database at all. This makes it run.

Two problems, both left behind by branches that were merged without anyone re-running the migrations:

**Two heads.** `ffff00000002` was written on top of the ten-way merge at `ffff00000001`, while `b3b8e5c2e123` was written on top of two of the branches feeding into it. Neither author saw the other's revision, so `head` stopped resolving:

```
FAILED: Multiple head revisions are present for given argument 'head'
```

**The branches disagree about `activities`.** `e1d173dca5be` (an autogenerate sweep) creates indexes on `activities.application_id`, `builder_flare_id`, `organization_id`, `project_id` and `repository_id`. `d87970cbb1e6` drops those five columns and replaces them with a generic `target_id` / `target_type` pair. Nothing recorded an order between them, so the topological sort picked one — and both orders failed:

| Order | Failure |
| --- | --- |
| refactor first | `column "application_id" does not exist` on `CREATE INDEX ix_activities_application_id` |
| index sweep first | `relation "ix_activities_created_at" already exists` |

I confirmed the second one is pre-existing by running `alembic upgrade heads` — the only invocation that worked before this PR — against a clean Postgres on unmodified `main`. It fails identically.

---

## Related Issue

Closes #926

---

## Type of Change

- [x] Bug fix
- [ ] New feature
- [x] Documentation update
- [ ] Performance improvement
- [ ] Refactoring
- [ ] UI/UX enhancement
- [x] Tests
- [ ] Other

---

## Changes Made

**The two headline fixes**

- Added `c9a1f2b3d4e5`, a merge revision joining `b3b8e5c2e123` and `ffff00000002`. No `op.*` calls, so it is safe against a database that already ran either branch.
- Pinned the `activities` ordering with `depends_on = ("e1d173dca5be",)` on `d87970cbb1e6`, so the index sweep runs first. Postgres drops a column's indexes along with the column, so the five stale indexes then clean themselves up. The one remaining collision, `ix_activities_created_at` (on a column that survives the refactor), is now created once by `e1d173dca5be`.

**Three more that surfaced once the chain ran far enough to reach them**

- `523764cd8096` had `down_revision = None`, making it a second base. It adds five columns to `projects` without depending on `3d3fd43150c1`, which creates the table — today's sort happened to order it correctly, but nothing guaranteed that. Reparented; revision id unchanged, so applied databases keep their `alembic_version` row.
- `3d3fd43150c1` and `ffff00000002` both create and drop a constraint named `None`. Postgres accepts the unnamed `ADD UNIQUE` going up and invents a name, so this only broke the downgrade: `CompileError: Can't emit DROP CONSTRAINT for constraint UniqueConstraint(); it has no name`. Named both with exactly the name Postgres generates (`users_id_key`, `users_microsoft_id_key`) so already-migrated databases drop the constraint they actually have.
- `b7f6e5d4c3a2` dropped the `submissionstatus` enum right after `hackathon_scores`, but the type belongs to `hackathon_submissions`, dropped later — `cannot drop type submissionstatus because other objects depend on it`. Moved after the owning table.
- `e1d173dca5be`'s downgrade dropped the five activities indexes unconditionally, and they are already gone by then. Switched those five to `DROP INDEX IF EXISTS`.

**Guardrails**

- `backend/scripts/check_migration_graph.py` — asserts one head, one base, and no revision unreachable from head. Reads revision metadata only, no database connection, so it fails fast with a readable message.
- `backend/tests/test_migration_graph.py` — the same assertions in the normal suite, plus regression guards for the orphaned base and the unnamed constraints.
- `.github/workflows/migrations.yml` — runs the graph check, then `alembic upgrade head` against a real Postgres 16 service, then `alembic downgrade base`. The existing suite builds its schema with `Base.metadata.create_all` and never runs a migration, which is exactly why all of this went unnoticed.
- `docs/migrations.md` — the rebase rule, and each of the traps above written up with the error it produces.

---

## Testing

Against Postgres 16 on a clean database:

```
$ alembic upgrade head
... 41 revisions ...
INFO  [alembic.runtime.migration] Running upgrade b3b8e5c2e123, ffff00000002 -> c9a1f2b3d4e5

$ psql -tAc "select count(*) from information_schema.tables where table_schema='public'"
35

$ alembic downgrade base
$ psql -tAc "select count(*) from information_schema.tables where table_schema='public'"
1
```

Before this PR the same commands stop at `column "application_id" does not exist`.

```
$ python -m scripts.check_migration_graph --verbose
head:      c9a1f2b3d4e5
base:      f533805006ed
revisions: 41
Alembic revision graph OK: 1 head, 1 base, no unreachable revisions.

$ pytest tests/test_migration_graph.py -q
6 passed
```

`black --check` and `ruff check` are clean on every file this PR touches. Two files under `alembic/versions/` are unformatted on `main` (`b2e4f6a8c0d1`, `b3b8e5c2e123`); I reformatted them by accident and reverted it to keep the diff focused.

- [x] Tested locally
- [x] Existing tests pass
- [x] Added new tests
- [ ] UI verified

---

## Checklist

---

## Additional Notes

**One thing I deliberately did not fix.** `alembic downgrade base` drops every table but leaves 14 enum types behind (`projectstage`, `activitytype`, `auditaction`, …), because most revisions create their enums implicitly through `sa.Enum(...)` inside a `create_table` and never drop them. The practical effect is that re-upgrading a downgraded database fails with `type "projectstage" already exists` — you have to drop and recreate it.

That is a separate sweep across many revisions and it would have buried this diff, so it is written up in `docs/migrations.md` under "Known limitation" rather than fixed here. Happy to do it as a follow-up.

**On editing already-published revisions.** Four of the changed revisions are ones that some environments may have applied. Every edit either preserves the exact SQL (`depends_on` and the reparent change ordering, not operations) or names a constraint using the identifier Postgres already generated, so no environment is asked to drop something it does not have. Since the chain never completed on a fresh database, any existing environment was built with `create_all` or by hand rather than by these migrations.


nensii21:
@MOHITKOURAV01 frontend job is failing.Check it out.

nensii21:

PR #932, opened 2026-08-06
fix(backend): store timezone-aware UTC everywhere (#927)

### LINKED ISSUE #927: Backend mixes naive datetime.utcnow() with timezone-aware columns, causing TypeErrors and silent time shifts
## Summary

The backend stores timestamps in `DateTime(timezone=True)` columns, and most of the code correctly writes timezone-aware values via `datetime.now(timezone.utc)`. Eight call sites still write **naive** values via the deprecated `datetime.utcnow()`. Mixing the two in the same column is a latent correctness bug and raises `TypeError` as soon as the two kinds of value meet in a comparison.

## Affected call sites

| File | Line | Field written |
| --- | --- | --- |
| `backend/app/services/daily_digest_service.py` | 31 | `since` cutoff used in a query filter |
| `backend/app/services/notification_service.py` | 156 | `Notification.read_at` |
| `backend/app/services/notification_service.py` | 178 | `Notification.read_at` |
| `backend/app/services/refresh_token_service.py` | 60 | `RefreshToken.revoked_at` |
| `backend/app/services/message_service.py` | 159 | `Message.edited_at` |
| `backend/app/services/message_service.py` | 173 | `Message.deleted_at` |
| `backend/app/services/notifications/channels/database_channel.py` | 49 | `Notification.created_at` |
| `backend/app/services/notifications/channels/database_channel.py` | 70 | `sent_at` |

Every one of those targets a column declared as `DateTime(timezone=True)` — for example `Message.edited_at` at `backend/app/models/message.py:157` and `Notification.read_at` at `backend/app/models/notification.py:246`.

## Why this is a bug

1. **Comparison crashes.** `datetime.utcnow()` returns a naive datetime. Any expression that puts it next to an aware value from the ORM raises:
   ```
   TypeError: can't compare offset-naive and offset-aware datetimes
   ```
   `daily_digest_service.py:31` builds its `since` cutoff this way and then filters rows whose columns are aware.

2. **Silent time shifts.** Writing a naive datetime into a `TIMESTAMP WITH TIME ZONE` column makes the driver interpret it in the session time zone rather than UTC. On any deployment where the DB session is not UTC, `read_at` / `edited_at` / `revoked_at` land at the wrong instant. Nothing errors — the data is just wrong, and only for the rows written by these eight lines.

3. **Deprecated API.** `datetime.utcnow()` is deprecated as of Python 3.12, which is the version pinned in `backend-ci.yml`. It emits a `DeprecationWarning` today and is scheduled for removal.

## Proposed fix

1. Add a single `backend/app/utils/time.py` helper exposing `utcnow()` (aware, UTC) plus small companions for the patterns already repeated across services — `ensure_utc()` to normalise a value that may be naive, and `isoformat_utc()` for serialisation.
2. Replace all eight `datetime.utcnow()` call sites with the helper.
3. Add unit tests: the helper returns an aware value, `ensure_utc()` normalises naive input without shifting the wall clock, and the previously-crashing digest comparison now works.
4. Add a guard so the deprecated call cannot come back — a `ruff` rule plus a pre-commit grep.

## Acceptance criteria

- [ ] `grep -rn "datetime.utcnow()" backend/app` returns nothing.
- [ ] Every timestamp written by the services is timezone-aware UTC.
- [ ] Tests cover the helper and the digest-cutoff comparison that used to raise.
- [ ] A reintroduced `datetime.utcnow()` fails lint.

MOHITKOURAV01:
# Pull Request

## Summary

Every timestamp column in the app is declared `DateTime(timezone=True)`, and most of the code already wrote timezone-aware values via `datetime.now(timezone.utc)`. Eight call sites across five services still used `datetime.utcnow()`, which returns a **naive** datetime.

While writing the tests for this I found it is not just a tidiness problem — one of those call sites means `DailyDigestService.generate_daily_digest()` has been raising on Postgres every time it ran:

```python
since = datetime.utcnow() - timedelta(days=1)     # naive
...
if notification.created_at >= since               # aware, from Postgres
# TypeError: can't compare offset-naive and offset-aware datetimes
```

The existing tests never caught it because the suite runs on SQLite, which returns **naive** datetimes — so both sides happened to match and the comparison worked. I added a test that forces `created_at` to be aware, the way Postgres returns it. It fails on `main` with exactly that `TypeError` and passes here.

The quieter half of the problem affects the other seven sites: writing a naive datetime into a `TIMESTAMP WITH TIME ZONE` column makes the driver interpret it in the session time zone rather than UTC. On any deployment whose session is not UTC, `read_at` / `edited_at` / `revoked_at` / `sent_at` land at the wrong instant, with nothing raised.

`datetime.utcnow()` is also deprecated as of Python 3.12, the version pinned in `backend-ci.yml`.

---

## Related Issue

Closes #927

---

## Type of Change

- [x] Bug fix
- [ ] New feature
- [x] Documentation update
- [ ] Performance improvement
- [x] Refactoring
- [ ] UI/UX enhancement
- [x] Tests
- [ ] Other

---

## Changes Made

**New helper — `backend/app/utils/time.py`**

| Function | Purpose |
| --- | --- |
| `utcnow()` | Aware UTC now. The direct replacement for `datetime.utcnow()`. |
| `ensure_utc(value)` | Normalise a value that may be naive or in another zone. Naive input is assumed UTC and tagged, matching how the existing naive rows were written. |
| `isoformat_utc(value)` | Serialise as ISO 8601 UTC, normalising first so nothing serialises without an offset. |
| `utc_from_timestamp(seconds)` | Aware replacement for the equally-deprecated `datetime.utcfromtimestamp()`. |
| `is_expired(expires_at, *, leeway=None)` | Expiry check that normalises first, so it works against rows written before this cleanup. |

**Call sites replaced** — all eight:

- `daily_digest_service.py:31` (the digest cutoff)
- `notification_service.py:156, 178` (`read_at`)
- `refresh_token_service.py:60` (`revoked_at`)
- `message_service.py:159, 173` (`edited_at`, `deleted_at`)
- `notifications/channels/database_channel.py:49, 70` (`created_at`, `sent_at`)

Plus the actual digest fix: `ensure_utc()` on `notification.created_at` before the comparison, so it holds on Postgres (aware), on SQLite (naive), and for rows written before this PR (naive).

**Consistency within the touched files.** I also converted the seven existing `datetime.now(timezone.utc)` / `datetime.now(UTC)` calls in those same five files. They were already correct — but `refresh_token_service.py` having `utcnow()` on line 60 and `datetime.now(timezone.utc)` on lines 101/135/145/172, all writing the same model, is precisely how this drift started. The other **121** occurrences elsewhere in `app/` are correct and deliberately untouched; converting them would have been churn.

**Guards, so it cannot come back**

- `backend/ruff.toml` enables `DTZ003` / `DTZ004`. It uses `extend-select`, so it *adds* to Ruff's defaults rather than replacing them — nothing else about how existing code is linted changes. Verified against the `v0.2.1` pinned in `.pre-commit-config.yaml`, and the existing pre-commit ruff hook picks it up with no config change.
- `tests/test_utc_time.py` walks `app/` and fails on any reintroduced `datetime.utcnow()` / `datetime.utcfromtimestamp()`. Blunt, but it is the only thing that catches one added to a service nobody has written a test for — which is how the original eight accumulated.
- `docs/coding-standards.md` gains a "Dates and Times" section with both failure modes.

---

## Testing

**No regressions.** I captured the full failure list on `main` and on this branch and diffed them:

```
main:   94    branch:   94
=== only in branch (regressions) ===
(none)
=== only in main (fixed) ===
(none)
```

`84 failed, 895 passed, 6 skipped, 10 errors` here versus `84 failed, 876 passed, 6 skipped, 10 errors` on `main` — the same 94 pre-existing failures, plus my 19 new tests passing. Those 94 are unrelated (websockets, rbac, search, tech_stack) and predate this branch.

**The bug, demonstrated.** With `main`'s `daily_digest_service.py` restored under the new test:

```
E       TypeError: can't compare offset-naive and offset-aware datetimes
1 failed
```

and with the fix in place, `4 passed`.

**New tests:** 18 in `tests/test_utc_time.py` covering each helper — including that a naive value is tagged rather than shifted, that another zone is converted (12:00 IST → 06:30 UTC), and that `is_expired` handles naive input — plus the digest regression test.

**Lint and format**, against the versions pinned in `.pre-commit-config.yaml` (`ruff==0.2.1`, `black==24.2.0`): every file this PR adds is clean. `notification_service.py` and `test_daily_digest_service.py` still carry pre-existing `F401`/`F821` and black findings in regions I did not touch; I left them rather than mix unrelated churn into this diff. For what it is worth the change reduces `notification_service.py` from four black hunks to two.

- [x] Tested locally
- [x] Existing tests pass
- [x] Added new tests
- [ ] UI verified

---

## Checklist

---

## Additional Notes

**Existing naive rows are not migrated.** Any timestamp already written naive into a `timestamptz` column by one of these eight sites is stored at whatever instant the session time zone implied, and this PR cannot tell which of those are wrong. `ensure_utc()` makes reading them safe — they will be treated as UTC — but a backfill would need to know each deployment's historical session time zone. If everything has been running in UTC, which is the common case, there is nothing to fix.

**`ensure_utc()` assumes naive means UTC.** That matches the intent of the code being replaced (`utcnow()` meant "UTC, just untagged"), and it is the only assumption that keeps old rows readable. Worth knowing if you ever feed it a datetime from a client that meant local time — normalise at the boundary instead.


nensii21:
@MOHITKOURAV01 frontend and backend job is failing.Check it out.

nensii21:

PR #866, opened 2026-08-03
feat: redesign landing page pricing section

dristicg:
### Description
Redesigned the pricing section on the landing page to improve visual hierarchy and readability. Added a monthly/yearly toggle, highlighted the recommended "Pro" plan, and updated the layout for clearer feature comparison.

Closes #764 

### Screenshots
WhatsApp Image 2026-08-03 at 8 12 12 PM

### Testing Performed
- Verified the monthly/yearly toggle correctly updates pricing.
- Confirmed responsive layout works perfectly on mobile, tablet, and desktop views.
- Verified animations and transitions trigger correctly.

### Notes
- Uses Framer Motion for the toggle animation.

PR #933, opened 2026-08-06
chore(security): untrack backend/.env and remove committed scratch files (#928)

### LINKED ISSUE #928: backend/.env is tracked: it leaks a dev database password and silently overrides config defaults
> **Correction (severity revised down).** This issue originally said the tracked `.env` carried a *live* `SECRET_KEY`. It does not — the committed value is the placeholder `CHANGE_THIS_TO_A_LONG_RANDOM_SECRET_KEY_AT_LEAST_64_CHARACTERS`, and every API key and OAuth secret in the file is an empty string. **No key rotation is required.** The real problems are smaller but still worth fixing: one developer's local Postgres password is exposed, the file silently overrides config defaults for everyone, and the `.gitignore` rule that appears to protect it does nothing. Rewritten below.

## Summary

`backend/.env` is tracked in git. So are a handful of one-off scratch scripts and a 39 KB API response dump that were committed by accident.

`.gitignore` lists `.env`, but the rule has never applied here, for two reasons:

1. **Git keeps tracking files already in the index.** `.gitignore` only affects *untracked* files, so a rule added after the file was committed does nothing.
2. **A bare `.env` only matches the repository root.** Even on a clean checkout it would not have matched `backend/.env`.

So the protection reads as if it is in place while the file sits in the repository.

## Tracked files that should not be

```
$ git ls-files | grep -iE "\.env$|pr866|fix\.py|resolve_merge|fix_all|patch_"
backend/.env
backend/fix_all.py
backend/patch_followers.py
backend/patch_imports.py
fix.py
pr866.json
resolve_merge.py
```

### `backend/.env` — 163 lines

`backend/.env.example` already exists and is the file that is meant to be committed. It documents every variable the real file holds, plus three it is missing.

**What is actually exposed.** Diffing the committed values against the defaults declared in `app/core/config.py` gives eight divergences, and only one of them is a credential:

| Setting | Committed value | Assessment |
| --- | --- | --- |
| `DATABASE_URL` | `postgresql+psycopg://postgres:<password>@localhost:5432/devlink` | **A real password.** Local-only, but developers reuse passwords. |
| `SECRET_KEY` | `CHANGE_THIS_TO_A_LONG_RANDOM_SECRET_KEY_AT_LEAST_64_CHARACTERS` | Placeholder. Not a live key. |
| `SMTP_HOST` | `smtp.gmail.com` | Not sensitive. |
| `SEARCH_RATE_LIMIT` | `60/minute` | **Silently overrides** the `30/minute` in `config.py` — see below. |
| `AWS_*` | `""` | Empty. |
| `SMTP_USERNAME`, `SMTP_PASSWORD`, `GITHUB_CLIENT_SECRET`, `GOOGLE_CLIENT_SECRET`, `OPENAI_API_KEY`, `GEMINI_API_KEY`, `CLOUDINARY_*` | `""` | Empty. |

**The quieter problem is the silent override.** pydantic-settings loads `backend/.env` automatically, so a tracked one wins over `config.py` on every machine and in CI. `tests/test_rate_limiting.py` asserts `settings.SEARCH_RATE_LIMIT == "60/minute"` and passes — not because that is the declared default (it is `30/minute`) but because the committed `.env` overrode it. Untracking the file makes that test fail, which is the bug surfacing rather than a new one.

### Scratch scripts

`fix.py`, `resolve_merge.py`, `backend/fix_all.py`, `backend/patch_followers.py`, `backend/patch_imports.py` and their siblings are single-use codemods from past merge cleanups. `fix.py` still hardcodes `c:\Users\drish\Desktop\...`; `patch_followers.py` inserts a debug `print` into a router. `backend/test_fastapi_yield.py` and `backend/test_pydantic_alias.py` match `test_*.py`, so pytest collects them — the latter just prints at import.

### `pr866.json`

A 39 KB dump of the GitHub API response for PR #866, in the repository root.

### `backend/gitignore`

Missing its leading dot, so git has never read it — 40 lines of silently inert rules.

## Impact

- A developer's Postgres password is in the history, in every clone and fork.
- Config defaults are overridden repository-wide by one person's local file, so tests assert values the code does not declare.
- New contributors copying `backend/.env` inherit someone else's database URL instead of writing their own from the example.
- `.gitignore` reads as if `.env` is protected, so nobody notices.
- Linters, test collection, and CodeQL all walk the stray scripts.

## Proposed fix

1. `git rm --cached backend/.env`; delete the scratch scripts, `pr866.json`, and the misnamed `gitignore`.
2. Tighten `.gitignore` so the patterns match at any depth, keeping `.env.example` committable.
3. Add a CI job that inspects the **index**, since that is the thing `.gitignore` cannot fix.
4. Mirror the check in `.pre-commit-config.yaml`.
5. Fix `test_rate_limiting.py` to assert the declared defaults rather than whatever the local `.env` supplies.
6. Document in the security docs why `.gitignore` does not untrack, and how to rotate if a real secret is ever committed.

## Acceptance criteria

- [ ] `git ls-files` lists no `.env`, no `pr866.json`, no `fix_*`/`patch_*` scratch script, and no misnamed `gitignore`.
- [ ] `backend/.env.example` stays tracked.
- [ ] A newly added `.env` at any depth is ignored.
- [ ] CI fails a PR that tracks one of these files.
- [ ] The rate-limit test passes both with and without a local `.env`.

MOHITKOURAV01:
# Pull Request

## Summary

> **Correction to my original description.** I first wrote that this `.env` carried a *live* `SECRET_KEY` and that rotation was urgent. That was wrong, and I have revised #928 to match. The committed value is the placeholder `CHANGE_THIS_TO_A_LONG_RANDOM_SECRET_KEY_AT_LEAST_64_CHARACTERS`, and every API key and OAuth secret in the file is an empty string. **No key rotation is needed.** Apologies for the false alarm — the rest of the change stands, for the reasons below.

`backend/.env` is tracked in git. `.gitignore` lists `.env`, but that rule has never applied here, for two reasons:

1. **Git keeps tracking files already in the index.** `.gitignore` only affects *untracked* files, so a rule added after the file was committed does nothing.
2. **A bare `.env` only matches the repository root.** Even on a clean checkout it would not have matched `backend/.env`.

So the protection reads as if it is in place while the file sits in the repository.

Two things actually go wrong as a result.

**A developer's local Postgres password is in the history**, inside `DATABASE_URL`. Local-only, but people reuse passwords, so it is worth changing.

**The tracked `.env` silently overrides `config.py` for everyone.** This is the one I did not expect. pydantic-settings loads `backend/.env` automatically, so a committed one wins on every machine *and in CI*:

| Setting | `config.py` declares | Committed `.env` supplies |
| --- | --- | --- |
| `SEARCH_RATE_LIMIT` | `30/minute` | `60/minute` |

`tests/test_rate_limiting.py` asserted `settings.SEARCH_RATE_LIMIT == "60/minute"` and passed everywhere — not because that is the default, but because one person's local file said so. See the testing section; this PR fixes that test properly rather than flipping the string.

Alongside the `.env`: twelve single-use scratch scripts, a 39 KB API-response dump, and a `.gitignore` missing its leading dot.

---

## Related Issue

Closes #928

---

## Type of Change

- [x] Bug fix
- [ ] New feature
- [x] Documentation update
- [ ] Performance improvement
- [ ] Refactoring
- [ ] UI/UX enhancement
- [ ] Tests
- [x] Other — repository hygiene / secret exposure

---

### `backend/.env` — untracked, not deleted

```bash
git rm --cached backend/.env
```

Your local copy is untouched. `backend/.env.example` was already tracked and documents every variable the real file held, plus three it was missing (`CELERY_TASK_ALWAYS_EAGER`, `LINKEDIN_CLIENT_ID`, `LINKEDIN_CLIENT_SECRET`) — so nothing is lost.

I diffed all 163 committed values against the defaults in `app/core/config.py`. Eight diverge, and only one is a credential:

| Setting | Committed value | Assessment |
| --- | --- | --- |
| `DATABASE_URL` | `postgresql+psycopg://postgres:<password>@localhost:5432/devlink` | **A real password.** Worth changing. |
| `SECRET_KEY` | `CHANGE_THIS_TO_A_LONG_RANDOM_SECRET_KEY_AT_LEAST_64_CHARACTERS` | Placeholder. Not a live key. |
| `SEARCH_RATE_LIMIT` | `60/minute` | Silently overrides `30/minute` — see the summary. |
| `SMTP_HOST` | `smtp.gmail.com` | Not sensitive. |
| `AWS_*`, all OAuth secrets, all API keys | `""` | Empty. |

### Twelve dead scratch files

| File | What it is |
| --- | --- |
| `fix.py` | Codemod against `c:\Users\drish\Desktop\devlink\...` — a hardcoded path on someone else's machine |
| `resolve_merge.py` | 255-line one-off conflict resolver from a past merge |
| `backend/fix_all.py`, `fix_override_get_db.py`, `fix_recs.py`, `fix_tests.py`, `fix_websockets.py`, `fix_ws.py` | Single-use test-file rewrites |
| `backend/patch_followers.py` | Inserts a `print(f'ENQUEUE ERROR: {e}')` into a router |
| `backend/patch_imports.py` | Rewrites the import block of `auth.py` |
| `backend/test_fastapi_yield.py` | Scratch FastAPI experiment with its own app and models; its one "test" asserts nothing and prints |
| `backend/test_pydantic_alias.py` | A bare `print()` at module scope — no tests at all |

The last two matter beyond tidiness: both match `test_*.py`, so pytest collects them, and `test_pydantic_alias.py` executes its print on import during every run.

### `pr866.json`

39 KB dump of the GitHub API response for PR #866, committed to the repository root.

### `backend/gitignore`

Missing its leading dot, so git never read it — 40 lines of rules that have been silently inert. Its entries that the root file lacked (`*.db`, `*.log`, `.pytest_cache/`, `htmlcov/`, `.coverage`, `docker-data/`, `Thumbs.db`) are folded into the root `.gitignore` rather than left in a second, easily-missed file.

### Guards

Because a `.gitignore` rule would not have caught any of this, the checks look at the index instead:

- **`.github/workflows/tracked-files.yml`** — fails a PR that tracks a `.env` at any depth (`.example` excluded), a scratch script, or a file named `gitignore`. A second step asserts the ignore rules actually match what they claim: `.env`, `backend/.env`, `frontend/.env`, `backend/.env.local` must be ignored, and both `.env.example` files must *not* be. A rule that does not do what it looks like it does is worse than no rule, which is the whole story here.
- **`.pre-commit-config.yaml`** — two `language: fail` hooks refusing these paths locally, plus `detect-private-key` for a key pasted into any file.
- **`.gitignore`** — `**/.env` / `**/.env.*` with `!**/.env.example` negations, so depth is covered and the templates stay committable.

### Docs

- `docs/security.md` — a "Secrets and Environment Files" section explaining why `.gitignore` does not untrack, and step-by-step `SECRET_KEY` rotation.
- `docs/environment-setup.md` — a note at the `cp .env.example .env` step.

---

## Testing

**Guard logic, run against the real index** — before the change each check found its targets; after, all three are clean:

```
secrets: PASS      scratch: PASS      misnamed: PASS
```

**Negative test.** Staged a `backend/.env.staging` and confirmed the secrets check fires on it, then removed it.

**Ignore rules:**

```
ignored:     .env, backend/.env, frontend/.env, backend/.env.local
committable: backend/.env.example, frontend/.env.example
```

**Nothing references the removed files.** `grep` across `*.py`, `*.yml`, `*.toml`, `*.cfg`, `*.md`, `*.json` returns only the two new guard patterns.

**Backend still imports:** `from app.main import app` succeeds.

**The test that the untracking broke, and how it is fixed.** Removing the `.env` made `test_rate_limit_settings_defined` fail, because it asserted `SEARCH_RATE_LIMIT == "60/minute"` — the committed override, not the `30/minute` declared in `config.py`. Flipping the string would only move the problem: the test would then fail for any contributor whose own `.env` sets a different value.

Instead it now builds a `Settings` subclass with `env_file` disabled, so it asserts what the code declares and means the same thing everywhere. A second test covers the looser property that the active config parses at all. Verified passing **both with and without** a local `.env` present — which the original could not manage.

**Test collection: 976 → 975.** Exactly the one scratch test (`test_fastapi_yield.py::test_fastapi`). Full suite failure set is byte-identical to `main` — the same 94 pre-existing failures, no regressions, nothing newly resolved.

Both new/changed YAML files parse.

- [x] Tested locally
- [x] Existing tests pass
- [ ] Added new tests
- [ ] UI verified

---

## Checklist

---

## Additional Notes

> **On the credential.** The only real one is the Postgres password in `DATABASE_URL`. It is a `localhost` URL, so it is not directly reachable, but whoever owns it should change it — and anywhere else they used the same password. `SECRET_KEY` was a placeholder and needs nothing.
>
> Removing the file from `HEAD` does not remove it from history; it is still in every clone and fork. I have deliberately **not** rewritten history — that would break every open PR and every fork, and it is a maintainer call. For a placeholder key and a localhost password I do not think it is warranted, but the procedure is written up in `docs/security.md` if you disagree.

**One thing worth a second opinion:** I folded `backend/gitignore`'s rules into the root `.gitignore` and deleted it, rather than renaming it to `backend/.gitignore`. One file is easier to reason about than two overlapping ones, but if you would rather keep backend concerns local, renaming is a one-line change.


nensii21:
image
resolve it.

nensii21:

Issue #667, opened 2026-07-27
Bug while creating user account


dharapandya85:
<img width="1875" height="916" alt="Image" src="https://github.com/user-attachments/assets/9db35ab7-1e25-4973-a46f-46261945df2c" />

dharapandya85:
please assign appropriate labels, @nensii21

dharapandya85:
@nensii21 , pr already submitted and merged, guide further if pr is required, or close this issue.

dharapandya85:
@nensii21 , pr submitted, please review.

PR #934, opened 2026-08-06
feat(backend): add a deterministic, idempotent database seeder (#929)

### LINKED ISSUE #929: No database seeding script: a fresh local setup leaves every table empty
## Summary

There is no way to get a populated local database. After a contributor runs the migrations they are left with 50+ empty tables, so every screen in the app renders its empty state and nothing can be exercised without hand-creating rows through the API first.

`backend/scripts/` currently holds `backup_database.py`, `backup_storage.py`, `export_openapi.py`, `generate_sdks.py`, and `restore_database.py`. Nothing seeds.

## Why this hurts

The app has 63 routers and 53 models. Most features are only visible when there is data behind them:

- The dashboard, feed, and recommendation panels all read from tables that start empty.
- Search returns nothing, so the full-text indexes and ranking code are untestable locally.
- Messaging needs two users and a conversation before the UI does anything.
- Reviewing a frontend PR means manually registering users, creating projects, and wiring memberships every time the database is reset.

`docs/environment-setup.md` walks through installing dependencies and running migrations and then stops. The gap is real enough that several open frontend issues describe work against mock data because live data is inconvenient to produce.

## Proposed fix

Add `backend/scripts/seed_database.py`:

- **Deterministic.** A fixed RNG seed so the same command produces the same fixtures every time, and screenshots in PR descriptions are comparable across contributors.
- **Idempotent.** Re-running upserts on a natural key rather than duplicating. Safe to run against a database that has already been seeded.
- **Covers the core graph**, not just users: users with hashed passwords and profiles, skills and user-skill links, projects with owners and members, applications, conversations with messages, notifications, activities, and bookmarks.
- **Known credentials.** A documented demo login so a contributor can sign in immediately.
- **CLI flags** for scale and cleanup:
  - `--users N`, `--projects N` to control volume
  - `--reset` to delete previously seeded rows before inserting
  - `--dry-run` to print the plan without writing
  - `--quiet` for CI use
- **Refuses to run against production.** Hard guard on `ENVIRONMENT`, overridable only with an explicit flag.

Plus tests that the seeder is idempotent and that `--dry-run` writes nothing, and a docs section wiring it into the setup flow.

## Acceptance criteria

- [ ] `python -m scripts.seed_database` populates a freshly migrated database.
- [ ] Running it twice does not duplicate rows.
- [ ] `--dry-run` reports what it would insert and commits nothing.
- [ ] `--reset` clears prior seed data without touching the schema.
- [ ] The script aborts when `ENVIRONMENT=production` unless explicitly forced.
- [ ] `docs/environment-setup.md` documents the command and the demo credentials.

MOHITKOURAV01:
# Pull Request

## Summary

A freshly migrated database is empty, so every screen in the app renders its empty state — the dashboard has nothing to show, search returns nothing, and messaging needs two users and a conversation before it does anything at all. Reviewing a frontend change meant hand-registering users and creating projects through the API every time the database was reset.

`backend/scripts/seed_database.py` fills in the core object graph:

```bash
$ python -m scripts.seed_database

Seeded:
  users                  13
  skills                 24
  user skills            55
  projects               18
  project members        64
  builder flares         10
  applications           18
  conversations           6
  messages               29
  notifications          46
  bookmarks              39

308 rows written.

Sign in with any demo account:
  admin@example.com          (superuser)
  aditialmeida@example.com   (regular user)
  password: DevlinkDemo!2026
```

---

## Related Issue

Closes #929

---

## Type of Change

- [x] Bug fix — two pre-existing blockers, see below
- [x] New feature
- [x] Documentation update
- [ ] Performance improvement
- [ ] Refactoring
- [ ] UI/UX enhancement
- [x] Tests
- [ ] Other

---

### The seeder

| Flag | Effect |
| --- | --- |
| `--users N`, `--projects N` | Control volume |
| `--reset` | Delete previously seeded rows first |
| `--dry-run` | Report what would be written, roll back |
| `--quiet` | Summary only, for CI |
| `--force` | Required when `ENVIRONMENT=production` |

**Idempotent.** Rows are matched on a natural key (email, slug, normalized skill name) and updated rather than duplicated.

Getting this right needed more than upserts, and it is the part I'd most like a second pair of eyes on. The first version used one shared `random.Random(SEED)` and looked completely correct — but on a second run the existence checks skipped work, which skipped the random draws that work would have made, which shifted every later draw along. The result was 144 duplicate rows:

```
run 1: bookmarks=39 applications=16 project_members=43
run 2: bookmarks=68 applications=33 project_members=86
```

Each entity now derives its own generator from its own key (`rng_for("bookmarks/{user.id}")`), so what a project or user gets depends only on itself and never on what happened to be created before it. Three consecutive runs are now byte-identical.

**Deterministic.** Fixed seed, fixed timestamp epoch, and UUIDs derived via `uuid5` from natural keys. The same flags produce the same database on every machine, so a screenshot in a PR means the same thing to the reviewer as it did to the author.

**`--reset` is not a `TRUNCATE`.** Seeded rows carry a `[demo]` marker in their bio/description; only those are removed, so anything you created by hand while testing survives. There is a test for exactly that.

**Safety.** Refuses to run when `ENVIRONMENT` is `production`, since it writes accounts whose password is published in the docs.

### Two pre-existing bugs that had to be fixed first

The seeder cannot get past either of these, so they are included rather than split out.

**1. `app/core/security.py` — password hashing was completely broken.**

The bcrypt-truncation patch appears **twice**, an identical block duplicated presumably by a bad merge. The second copy captured the *already-patched* function as its `_original_hashpw`, so `_patched_hashpw` called itself:

```
$ python -c "from app.core.security import hash_password; hash_password('hunter2')"
  [Previous line repeated 980 more times]
RecursionError: maximum recursion depth exceeded
```

`hash_password()` has been unusable, which takes registration with it. Nothing caught it because `tests/conftest.py` replaces `pwd_context` with a mock before the suite runs, so bcrypt is never exercised in tests. This may well be what open issue #667 ("Bug while creating user account") is reporting — I can't confirm from the screenshot, but the symptom fits.

Removed the duplicate and made reapplying the patch a no-op via a `_devlink_patched` marker, so a re-import or a future duplicate cannot recreate the loop.

**2. `User.badges` — `create_all()` could not build the schema on Postgres.**

```python
server_default="[]",   # a Postgres text[] wants '{}'
```

```
psycopg.errors.InvalidTextRepresentation: malformed array literal: "[]"
```

Unnoticed because the tests `create_all` against SQLite, where the column is JSON and `[]` is valid. The migration that adds this column (`1a2b3c4d5e6f`) already uses `'{}'`, so this brings the model in line with the schema that actually exists rather than changing it.

---

## Testing

**30 new tests** in `tests/test_seed_database.py` — all pass:

- Every table the seeder claims to fill is non-empty; every project has an owner member row; every application points at a flare on its own project; every conversation has at least two messages.
- Idempotency across **two and three** runs, and no duplicate emails or slugs.
- Determinism: same IDs and same profile content after a reset-and-reseed.
- Reset: removes seeded rows, cascades to dependents, and leaves a hand-created user alone.
- Argument parsing and slug generation.

**Manual verification** against Postgres 16 and SQLite:

```
run 1: users=13 skills=24 user_skills=55 projects=18 project_members=64 ... bookmarks=39
run 2: users=13 skills=24 user_skills=55 projects=18 project_members=64 ... bookmarks=39
run 3: users=13 skills=24 user_skills=55 projects=18 project_members=64 ... bookmarks=39
```

`--dry-run` reports its plan and commits nothing. `--reset` followed by a seed restores identical counts. `ENVIRONMENT=production` is refused with exit code 2.

**No regressions.** Failure list diffed against `main`: identical 94 pre-existing failures, nothing new, nothing accidentally resolved. `84 failed, 906 passed` here versus `84 failed, 876 passed` on `main` — the +30 is exactly this PR's tests.

**Lint**, at the pinned `ruff==0.2.1` / `black==24.2.0`: the two files I touched go from **8 errors on `main` to 3**, all three pre-existing (a duplicated `UserRole` class in `user.py`, and the `E402` inherent to patching bcrypt before importing passlib).

- [x] Tested locally
- [x] Existing tests pass
- [x] Added new tests
- [ ] UI verified

---

## Checklist

---

### Something bigger I found and did **not** fix

While trying to verify the seeder against a properly migrated Postgres database, I hit two more problems that are well beyond this PR. Flagging them because they are more serious than anything here, and I think they deserve their own issues:

**The migrations do not produce the schema the models describe.** Comparing `Base.metadata` against a database migrated to head:

- **25 tables** in the models have no migration at all: `api_keys`, `badges`, `user_badges`, `project_documents`, `project_milestones`, `project_versions`, `project_views`, `security_events`, `verification_requests`, `webhook_deliveries`, `workspace_api_tokens`, `plugins`, `maintenance_windows`, `password_reset_tokens`, `request_logs`, and ten more.
- **27 columns** are missing across four existing tables — including `users.availability`, `users.system_role`, `users.mfa_enabled`, `projects.is_published`, `projects.tags`, and both of `bookmarks.target_id` / `target_type`.

The practical effect is that a migrated database cannot serve a `SELECT` on `User` at all:

```
psycopg.errors.UndefinedColumn: column users.availability does not exist
```

Those tables only ever existed because dev and test environments build the schema with `create_all` rather than by migrating.

**`create_all` does not work on Postgres either**, even after the `badges` fix above: `verification_requests.user_id` is `String(36)` while `users.id` is `UUID`, so the foreign key cannot be created (`DatatypeMismatch`). SQLite doesn't care, which is why the test suite is green.

So the seeder is verified against a `create_all` SQLite schema and against Postgres up to the point those two problems allow. Once they are resolved it should work unchanged on a migrated Postgres — nothing in it depends on how the schema was built.

### Depends on #931

The seeder needs a database, and `alembic upgrade head` currently fails outright. #931 fixes that. This PR does not depend on it to *merge*, but the documented workflow (migrate, then seed) only works end-to-end with both.

### The demo password is intentionally public

`DevlinkDemo!2026` is in the source, in the docs, and printed on every run. That is the point — it is a shared local fixture, not a secret. The `ENVIRONMENT=production` guard exists precisely because a published password must never reach a real deployment.


nensii21:
image
resolve it.

nensii21:

Issue #492, opened 2026-07-23
Rich Text Editor


nensii21:
Support

Images
Videos
Code Blocks
Markdown
Mentions
Emoji

Santhosh1040:
Hi @nensii21 I'd like to work on this issue under EcSoC'26. Could you please assign it to me???

Santhosh1040:
working on it

Santhosh1040:
Hi @nensii21 , I've looked into the current implementation before starting work. The project currently has a lightweight MarkdownEditor component that uses a plain <textarea> with a markdown preview (react-markdown). Markdown and code blocks are already supported, but features like image/video embedding, mentions (@user), and emoji support are not present and would require introducing a full rich text editor framework (such as TipTap, Lexical, or Slate) or a significant rewrite of the existing component.

Could you clarify the expected scope for this issue? Should this be a full editor migration, or would you prefer implementing a smaller subset of these features first?

nensii21:
@Santhosh1040 A full editor migration isn't required for this issue. Please keep the existing MarkdownEditor and implement only the features that can be added without a major rewrite. If a feature requires migrating to a rich text editor like TipTap, Lexical, or Slate, leave it out for now. Focus on incremental improvements that integrate well with the current implementation, and document any limitations or follow-up work needed for the remaining features.

PR #935, opened 2026-08-06
feat(projects): add threaded project discussions (#930)

### LINKED ISSUE #930: Add threaded project discussions so collaboration questions do not have to happen in DMs
## Summary

A project page has no discussion surface. There is no comment model anywhere in `backend/app/models/` — the only hits for "comment" in the backend are an unrelated `hackathon_score.comment` field and a column on the activity model. If two people want to talk about a project they have to leave for direct messages, which loses the connection to the project entirely.

## Current state

- Projects have owners, members, milestones, documents, tags, versions, and an audit trail.
- Conversations and messages exist, but they are private 1:1/group DMs, not attached to a project.
- Applications carry a single message from the applicant, which is not a thread.

So the one thing missing is the public, threaded, project-scoped conversation — the "issue comments" equivalent that every collaboration platform has.

## Why it matters

Someone browsing a project and deciding whether to apply has no way to ask a question in the open. The answer is useful to the next twenty people who look at the project, but a DM means it is written once and seen by one person. This is the discussion layer the collaboration flow currently routes around.

## Proposed scope

**Data model** — a `project_comments` table:

- `id`, `project_id`, `author_id`, `body`
- `parent_id` self-referential FK for one level of replies (flat replies under a top-level comment, not unbounded nesting)
- `is_edited`, `edited_at`
- `deleted_at` for soft delete, matching the convention already used by `Project`, `User`, `Organization`, and `Message`
- `created_at` / `updated_at` as timezone-aware columns
- Indexes on `(project_id, created_at)` and `parent_id`

**API** under `/api/v1/projects/{project_id}/comments`:

- `GET` — paginated top-level comments with their replies and reply counts
- `POST` — create a comment or a reply
- `PATCH /{comment_id}` — edit, author only, sets the edited marker
- `DELETE /{comment_id}` — soft delete, author or project owner
- Visibility follows the project: comments on a private project are only readable by people who can read the project.

**Rules**

- Body length bounded, empty/whitespace-only rejected.
- Replies must belong to the same project as their parent.
- A reply cannot itself be replied to — one level only.
- Deleted comments are tombstoned rather than removed, so reply chains stay intact.

**Frontend** — an API module plus a comment thread component on the project page, with compose, reply, edit, and delete.

## Acceptance criteria

- [ ] Migration creates `project_comments` with the indexes and FKs above, and downgrades cleanly.
- [ ] All five endpoints are implemented, authenticated, and permission-checked.
- [ ] A user cannot edit or delete someone else's comment; a project owner can delete any comment on their project.
- [ ] Soft-deleted comments are excluded from listings but their replies survive.
- [ ] Backend tests cover the permission matrix and the reply-depth rule.
- [ ] The project page renders the thread and supports compose/reply/edit/delete.

MOHITKOURAV01:
# Pull Request

## Summary

A project page has no discussion surface. Someone browsing a project and deciding whether to apply can only ask in a DM — where the answer is written once and seen by one person, and the next twenty ask the same question. This adds the public, project-scoped thread that the collaboration flow currently routes around.

Five endpoints under `/api/v1/projects/{project_id}/comments`, backed by a new `project_comments` table, plus the frontend module and component.

---

## Related Issue

Closes #930

---

## Type of Change

- [x] Bug fix — a pre-existing validation-handler crash, see below
- [x] New feature
- [x] Documentation update
- [ ] Performance improvement
- [ ] Refactoring
- [x] UI/UX enhancement
- [x] Tests
- [ ] Other

---

### Design decisions worth reviewing

**Threading is one level deep.** A comment is either top-level or a reply to one that is. Arbitrary nesting reads badly on narrow screens and complicates the query and the UI for very little gain.

Replying to a *reply* is **not** an error — the new comment attaches to that reply's parent instead. That is what the person meant, and it keeps the thread flat without making them retry.

**Deletion is soft.** The row is tombstoned so replies keep their parent and the thread does not develop holes. The API returns `body: "[deleted]"`, `author: null`, `is_deleted: true`, and never sends the original text again. A deleted **top-level** comment stays in the listing because its replies still need the context; a deleted **reply** has nothing hanging off it, so it is dropped entirely.

**A denied read is 404, not 403.** Confirming that a private project exists is itself a disclosure.

**Edit is stricter than delete.**

| Action | Who |
| --- | --- |
| Post | Any signed-in user who can view the project |
| Edit | The author, and only the author |
| Delete | The author, the project owner, or an admin |

Removing someone's words is moderation; rewriting them under their name is a different thing, so not even the project owner can edit. There is a test asserting exactly that.

### Backend

- `models/project_comment.py` — `project_comments` with self-referential `parent_id`, soft-delete columns, and composite indexes on `(project_id, created_at)` and `(parent_id, created_at)` matching the two real queries.
- `alembic/versions/d4e5f6a7b8c9_…` — the migration.
- `schemas/project_comment.py` — body trimmed then bounded to 1–5000 characters. `min_length=1` alone lets a single space through, which renders as an empty bubble. The author payload is deliberately narrow so a comment list is not a way to enumerate email addresses.
- `services/project_comment_service.py` — visibility, permissions, and the threading rules. Replies are eager-loaded with `selectinload`, one extra query per page rather than one per comment.
- `routers/project_comments.py` — `GET`, `GET /count`, `POST`, `PATCH /{id}`, `DELETE /{id}`.

`total` in a listing counts **top-level comments only**, so it agrees with the page size; `/count` includes replies, which is the number a "12 comments" badge wants.

### A pre-existing bug this ran straight into

When a Pydantic `field_validator` rejects a value by raising `ValueError`, Pydantic v2 puts the exception **object** into the error entry's `ctx`:

```python
{'type': 'value_error', 'loc': (...), 'msg': '...', 'ctx': {'error': ValueError('...')}}
```

`validation_exception_handler` passed that straight to `JSONResponse`:

```
TypeError: Object of type ValueError is not JSON serializable
```

So the handler **crashed while rendering the response** instead of returning a 422. This affects every endpoint whose schema uses a custom validator — `schemas/skill.py` and `schemas/user.py` already have some, so it is live today, not something this PR introduces.

Probing my own endpoint before the fix:

```
missing field: 422        ← plain Pydantic error, fine
too long:      422        ← plain Pydantic error, fine
blank body:    RAISED TypeError   ← custom validator
```

Validation error entries are now sanitised before serialising — `loc` coerced to strings, non-primitive `ctx` and `input` values stringified rather than dropped, so the detail stays useful. All three cases now return 422. Two regression tests in `test_error_responses.py`, both of which fail against `main` with the exact `TypeError`.

### Frontend

- `api/modules/projectComments.ts` — `list`, `count`, `create`, `update`, `remove`. `list`/`count` take an `AbortSignal` so a navigation can cancel in flight. `create` always sends `parent_id` explicitly so a comment and a reply have the same body shape.
- `components/project/ProjectComments.tsx` — thread with compose, reply, edit, delete. Because deletion is soft, every mutation invalidates and refetches rather than removing the row locally — the tombstone has to come back from the server.
- `docs/project_discussions.md` — model, API, permission matrix, validation rules.

---

## Testing

**61 backend tests** (`test_project_comments.py` 52 + `test_error_responses.py` 9), all passing. The service tests run against the real session rather than mocks, because most of what matters here is relational:

- Threading: replies attach correctly, a reply-to-a-reply reparents, a `parent_id` from another project is a 404, replying to a deleted comment is a 404.
- Visibility: public/anonymous, private/anonymous, private/non-member, private/owner, private/active-member, private/**inactive**-member, private/admin, and that denial is 404.
- Editing: author only; project owner explicitly cannot; editing to identical text does not set the edited marker.
- Deleting: author, project owner, admin, and a stranger rejected; the row survives; double-delete keeps the first timestamp; replies survive their parent.
- Listing: top-level only, newest first, replies oldest first, deleted replies dropped, deleted parents retained, other projects excluded, pagination, and `limit` clamped at 100.
- HTTP layer: anonymous read allowed, anonymous post rejected, 201 shape, 404 on unknown project, 422 on a blank body, 403 on editing someone else's, 204 on delete with the body tombstoned — including asserting the original text appears nowhere in the listing response.

**9 frontend tests** for the API module — paths, query params, abort signal, and the `parent_id` normalisation.

**Migration, against Postgres 16** (applied on top of #931 so the graph resolves):

```
Running upgrade c9a1f2b3d4e5 -> d4e5f6a7b8c9, create project comments table
columns: id project_id author_id parent_id body is_edited edited_at deleted_at deleted_by_id created_at updated_at
indexes: ix_project_comments_author_id, ix_project_comments_deleted_at,
         ix_project_comments_parent_created, ix_project_comments_parent_id,
         ix_project_comments_project_created, ix_project_comments_project_id
```

Downgrade drops it cleanly and re-upgrade recreates it. `check_migration_graph` reports 1 head, 1 base, 42 revisions.

**No regressions.** Backend failure list diffed against `main`: identical 94 pre-existing failures, nothing new. Frontend `tsc --noEmit`: 34 errors on this branch and 34 on `main` — none of them mine. ESLint and Prettier clean on every file added. `ruff`/`black` at the pinned versions clean, except two findings that are byte-identical on `main` (`E741` in the untouched part of the error handler, an unused `limits.parse` import in the test file).

- [x] Tested locally
- [x] Existing tests pass
- [x] Added new tests
- [ ] UI verified

---

## Checklist

---

## Additional Notes

> ### Merge #931 first
>
> The migration's `down_revision` is `c9a1f2b3d4e5`, the merge revision added by **#931**. That is deliberate: `main` currently has **two** Alembic heads and `alembic upgrade head` does not resolve at all, so basing this on either existing head would leave the graph at two heads either way. Descending from the merge point means that once both land there is exactly one head.
>
> **Merging this before #931 will break `main`'s migration graph** with an unknown-revision error. If you would rather take this one first, say so and I will repoint `down_revision` and add a merge revision here instead.

**Not included, deliberately:**

- **No notifications on reply.** `NotificationType` has no comment member, and adding one means an enum migration plus a preferences toggle. Better as a follow-up than smuggled into this diff.
- **No `@mention` parsing.** #333 covers mentions; this stores plain text and the component renders it with `whitespace-pre-wrap`, so there is nothing to unpick when that lands.
- **No `comment_count` denormalised onto `projects`.** `/count` is a single indexed aggregate; a cached column is worth adding when a listing page needs it, not before.

**Rendering is plain text.** The component does not interpret Markdown or HTML, so there is no injection surface. If rich text is wanted later, it should go through the existing editor from #492 with sanitisation, not a raw `dangerouslySetInnerHTML`.


nensii21:
image
kindly resolve it.

nensii21:
@MOHITKOURAV01 frontend job hasn't been resolved.

PR #1045, opened 2026-08-10
feat(profile): add contribution streaks and activity heatmap (#1040)

### LINKED ISSUE #1040: Add contribution streaks and an activity heatmap to builder profiles
## Problem

A DevLink profile currently shows *what* a builder has done, but nothing about *how consistently* they show up. We already write every meaningful action into the `activities` table (`ActivityType.PROJECT_CREATED`, `COMMENT_CREATED`, `APPLICATION_SUBMITTED`, ...), but that data is only ever replayed as a reverse-chronological feed. There is no way to answer "has this person been active for the last three weeks, or did they do everything in one burst nine months ago?"

`grep -rn "streak\|heatmap" backend/app frontend/src` returns nothing, so this is not implemented anywhere today.

## Proposal

Derive a contribution heatmap and streak summary from the existing `activities` table — no new writes, no duplicated bookkeeping.

**Backend**

- `ActivityHeatmapService` that buckets a user's activities into UTC calendar days over a requested window (default 365 days, max 366).
- Streak maths on those buckets:
  - `current_streak` — consecutive active days ending today *or* yesterday (so a streak is not broken mid-day by the user's timezone).
  - `longest_streak` and the date range it covers.
  - `total_activities`, `active_days`, `busiest_day`, `daily_average`.
- Intensity levels 0–4 per day, computed from the window's own distribution (quartiles) rather than hardcoded thresholds, so a quiet user's heatmap is still readable.
- Optional `activity_types` filter so the caller can look at only, say, project + comment activity.
- Endpoints:
  - `GET /api/v1/users/{username}/activity-heatmap?days=365`
  - `GET /api/v1/users/me/activity-heatmap`
- Respect the existing privacy rules: a user whose profile is not public should not expose their heatmap to strangers.

**Frontend**

- Typed API client + a `ContributionHeatmap` component (52-week grid, month labels, weekday labels, tooltip per cell) rendered on the profile page.

## Acceptance criteria

- [ ] Days with zero activity are present in the response with `count: 0` — the grid must not have holes.
- [ ] A single-day gap breaks `current_streak`; activity today or yesterday both keep it alive.
- [ ] `longest_streak` is correct when the longest run is at the very start or very end of the window.
- [ ] `days` is validated (1–366) and rejected with 422 outside that range.
- [ ] Unknown username returns 404.
- [ ] Unit tests cover streak edges: empty history, single day, all days active, gap at boundaries.

## Out of scope

Repository-level GitHub contribution graphs (that is #979) — this is DevLink's own activity only.

MOHITKOURAV01:
Closes #1040

## What this does

Adds a contribution heatmap and streak summary to builder profiles, derived entirely from the `activities` table we already write. No new writes, no new table, no duplicated bookkeeping — which means the heatmap is correct retroactively for existing users and can never drift out of sync with the feed it summarises.

## Backend

`ActivityHeatmapService` buckets a user's activity into UTC calendar days over a 1–366 day window and returns:

- **The grid** — one entry per day, ascending, *including* days with zero activity. The client should not have to fill gaps to draw a calendar.
- **Streak stats** — current streak, longest streak with its date range, active days, busiest day, daily and per-active-day averages.
- **A type breakdown** — activity counts grouped by `ActivityType`, descending.

Endpoints:

```
GET /api/v1/users/{username}/activity-heatmap?days=365&activity_types=project_created,comment_created
GET /api/v1/users/me/activity-heatmap
```

Also mounted under `/api/users/...` alongside the rest of the unversioned surface in `main.py`.

## Three decisions worth calling out

**Bucketing happens in Python, not in SQL.** The obvious implementation is `GROUP BY date_trunc('day', created_at)`. The problem is that Postgres (production) and SQLite (pytest) disagree about which timezone a timestamp-to-date cast resolves in, so the same data can land on different days in different environments. A heatmap that silently shifts by one day depending on where it runs is a nasty bug to chase. The window is capped at 366 days for one user, so pulling two scalar columns and bucketing them is cheap, and it is unambiguously UTC.

**Intensity levels are quartiles of the user's own distribution, not fixed thresholds.** Hardcoding 1/3/6/9 makes a casual contributor's entire year render as one flat shade. Deriving the three cut points from the non-zero counts in the window keeps the gradient meaningful whether someone logs two things a week or forty things a day. Thresholds are clamped non-decreasing and to a minimum of 1, so a user whose every active day has exactly one activity lands cleanly on level 1 rather than being split arbitrarily.

**`current_streak` anchors on yesterday when today is empty.** Otherwise every streak on the platform would read as broken from midnight until the user's first action of the day — which is both wrong and, for a feature whose whole point is encouragement, actively counterproductive.

## Privacy

A heatmap shows when someone works and when they stopped, which is about as revealing as the profile itself. Private profiles therefore follow the same rule the profile endpoint already uses: owner only, 403 for everyone else. `days` is validated at 1–366 and an unknown `activity_types` value is a 422 rather than a silent widening to "everything" — a typo that quietly changes the meaning of the numbers is the kind of thing nobody notices until the numbers are wrong.

## Frontend

- `activityHeatmapApi` typed client.
- Grid layout maths (`buildWeeks`, `buildMonthLabels`, `formatDayLabel`) extracted to `src/lib/heatmapGrid.ts` so it is testable on its own and the component file exports components only.
- `ContributionHeatmap` card — 53-column grid, month and weekday labels, per-cell tooltip and `aria-label`, streak stat row, Less→More legend. Month labels are suppressed when they would sit within two columns of the previous one, which otherwise smears them together on short windows.
- Rendered on `/profile/$username`.

## Tests

**48 backend cases**, focused on the parts that are easy to get subtly wrong:

- Streaks: empty history, single day, run ending today, run ending yesterday, one-day gap, gap mid-run, longest run at the very start and very end of the window, ties, every-day-active.
- Levels: monotonic in count, non-decreasing thresholds, uniform activity, busiest day reaching level 4.
- Bucketing: two activities on the same UTC day, naive timestamps treated as UTC, rows outside the window discarded.
- Access: private profile 403 for anonymous/other/allowed-for-owner.
- HTTP: 404 unknown user, 422 out-of-range `days`, auth required on `/me`.

**10 frontend cases** for the grid maths: leading and trailing padding, Sunday-start windows, every column exactly seven cells, no day lost or duplicated across 365 days, month labels unique and never adjacent.

```
backend:  48 passed
frontend: 10 passed
tsc --noEmit: clean   eslint: clean
```

## Notes for review

- `tests/test_activities.py` has 3 pre-existing failures on `main` (unrelated to this change — they fail identically without this branch applied).
- No migration: this feature reads existing rows only.


nensii21:
@MOHITKOURAV01 attach screenshots of the changes that have beeb made related frontend.

PR #1046, opened 2026-08-10
feat(projects): add time tracking so owners can see where effort goes (#1041)

### LINKED ISSUE #1041: Add project time tracking so owners can see where effort actually goes
## Problem

Projects on DevLink have milestones (#618), members with roles, and applications — but no notion of effort. A project owner cannot answer basic questions that matter when running a volunteer team:

- How many hours went into this project last month?
- Which milestone is eating all the time?
- Is a contributor logging work at all, or have they gone quiet?

`grep -rn "time_log\|time_entry\|timesheet" backend/app` returns nothing.

## Proposal

A lightweight work-log ledger scoped to a project, deliberately *not* a full timesheet product.

**Data model** — `project_time_logs`

| column | notes |
| --- | --- |
| `project_id` | FK, `ON DELETE CASCADE` |
| `user_id` | FK, the person who did the work |
| `milestone_id` | nullable FK to `project_milestones`, `ON DELETE SET NULL` |
| `minutes` | integer, 1 … 1440 — stored as minutes to dodge float drift |
| `work_date` | `Date`, the day the work happened (not the day it was logged) |
| `description` | short text, optional |
| `is_billable` | boolean, default false |
| `created_at` / `updated_at` | |

Indexes on `(project_id, work_date)` and `(user_id, work_date)` — those are the two queries that matter.

**Rules**

- Only active project members (or the owner) may log time against a project.
- A member may edit/delete **their own** entries; project maintainers may delete anyone's.
- `work_date` may not be in the future, and may not be more than 90 days in the past (stops backfill nonsense).
- A single entry is capped at 24h; the service also rejects a day whose total would exceed 24h for that user on that project.

**Endpoints** (`/api/v1/projects/{project_id}/time-logs`)

- `POST ""` — log time
- `GET ""` — list, filterable by `user_id`, `milestone_id`, `from_date`, `to_date`, paginated
- `PATCH /{log_id}` / `DELETE /{log_id}`
- `GET /summary` — totals for the project: overall hours, per-contributor breakdown, per-milestone breakdown, billable vs non-billable
- `GET /me` — the caller's own entries across the project

## Acceptance criteria

- [ ] Non-member gets 403 when logging time.
- [ ] Future `work_date` → 422.
- [ ] Exceeding 24h/day for a user on a project → 400 with a clear message.
- [ ] Deleting a milestone leaves its time logs intact with `milestone_id = NULL`.
- [ ] Summary hours are rounded to two decimals and reconcile exactly with the sum of `minutes`.
- [ ] Alembic migration up **and** down.

MOHITKOURAV01:
Closes #1041

## What this does

Adds a work-log ledger scoped to a project, so an owner can answer the questions that actually come up when running a volunteer team: how many hours went into this last month, which milestone is eating the time, and has a contributor gone quiet.

Deliberately *not* a timesheet product — no timers, no approvals, no invoicing. One table, one set of rules.

## Storage decisions

**Minutes, not hours-as-float.** Hours are a presentation concern. Storing `1.5` and summing hundreds of those is how a summary ends up disagreeing with its own line items by a fraction of an hour, and then someone has to work out which number is the real one. `minutes` is an integer end to end; `to_hours()` rounds once, at the response edge. There is a test that pins this: twenty 20-minute entries sum to exactly 400 minutes, and rounding each one to hours first would not.

**`work_date` is a `Date`, separate from `created_at`.** People log Friday's work on Monday morning. Conflating "when it happened" with "when it was typed" makes every weekly report wrong at the boundary.

**A deleted milestone `SET NULL`s its logs instead of cascading.** The milestone can go away; the work still happened. Those entries then group under a null `milestone_id` in the summary rather than silently vanishing from the project total.

Three composite indexes: `(project_id, work_date)` for the project report, `(user_id, work_date)` for a person's history, and `(project_id, user_id, work_date)` for the daily-cap check, which runs on every write.

## Rules

- **Only active members (or the owner) may log time.** Logs feed contributor-level reporting, so an outsider adding rows would be writing into other people's numbers.
- **Edit and delete are split.** The author may edit their own entries. Maintainers may *delete* anyone's — a project needs to clean up after a departed member — but may not *rewrite* them. An edited entry still carrying someone else's name is worse than no entry at all.
- **`work_date` may not be in the future**, nor more than 90 days back. Past that point the number stops being a recollection and starts being a guess.
- **24 hours per user, per project, per day.** Enforced across entries, not just within one. Crucially it is re-checked whenever *either* the duration or the date changes on an edit — moving an existing entry onto an already-full day is exactly as much a violation as growing it in place, and it is the case that a naive implementation misses. The check excludes the entry being edited from its own total, otherwise resizing a 600-minute entry would be measured against a day that still contains the old 600.
- A `CHECK` constraint in the migration backs the per-entry bounds at the database level too.

## Endpoints

```
POST   /api/v1/projects/{id}/time-logs
GET    /api/v1/projects/{id}/time-logs        ?user_id&milestone_id&from_date&to_date&is_billable
GET    /api/v1/projects/{id}/time-logs/me
GET    /api/v1/projects/{id}/time-logs/summary ?from_date&to_date
GET    /api/v1/projects/{id}/time-logs/{log_id}
PATCH  /api/v1/projects/{id}/time-logs/{log_id}
DELETE /api/v1/projects/{id}/time-logs/{log_id}
```

`total` and `total_minutes` on the list response cover the whole filtered set, not the returned page. A report that only adds up what happens to be visible is a report nobody can trust.

## Migration

`e7c1a9b45d20`, branching from the current head `d4e5f6a7b8c9`. `alembic heads` is a single head with this applied. Downgrade drops the indexes and table cleanly.

## Tests

48 cases. The ones worth reading:

- **Cap boundary** — exactly at 24h passes, one minute over is a 400 with the remaining budget in the message, a full day reports "0 minutes remain", and the edit path excludes the entry from its own total.
- **Backfill window** — day 90 is accepted, day 91 is a 422.
- **Permission split** — a maintainer can delete but not edit someone else's entry; a contributor can do neither; the author can always do both. After a rejected edit the test re-reads the entry and asserts it is unchanged.
- **Summary reconciliation** — `billable + non_billable == total`, and both the per-contributor and per-milestone breakdowns sum back to the project total, including the null-milestone group.
- **Pagination** — a 2-item page still reports `total: 5` and the full `total_minutes`.

```
48 passed
```

### DIFF
--- a/backend/alembic/versions/e7c1a9b45d20_create_project_time_logs_table.py
+++ b/backend/alembic/versions/e7c1a9b45d20_create_project_time_logs_table.py
@@ -1,7 +1,7 @@
 """create project time logs table
 
 Revision ID: e7c1a9b45d20
-Revises: d4e5f6a7b8c9
+Revises: f2a8c61d97b5
 Create Date: 2026-08-10 20:10:00.000000
 
 """
@@ -14,7 +14,7 @@ from sqlalchemy.dialects.postgresql import UUID
 
 # revision identifiers, used by Alembic.
 revision: str = "e7c1a9b45d20"
-down_revision: Union[str, Sequence[str], None] = "d4e5f6a7b8c9"
+down_revision: Union[str, Sequence[str], None] = "f2a8c61d97b5"
 branch_labels: Union[str, Sequence[str], None] = None
 depends_on: Union[str, Sequence[str], None] = None
 
--- a/backend/alembic/versions/e7c1a9b45d20_create_project_time_logs_table.py
+++ b/backend/alembic/versions/e7c1a9b45d20_create_project_time_logs_table.py
@@ -1,0 +1,120 @@
+"""create project time logs table
+
+Revision ID: e7c1a9b45d20
+Revises: d4e5f6a7b8c9
+Create Date: 2026-08-10 20:10:00.000000
+
+"""
+
+from typing import Sequence, Union
+
+import sqlalchemy as sa
+from alembic import op
+from sqlalchemy.dialects.postgresql import UUID
+
+# revision identifiers, used by Alembic.
+revision: str = "e7c1a9b45d20"
+down_revision: Union[str, Sequence[str], None] = "d4e5f6a7b8c9"
+branch_labels: Union[str, Sequence[str], None] = None
+depends_on: Union[str, Sequence[str], None] = None
+
+
+def upgrade() -> None:
+    op.create_table(
+        "project_time_logs",
+        sa.Column("id", UUID(as_uuid=True), nullable=False),
+        sa.Column("project_id", UUID(as_uuid=True), nullable=False),
+        sa.Column("user_id", UUID(as_uuid=True), nullable=False),
+        # A milestone can be deleted while the work that went into it still
+        # happened, so the log survives the milestone rather than cascading.
+        sa.Column("milestone_id", UUID(as_uuid=True), nullable=True),
+        # Whole minutes, not hours-as-float: summing floats across hundreds of
+        # entries lets a summary drift away from its own line items.
+        sa.Column("minutes", sa.Integer(), nullable=False),
+        # The day the work happened, which is often not the day it was entered.
+        sa.Column("work_date", sa.Date(), nullable=False),
+        sa.Column("description", sa.String(length=500), nullable=True),
+        sa.Column(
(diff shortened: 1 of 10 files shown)


MOHITKOURAV01:

```
```

```bash
cd backend
```

```python
```

nensii21:
@MOHITKOURAV01 Which tool did you use for coding?

PR #1047, opened 2026-08-10
feat(profile): back pinned projects with a real API (#1042)

### LINKED ISSUE #1042: Pinned projects on profiles are hardcoded mock data with no backend
## Problem

The profile page already renders a "Pinned Projects" section:

```
frontend/src/routes/_app.profile.$username.tsx:539
  {b.pinnedProjects?.length ? (
    <p ...>Pinned Projects</p>
```

…but `pinnedProjects` comes from `frontend/src/mocks/seed.ts`. It is a hardcoded array of two strings. There is no table, no endpoint, and no way for a real user to pin anything. Every profile shows the same two fake pins or nothing at all.

## Proposal

Make pinning real.

**Data model** — `pinned_projects`

- `user_id` FK → `users.id` (CASCADE)
- `project_id` FK → `projects.id` (CASCADE)
- `position` smallint, 0-based ordering
- `created_at`
- `UNIQUE (user_id, project_id)` and `UNIQUE (user_id, position)` — the second one is deferred-unfriendly on Postgres during reorders, so the service rewrites positions in one statement rather than row-by-row.

**Rules**

- Maximum **6** pins per user.
- You may only pin a project you own or are an active member of.
- You may not pin a project that is archived, soft-deleted, or `visibility = private` unless you own it — a pin is a public shop-window, and a stranger following the pin must be able to open it.
- Unpinning compacts the remaining positions (no gaps).

**Endpoints**

- `GET /api/v1/users/{username}/pinned-projects` — public, ordered by `position`
- `GET /api/v1/users/me/pinned-projects`
- `POST /api/v1/users/me/pinned-projects` — `{project_id}`, appends at the end
- `PUT /api/v1/users/me/pinned-projects` — `{project_ids: [...]}`, replaces the whole set in the given order (this is what a drag-and-drop UI wants)
- `DELETE /api/v1/users/me/pinned-projects/{project_id}`

**Frontend**

Typed API client, and swap the profile page's mock read for the real endpoint so pins render as project cards that link through to the project.

## Acceptance criteria

- [ ] Pinning a 7th project → 400 with the limit in the message.
- [ ] Pinning the same project twice → 409, not a duplicate row.
- [ ] Pinning someone else's private project → 403.
- [ ] `PUT` with a project the user cannot pin rejects the **whole** batch (no partial write).
- [ ] After unpinning the middle of three pins, remaining positions are `0, 1`.
- [ ] `pinnedProjects` is gone from `mocks/seed.ts` consumers on the profile route.

MOHITKOURAV01:
Closes #1042

## The problem

`_app.profile.$username.tsx` has rendered a "Pinned Projects" card since it was written:

```tsx
{b.pinnedProjects?.length ? (
  <Card className="p-4">
    <p ...>Pinned Projects</p>
```

`pinnedProjects` came from `mocks/seed.ts` — a hardcoded `["AI Chatbot", "DevOps Dashboard"]` on exactly one seeded builder. Every profile on the platform showed either those same two fake strings or nothing at all, and there was no table, no endpoint, and no way for a real user to pin anything. The strings were not even links.

This makes it real.

## Data model

`pinned_projects` — `user_id`, `project_id`, `position` (0-based, contiguous), `created_at`. Unique on `(user_id, project_id)`, indexed on `(user_id, position)` for the read path.

**There is deliberately no unique constraint on `(user_id, position)`.** It looks like the obvious thing to add, and it is a trap: any reorder has to pass through intermediate states where two rows briefly share a slot, and a non-deferrable unique index rejects the write halfway through. Positions are instead owned by the service, which compacts on every removal so the sequence stays `0..n-1`. A sparse sequence would leak out through the API and force every client to renumber for itself. `(user_id, project_id)` *is* enforced in the database, because unlike position it never legitimately collides.

## The rules, and why

A pin is a public shop-window, not a project list. Everything else follows from that:

- **You must own or be an active member of the project.** You cannot decorate your profile with work you had no part in.
- **Private projects cannot be pinned, even by their owner.** A visitor who clicks a pin and hits a 403 has been shown a dead link. If you want it on your profile, make it public.
- **Archived projects cannot be pinned.**
- **Ownership is checked before visibility.** A stranger enumerating project ids gets "you can only pin projects you own", not a message that confirms the id exists and happens to be private.
- **Cap of 6** — enough to show range, few enough that a visitor reads all of them.

## Endpoints

```
GET    /api/v1/users/{username}/pinned-projects   public, ordered
GET    /api/v1/users/me/pinned-projects
POST   /api/v1/users/me/pinned-projects           {project_id}, appends
PUT    /api/v1/users/me/pinned-projects           {project_ids: [...]}, replaces in order
DELETE /api/v1/users/me/pinned-projects/{project_id}
```

`PUT` is the one a drag-and-drop UI wants: one call, one final state, nothing for the client to reconstruct. **The whole batch is validated before a single row is touched.** A partial write here is particularly bad — it leaves the profile in an order the user never asked for and, since the request appeared to fail, would not think to go and check.

## Frontend

- `pinnedProjectsApi` typed client.
- `PinnedProjectsCard` — real project cards with tagline, stage chip, star and view counts, each linking through to the project. Replaces the mock block on the profile route.
- The card **hides itself entirely** for visitors when nothing is pinned, rather than growing an empty card on every profile that has not used the feature. The owner gets a prompt instead, because they are the only one who can act on it.
- `pinnedProjects` deleted from the `Builder` type and the seed fixture, so the mock cannot quietly come back.

## Migration

`b6d3f2a71c48`, branching from `d4e5f6a7b8c9`. Single head. `CHECK (position >= 0)` and the unique constraint are enforced at the database level; downgrade drops cleanly.

## Tests

28 backend cases:

- **The cap** — six pins succeed, the seventh is a 400 naming the limit.
- **Duplicate** — second pin of the same project is a 409, and the listing still shows exactly one row.
- **Eligibility** — private rejected, archived rejected, someone else's project rejected, ownership checked ahead of visibility.
- **Compaction** — after unpinning the middle of three, positions are `[0, 1]` and the surviving slugs are the right two.
- **Batch atomicity** — a `PUT` containing one ineligible project returns 403 *and the previous single pin is still there, unchanged*. That assertion is the point of the test.
- **Reorder** — `PUT` with a rotated list produces exactly that order at positions 0, 1, 2.
- Empty list clears pins; duplicates in the batch are a 400; over-limit is a 422 at the schema.

```
backend:  28 passed
frontend: 468 passed (full suite)
tsc --noEmit: clean   eslint: clean
```

### DIFF
--- a/backend/alembic/versions/b6d3f2a71c48_create_pinned_projects_table.py
+++ b/backend/alembic/versions/b6d3f2a71c48_create_pinned_projects_table.py
@@ -1,0 +1,64 @@
+"""create pinned projects table
+
+Revision ID: b6d3f2a71c48
+Revises: d4e5f6a7b8c9
+Create Date: 2026-08-10 20:40:00.000000
+
+"""
+
+from typing import Sequence, Union
+
+import sqlalchemy as sa
+from alembic import op
+from sqlalchemy.dialects.postgresql import UUID
+
+# revision identifiers, used by Alembic.
+revision: str = "b6d3f2a71c48"
+down_revision: Union[str, Sequence[str], None] = "d4e5f6a7b8c9"
+branch_labels: Union[str, Sequence[str], None] = None
+depends_on: Union[str, Sequence[str], None] = None
+
+
+def upgrade() -> None:
+    op.create_table(
+        "pinned_projects",
+        sa.Column("id", UUID(as_uuid=True), nullable=False),
+        sa.Column("user_id", UUID(as_uuid=True), nullable=False),
+        sa.Column("project_id", UUID(as_uuid=True), nullable=False),
+        # 0-based and contiguous, compacted by the service on every removal.
+        sa.Column("position", sa.Integer(), nullable=False, server_default=sa.text("0")),
+        sa.Column(
+            "created_at",
+            sa.DateTime(timezone=True),
+            server_default=sa.text("now()"),
+            nullable=False,
+        ),
+        sa.ForeignKeyConstraint(["user_id"], ["users.id"], ondelete="CASCADE"),
+        sa.ForeignKeyConstraint(["project_id"], ["projects.id"], ondelete="CASCADE"),
+        sa.PrimaryKeyConstraint("id"),
+        # One pin per project per user. Unlike position this never
+        # legitimately collides, so it is safe to enforce here.
+        sa.UniqueConstraint("user_id", "project_id", name="uq_pinned_projects_user_project"),
+        sa.CheckConstraint("position >= 0", name="ck_pinned_projects_position"),
+    )
+
+    op.create_index(
+        op.f("ix_pinned_projects_user_id"), "pinned_projects", ["user_id"], unique=False
+    )
+    op.create_index(
+        op.f("ix_pinned_projects_project_id"), "pinned_projects", ["project_id"], unique=False
+    )
+    # The read path: one user's pins in display order.
+    op.create_index(
+        "ix_pinned_projects_user_position",
+        "pinned_projects",
+        ["user_id", "position"],
+        unique=False,
+    )
+
+
+def downgrade() -> None:
+    op.drop_index("ix_pinned_projects_user_position", table_name="pinned_projects")
+    op.drop_index(op.f("ix_pinned_projects_project_id"), table_name="pinned_projects")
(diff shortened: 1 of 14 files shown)


MOHITKOURAV01:

```
```

```bash
cd backend
```

```python
```

Issue #606, opened 2026-07-27
Implement Project Version History


nensii21:
Description

Maintain a history of edits made to project details, allowing owners to review previous versions and restore changes if needed.

Track Changes
Project title
Description
Tech stack
Team roles
Requirements

Acceptance Criteria
Version history stored efficiently
Users can compare versions
Restore previous version functionality
Audit logs included
Tests added

nayanraj864-cmyk:
i want to work on this assign to me @nensii21

nayanraj864-cmyk:
i working on the same project but other issue i need some time

PR #1048, opened 2026-08-10
feat(projects): add publishable release notes (#1043)

### LINKED ISSUE #1043: Projects have no changelog — add publishable release notes
## Problem

`project_versions` (#606) exists, but it is an *internal edit-history snapshot* — it records that someone changed the description at 14:02. It is not something a project wants to publish.

There is no way for a project to say "v1.2.0 is out, here is what changed." Today that news has nowhere to go: it gets typed into a message thread or a Builder Flare and disappears. Followers of a project have no changelog to read, and a visitor landing on a project page cannot tell whether it shipped last week or died last year.

## Proposal

A published, versioned changelog per project.

**Data model** — `project_releases`

- `project_id` FK (CASCADE), `author_id` FK (SET NULL — a release outlives its author leaving)
- `version` — free-form string, max 50 (`v1.2.0`, `2026.08`, `beta-3`). Unique per project, case-insensitively.
- `title`, `body` (markdown), `highlights` (JSON array of short strings)
- `release_type` — enum `major | minor | patch | prerelease`
- `status` — enum `draft | published`
- `published_at` (nullable), `is_pinned` (one pinned release per project)
- `created_at` / `updated_at`

**Rules**

- Only project owners/maintainers may create, edit, publish, or delete a release. Reuses `ProjectMilestoneService`-style maintainer checks.
- Drafts are visible only to maintainers. Public listing returns published releases only, newest `published_at` first.
- Publishing sets `published_at` once and is idempotent — re-publishing an already-published release does not bump the date.
- Publishing (not drafting) emits `ActivityType.PROJECT_ANNOUNCEMENT` so the release shows up in followers' feeds.
- Pinning a release un-pins the previously pinned one, atomically.

**Endpoints** (`/api/v1/projects/{project_id}/releases`)

- `GET ""` — published releases, paginated (`include_drafts=true` for maintainers)
- `POST ""` — create (defaults to draft)
- `GET /latest` — the newest published release, 404 if none
- `GET /{release_id}` — 404 for a draft when the caller is not a maintainer (not 403; do not leak that a draft exists)
- `PATCH /{release_id}`, `DELETE /{release_id}`
- `POST /{release_id}/publish`, `POST /{release_id}/pin`

## Acceptance criteria

- [ ] Duplicate version within one project → 409, and the check is case-insensitive (`V1.0.0` collides with `v1.0.0`).
- [ ] Same version string in two different projects is fine.
- [ ] Anonymous `GET /{release_id}` on a draft → 404.
- [ ] Publishing twice does not move `published_at`.
- [ ] Pinning release B while A is pinned leaves exactly one pinned release.
- [ ] `highlights` is capped (10 items, 200 chars each) and rejected past that.
- [ ] Activity row is written on publish, not on create.

MOHITKOURAV01:
Closes #1043

## Why this is not `project_versions`

`project_versions` (#606) already exists, and it is a different thing: an internal edit-history snapshot that records someone changed the description at 14:02. It is a diff log, not an announcement.

There is currently nowhere for a project to say "v1.2.0 is out, here is what changed." That news gets typed into a message thread or a Builder Flare and disappears. Followers have no changelog to read, and a visitor landing on a project page cannot tell whether it shipped last week or died last year.

## What this adds

`project_releases` — a version people quote, a markdown body people read, short `highlights` for the collapsed card, a `release_type`, and a publish moment.

```
GET    /api/v1/projects/{id}/releases            ?include_drafts (maintainers)
POST   /api/v1/projects/{id}/releases            defaults to draft
GET    /api/v1/projects/{id}/releases/latest
GET    /api/v1/projects/{id}/releases/{release_id}
PATCH  /api/v1/projects/{id}/releases/{release_id}
POST   /api/v1/projects/{id}/releases/{release_id}/publish
POST   /api/v1/projects/{id}/releases/{release_id}/pin
DELETE /api/v1/projects/{id}/releases/{release_id}
```

## Decisions worth reviewing

**Draft is the default.** Writing a changelog entry and announcing it are two different decisions. Defaulting to published means every half-finished entry and every typo goes out to followers the moment it is saved.

**A draft is a 404 for non-maintainers, not a 403.** This is the one I would push back on if I saw the opposite. A 403 tells an unauthorised caller that the release exists — which leaks both that something unannounced is in the pipeline and its id. 404 is the correct answer to "is there a resource here that you may see."

**`include_drafts=true` is silently downgraded for the public rather than rejected.** A client that always sends the flag still gets a working public listing instead of a 403 it has to special-case.

**Versions are unique per project, compared case-insensitively.** `V1.0.0` and `v1.0.0` are the same release to every human reading the changelog; storing both produces two entries nobody can tell apart. The uniqueness check excludes the row being edited — without that, saving a form that did not touch the version field would fail with a conflict against itself. There is a test for exactly that.

**Publish is idempotent.** Re-publishing an already-published release returns it unchanged rather than moving `published_at` or firing a second announcement. A double-clicked button should not be able to rewrite a release date.

**The activity row is written on publish, never on create.** A draft is by definition not news, and pushing every work-in-progress into followers' feeds would make the feed useless. Recording it is wrapped so a feed failure is logged rather than raised: the release is the user's work, the feed entry is a side effect, and losing the former to save the latter is the wrong trade.

**`author_id` is `SET NULL`.** A release outlives the person who cut it — a maintainer leaving should not erase the project's history.

**Pinning is atomic** and restricted to published releases. Pinning B while A is pinned leaves exactly one pinned release, sorted to the top of the listing.

## Migration

`c8f4a1e93b27`, branching from `d4e5f6a7b8c9`. Creates the `releasetype` and `releasestatus` enums with `checkfirst=True` and drops them on downgrade. Single head.

## Tests

40 cases. The interesting ones:

- **Case-insensitive collision** — `V1.0.0` after `v1.0.0` is a 409, but the same version in two *different* projects is fine.
- **Self-collision** — editing a release and resubmitting its own version succeeds.
- **Draft visibility** — anonymous `GET` of a draft is 404; the maintainer gets 200 for the same id; `include_drafts` returns 0 for the public and 1 for the maintainer.
- **Publish idempotency** — the second publish returns the *same* `published_at`.
- **Announcement timing** — counts `PROJECT_ANNOUNCEMENT` rows before create, after create (unchanged — "a draft is not news yet"), and after publish (+1).
- **Pinning** — after pinning two in sequence exactly one is pinned, it is the second, and it sorts first.
- **`latest`** — 404 when there are none, ignores drafts, returns the newest published.
- **Highlights** — blanks dropped, whitespace stripped, exactly 10 allowed, 11 rejected, 201 chars rejected.

```
40 passed
```

### DIFF
--- a/backend/alembic/versions/c8f4a1e93b27_create_project_releases_table.py
+++ b/backend/alembic/versions/c8f4a1e93b27_create_project_releases_table.py
@@ -1,0 +1,119 @@
+"""create project releases table
+
+Revision ID: c8f4a1e93b27
+Revises: d4e5f6a7b8c9
+Create Date: 2026-08-10 21:05:00.000000
+
+"""
+
+from typing import Sequence, Union
+
+import sqlalchemy as sa
+from alembic import op
+from sqlalchemy.dialects.postgresql import UUID
+
+# revision identifiers, used by Alembic.
+revision: str = "c8f4a1e93b27"
+down_revision: Union[str, Sequence[str], None] = "d4e5f6a7b8c9"
+branch_labels: Union[str, Sequence[str], None] = None
+depends_on: Union[str, Sequence[str], None] = None
+
+release_type_enum = sa.Enum(
+    "MAJOR", "MINOR", "PATCH", "PRERELEASE", name="releasetype"
+)
+release_status_enum = sa.Enum("DRAFT", "PUBLISHED", name="releasestatus")
+
+
+def upgrade() -> None:
+    bind = op.get_bind()
+    release_type_enum.create(bind, checkfirst=True)
+    release_status_enum.create(bind, checkfirst=True)
+
+    op.create_table(
+        "project_releases",
+        sa.Column("id", UUID(as_uuid=True), nullable=False),
+        sa.Column("project_id", UUID(as_uuid=True), nullable=False),
+        # A release outlives the person who cut it, so the author reference is
+        # nulled rather than cascading the row away.
+        sa.Column("author_id", UUID(as_uuid=True), nullable=True),
+        # Free-form on purpose: v1.2.0, 2026.08 and beta-3 are all legitimate.
+        # Uniqueness per project is enforced case-insensitively in the service.
+        sa.Column("version", sa.String(length=50), nullable=False),
+        sa.Column("title", sa.String(length=200), nullable=False),
+        sa.Column("body", sa.Text(), nullable=True),
+        sa.Column(
+            "highlights", sa.JSON(), server_default=sa.text("'[]'"), nullable=False
+        ),
+        sa.Column("release_type", release_type_enum, nullable=False),
+        sa.Column("status", release_status_enum, nullable=False),
+        sa.Column("published_at", sa.DateTime(timezone=True), nullable=True),
+        sa.Column(
+            "is_pinned", sa.Boolean(), server_default=sa.text("false"), nullable=False
+        ),
+        sa.Column(
+            "created_at",
+            sa.DateTime(timezone=True),
+            server_default=sa.text("now()"),
+            nullable=False,
+        ),
+        sa.Column(
+            "updated_at",
+            sa.DateTime(timezone=True),
+            server_default=sa.text("now()"),
+            nullable=False,
+        ),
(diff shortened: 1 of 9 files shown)


MOHITKOURAV01:

```
```

```bash
cd backend
```

```python
```

Issue #257, opened 2026-07-12
Add Skill Endorsement API


nensii21:
Description

Allow users to endorse skills on other developers' profiles.

Acceptance Criteria

One endorsement per user
Endorsement count returned
Duplicate prevention

Payal29037:
Hi! I'd like to work on this issue as part of ECSoC. Could you please assign it to me?

nensii21:
@Payal29037 no update till now?

PR #1049, opened 2026-08-10
feat(profile): add peer testimonials (#1044)

### LINKED ISSUE #1044: Add peer testimonials so trust on a profile comes from people, not just metrics
## Problem

DevLink has plenty of *machine-generated* trust signals — reputation logs, badges, repository quality scores, profile completion. What it does not have is the one signal that actually persuades a recruiter or a project owner: another human writing a paragraph about what you were like to work with.

`grep -rn "testimonial" backend/app frontend/src` returns nothing.

Note this is **not** #257 (Skill Endorsement API). An endorsement is a +1 on a skill tag. A testimonial is free text with an author, a relationship, and a moderation lifecycle — different table, different rules.

## Proposal

**Data model** — `testimonials`

- `subject_id` FK → `users.id` (CASCADE) — who the testimonial is about
- `author_id` FK → `users.id` (CASCADE) — who wrote it
- `project_id` nullable FK → `projects.id` (SET NULL) — the collaboration it came out of
- `relationship` enum — `collaborator | mentor | mentee | client | teammate`
- `body` text, 50–2000 chars (a two-word testimonial is noise)
- `status` enum — `pending | approved | hidden`
- `is_featured` boolean — subject can feature up to 3
- `responded_at`, `created_at`, `updated_at`
- `UNIQUE (author_id, subject_id)` — one testimonial per pair; editing replaces it

**Rules — the whole point of the feature is that these are right**

- You cannot write a testimonial about yourself → 400.
- If either party has blocked the other (`user_blocks` already exists), the write is rejected → 403.
- A new testimonial lands in `pending`. It is visible to the author and the subject, nobody else.
- Only the **subject** can approve or hide it. The author can never self-approve.
- The author may edit the body while it is `pending`; editing an `approved` testimonial sends it back to `pending` (otherwise approval is a rubber stamp on text that can be swapped out afterwards).
- Featuring requires `approved`, capped at 3.
- Deleting: author can always delete their own; subject can hide but not delete (their opinion is not theirs to erase).

**Endpoints**

- `POST /api/v1/testimonials` — write one
- `GET /api/v1/users/{username}/testimonials` — approved only for the public; subject sees all with `?status=`
- `GET /api/v1/testimonials/received` / `GET /api/v1/testimonials/written` — the caller's own
- `POST /api/v1/testimonials/{id}/approve`, `/hide`, `/feature`
- `PATCH /api/v1/testimonials/{id}`, `DELETE /api/v1/testimonials/{id}`
- `GET /api/v1/users/{username}/testimonials/summary` — counts by relationship, featured list

## Acceptance criteria

- [ ] Self-testimonial → 400.
- [ ] Blocked in either direction → 403.
- [ ] Author approving their own testimonial → 403.
- [ ] A `pending` testimonial is absent from the public listing but present for the subject.
- [ ] Editing an approved testimonial resets `status` to `pending` and clears `is_featured`.
- [ ] Featuring a 4th → 400.
- [ ] Second testimonial from the same author about the same subject → 409.

MOHITKOURAV01:
Closes #1044

## Why

Every trust signal on DevLink is currently machine-generated — reputation logs, badges, repository quality scores, profile completion percentages. None of it is the thing that actually persuades a recruiter or a project owner: another human writing a paragraph about what you were like to work with.

**This is not #257 (Skill Endorsement API).** An endorsement is a +1 on a skill tag — a counter. A testimonial is free text with an author, a stated relationship, and a moderation lifecycle. Different table, different rules, and they compose fine.

## The rules are the feature

A testimonial system where the subject can quietly rewrite what was said about them, or where the author can approve their own praise, is worth *less* than having no testimonials at all: it looks like social proof while carrying none. So the interesting part of this PR is the permission matrix, not the storage.

| | Author | Subject | Everyone else |
|---|---|---|---|
| Write | ✅ | ✗ (self) | — |
| Edit text | ✅ | ✗ 403 | ✗ |
| Approve / hide | ✗ 403 | ✅ | ✗ 403 |
| Feature | ✗ | ✅ | ✗ |
| Delete | ✅ | ✗ 403 | ✗ |
| See while pending | ✅ | ✅ | ✗ 404 |

Specifically:

- **You cannot write about yourself** — 400, and the check runs *before* the block lookup, since it is wrong regardless.
- **A block in either direction rejects the write.** `BlockService.is_blocked` is already bidirectional. A block means these two have opted out of each other, and a testimonial is a channel from one onto the other's profile.
- **Editing the text of an approved testimonial returns it to `pending` and clears `is_featured`.** This is the rule that makes approval mean something. Without it: get something bland approved, then swap the text for anything at all, and it stays live on their profile under their approval.
- **The author may delete; the subject may hide but not erase.** A subject who could delete would be curating rather than receiving. Hiding also un-features — otherwise hiding would remove it from the list while leaving it pinned at the top of the profile.
- **Featuring requires `approved` and caps at 3.** Un-featuring frees a slot.
- **A pending testimonial is a 404, not a 403**, for anyone but the two parties. A 403 confirms that an unapproved testimonial about this person exists, which is precisely what moderation hides.
- **`?status=` is honoured for the subject and ignored for everyone else**, so the public listing cannot be coaxed into showing the queue.

**The 50-character body floor is deliberate.** "Great dev!" carries no information and dilutes the testimonials that do. Whitespace padding does not get you over the line — the validator strips first, then measures.

## Data model

`testimonials` — `subject_id`, `author_id`, nullable `project_id` (`SET NULL`; the testimonial is about the person, not the project), `relationship` enum, `body`, `status` enum, `is_featured`, `responded_at`.

`UNIQUE (author_id, subject_id)` — one per pair. Someone with an opinion worth stating twice can edit the one they already wrote. A `CHECK (author_id <> subject_id)` backs the self-testimonial rule at the database level too.

One wrinkle worth flagging for review: the column is `relationship`, which is also SQLAlchemy's own name in that module's namespace, so the mapped attribute is `relationship_type` with `name="relationship"` and a small read-only `relationship` property bridging back for the response schema. It is commented in place.

## Endpoints

```
POST   /api/v1/testimonials
PATCH  /api/v1/testimonials/{id}
DELETE /api/v1/testimonials/{id}
POST   /api/v1/testimonials/{id}/approve
POST   /api/v1/testimonials/{id}/hide
POST   /api/v1/testimonials/{id}/feature       ?featured=false to un-feature
GET    /api/v1/testimonials/received           ?status=
GET    /api/v1/testimonials/written
GET    /api/v1/testimonials/{id}
GET    /api/v1/users/{username}/testimonials    ?status= (subject only)
GET    /api/v1/users/{username}/testimonials/summary
```

## Migration

`d9e2b7c4f183`, branching from `d4e5f6a7b8c9`. Creates `testimonialrelationship` and `testimonialstatus` with `checkfirst=True`, drops them on downgrade. Single head.

## Tests

41 cases, weighted towards the permission matrix:

- Author approving their own → 403, **and the testimonial is still absent from the public profile afterwards**.
- Stranger approving → 403. Subject editing the text → 403. Subject deleting → 403.
- Approve, feature, then edit → back to `pending`, `is_featured: false`, and gone from the public listing again.
- Fourth feature → 400 naming the limit; un-featuring one then lets the fourth through.
- Hiding an approved+featured testimonial clears `is_featured`.
- Pending is 404 for a stranger and 200 for its author, same id.
- `?status=pending` on the public endpoint returns 0.
- Self-testimonial 400; second testimonial for the same pair 409; body floor, ceiling, and whitespace-padding rejection.
- Summary counts approved only and groups by relationship.

```
41 passed
```

### DIFF
--- a/backend/alembic/versions/d9e2b7c4f183_create_testimonials_table.py
+++ b/backend/alembic/versions/d9e2b7c4f183_create_testimonials_table.py
@@ -1,0 +1,116 @@
+"""create testimonials table
+
+Revision ID: d9e2b7c4f183
+Revises: d4e5f6a7b8c9
+Create Date: 2026-08-10 21:35:00.000000
+
+"""
+
+from typing import Sequence, Union
+
+import sqlalchemy as sa
+from alembic import op
+from sqlalchemy.dialects.postgresql import UUID
+
+# revision identifiers, used by Alembic.
+revision: str = "d9e2b7c4f183"
+down_revision: Union[str, Sequence[str], None] = "d4e5f6a7b8c9"
+branch_labels: Union[str, Sequence[str], None] = None
+depends_on: Union[str, Sequence[str], None] = None
+
+relationship_enum = sa.Enum(
+    "COLLABORATOR",
+    "MENTOR",
+    "MENTEE",
+    "CLIENT",
+    "TEAMMATE",
+    name="testimonialrelationship",
+)
+status_enum = sa.Enum("PENDING", "APPROVED", "HIDDEN", name="testimonialstatus")
+
+
+def upgrade() -> None:
+    bind = op.get_bind()
+    relationship_enum.create(bind, checkfirst=True)
+    status_enum.create(bind, checkfirst=True)
+
+    op.create_table(
+        "testimonials",
+        sa.Column("id", UUID(as_uuid=True), nullable=False),
+        # Who it is about.
+        sa.Column("subject_id", UUID(as_uuid=True), nullable=False),
+        # Who wrote it.
+        sa.Column("author_id", UUID(as_uuid=True), nullable=False),
+        # The collaboration it came out of. SET NULL because the testimonial is
+        # about the person, not the project.
+        sa.Column("project_id", UUID(as_uuid=True), nullable=True),
+        sa.Column("relationship", relationship_enum, nullable=False),
+        sa.Column("body", sa.Text(), nullable=False),
+        sa.Column("status", status_enum, nullable=False),
+        sa.Column(
+            "is_featured", sa.Boolean(), server_default=sa.text("false"), nullable=False
+        ),
+        # When the subject last approved or hid it.
+        sa.Column("responded_at", sa.DateTime(timezone=True), nullable=True),
+        sa.Column(
+            "created_at",
+            sa.DateTime(timezone=True),
+            server_default=sa.text("now()"),
+            nullable=False,
+        ),
+        sa.Column(
+            "updated_at",
+            sa.DateTime(timezone=True),
+            server_default=sa.text("now()"),
+            nullable=False,
+        ),
+        sa.ForeignKeyConstraint(["subject_id"], ["users.id"], ondelete="CASCADE"),
+        sa.ForeignKeyConstraint(["author_id"], ["users.id"], ondelete="CASCADE"),
+        sa.ForeignKeyConstraint(["project_id"], ["projects.id"], ondelete="SET NULL"),
+        sa.PrimaryKeyConstraint("id"),
+        # One testimonial per author per subject. Someone with an opinion worth
+        # stating twice can edit the one they already wrote.
+        sa.UniqueConstraint("author_id", "subject_id", name="uq_testimonials_author_subject"),
+        # Nobody writes a testimonial about themselves.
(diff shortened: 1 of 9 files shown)


MOHITKOURAV01:

```
```

```bash
cd backend
```

```python
```

PR #1107, opened 2026-08-12
fix(api): stop replaying non-idempotent requests on retry (fixes #1102)

### LINKED ISSUE #1102: API client retries POST/PATCH/DELETE, so a flaky network can create duplicate applications and follows
## Description

`coreFetch` in `frontend/src/api/client.ts` retries on network failure and on any 5xx, but it does not look at the HTTP method first. Every call made through `api.post` / `api.patch` / `api.delete` is therefore replayed up to two extra times.

The retry happens in two places:

```ts
// network / abort path
if (attempt < retries) {
  await new Promise((r) => setTimeout(r, 200 * 2 ** attempt));
  return coreFetch(path, opts, attempt + 1);
}

// 5xx path
if (res.status >= 500 && attempt < retries) {
  await new Promise((r) => setTimeout(r, 200 * 2 ** attempt));
  return coreFetch(path, opts, attempt + 1);
}
```

`retries` defaults to `2`, so the default for a POST is "send it up to three times".

## Why this matters

A retry is only safe when the request is idempotent. Most of our writes are not:

- `applyToFlare()` → `POST /applications`. A 502 from a proxy after the row was already written produces two or three applications for the same flare.
- `followUser()` → `POST /followers/{id}`, and the like/comment endpoints behave the same way.
- Anything that increments a counter server-side gets incremented once per replay.

The dangerous case is not "the server is down" — it is the request that *succeeded* and then failed on the way back (gateway timeout, connection reset after the response was committed, a `timeout` abort that fired while the server was still writing). The client cannot tell that apart from "never arrived", so it sends it again.

## Other problems in the same function

While reading through this, three smaller things stood out:

1. **The post-refresh replay drops the timeout and the caller's abort signal.** After a 401 triggers `refreshAccessToken()`, the retry is a bare `fetch(buildUrl(path, query), { ...init, headers: finalHeaders })` — no `controller.signal`. That request can hang forever, and cancelling the caller's signal will not stop it.

2. **`window` is touched without a guard.** The 503 maintenance branch reads `window.location.pathname` directly. This is a TanStack Start app with SSR (`src/server.ts`), so a 503 from a server-side loader throws `window is not defined` and turns a maintenance page into a 500.

3. **Backoff has no jitter and ignores `Retry-After`.** `200 * 2 ** attempt` is identical on every client, so when the API sheds load every open tab retries in lockstep. If the server sends `Retry-After` on a 429 or 503 we ignore it and hammer it early. Also note 429 is not retried at all today even though it is the one status that explicitly asks you to.

## Suggested fix

- Retry by default only for methods that are idempotent by spec (GET, HEAD, PUT, DELETE) plus anything the caller explicitly opts in with a flag such as `idempotent: true`.
- Let callers that genuinely need a safe retry on POST pass an idempotency key header instead of guessing.
- Honour `Retry-After` (both delta-seconds and HTTP-date forms) and add jitter to the backoff.
- Route the post-refresh replay through the same timeout/abort plumbing as the first attempt.
- Guard the `window` access.

Happy to pick this up.

MOHITKOURAV01:
# Pull Request

## Summary

`coreFetch` retried on network failures and on any 5xx without checking the HTTP method first, so every `api.post` / `api.patch` was replayed up to two extra times. This makes retrying conditional on whether the request is safe to replay, and fixes four related problems in the same code path.

---

## Related Issue

Closes #1102

---

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Refactoring
- [ ] UI/UX enhancement
- [x] Tests
- [ ] Other

---

## Changes Made

**The main fix — retry only what is safe to replay**

- Added `IDEMPOTENT_METHODS` (GET, HEAD, OPTIONS, PUT, DELETE) and gate the retry loop on it. POST and PATCH no longer retry by default.
- Added an `idempotent?: boolean` request option so a caller can override the decision either way: `true` for a POST that is genuinely safe to replay (or that carries an `Idempotency-Key`), `false` to opt an expensive GET out.

The case that motivated this is the one that is easy to miss: a request that *succeeded* and then failed on the way back — gateway timeout, connection reset after the response was committed, our own 10s timeout firing while the server was still writing. The client cannot tell that apart from "never arrived", so replaying it duplicates the write. `applyToFlare`, `followUser` and the like/comment endpoints all sit on that path.

**Same code path, four smaller fixes**

- The replay after a 401 token refresh was a bare `fetch(...)` with no signal, so it ignored `timeout` and could not be cancelled by the caller. Extracted `sendOnce()` so that every attempt, including that one, gets the same timeout and abort plumbing.
- The 503 maintenance branch read `window.location.pathname` unguarded. This client runs during SSR too, so a 503 from a server-side loader threw `window is not defined` and turned a maintenance page into a 500. Now guarded in `redirectToMaintenance()`.
- Backoff was `200 * 2 ** attempt` with no jitter, so every open tab retried in lockstep against an API that was already shedding load. Now full jitter, capped at 8s.
- `Retry-After` is now honoured in both the delta-seconds and HTTP-date forms, capped at 20s so a hostile or misconfigured header cannot park a request indefinitely. 429 is now retried at all — it previously was not, despite being the one status that explicitly asks you to.

**One more thing**

`signal.addEventListener("abort", ...)` was added per attempt and never removed, so a retried request left a listener behind on the caller's signal for each attempt. `sendOnce` now removes it in a `finally`. It also short-circuits when the caller's signal is already aborted, rather than opening a socket and immediately abandoning it.

---

## Testing

- [x] Tested locally
- [x] Existing tests pass
- [x] Added new tests

Added `frontend/src/api/__tests__/client.test.ts` — 32 tests covering:

- POST/PATCH not replayed after a network error, a 502, a 503, or a timeout
- GET/DELETE still retried, on both network errors and retryable statuses
- 429 retried; 404 and other 4xx not retried
- `retries: 0` and the `idempotent` override in both directions
- `idempotent` is not leaked into the request body or headers
- already-aborted signals short-circuit before `fetch`; mid-flight aborts stop the retry loop
- abort listeners are added and removed in equal numbers
- 503 maintenance handling does not throw when `window` is absent
- `parseRetryAfter` across both header forms, past dates, the cap, and malformed input (including `-5`, which `Date.parse` would otherwise read as a year)
- `backoffDelay` growth, jitter floor, and ceiling

Full suite: 514 tests pass, `tsc --noEmit` and `eslint` clean.

---

## Notes for reviewers

This is a behaviour change for POST and PATCH callers that were quietly relying on the retry. I went looking and did not find one that needs it — the POSTs in `src/api/modules` are all writes — but if a reviewer knows of a POST endpoint that is a disguised read, `idempotent: true` is the one-line fix for it.

### DIFF
--- a/frontend/src/api/__tests__/client.test.ts
+++ b/frontend/src/api/__tests__/client.test.ts
@@ -320,3 +320,95 @@ describe("backoffDelay", () => {
     expect(backoffDelay(20)).toBe(8000);
   });
 });
+
+/**
+ * The refresh path only engages when a base URL is configured, so these load a
+ * fresh copy of the module with `VITE_API_BASE_URL` stubbed. Without that,
+ * `refreshAccessToken` short-circuits and the replay never happens — a test
+ * written against the default module passes without exercising anything.
+ */
+describe("the post-refresh replay", () => {
+  let fetchMock: ReturnType<typeof vi.fn>;
+  let client: typeof import("../client");
+
+  beforeEach(async () => {
+    vi.stubEnv("VITE_API_BASE_URL", "https://api.test");
+    vi.resetModules();
+
+    fetchMock = vi.fn();
+    vi.stubGlobal("fetch", fetchMock);
+    vi.spyOn(Math, "random").mockReturnValue(0);
+
+    client = await import("../client");
+    const tokens = await import("../tokens");
+    tokens.tokenStore.set("stale-access", "refresh-token");
+  });
+
+  afterEach(async () => {
+    const tokens = await import("../tokens");
+    tokens.tokenStore.clear();
+    vi.unstubAllEnvs();
+    vi.unstubAllGlobals();
+    vi.restoreAllMocks();
+  });
+
+  it("replays the request once with the refreshed token", async () => {
+    fetchMock
+      .mockResolvedValueOnce(jsonResponse(401, { detail: "Expired" }))
+      .mockResolvedValueOnce(jsonResponse(200, { access_token: "fresh" }))
+      .mockResolvedValueOnce(jsonResponse(200, { ok: true }));
+
+    await expect(client.api.get("/projects", { retries: 0 })).resolves.toEqual({ ok: true });
+
+    const [, replayInit] = fetchMock.mock.calls[2];
+    expect(new Headers(replayInit.headers).get("Authorization")).toBe("Bearer fresh");
+  });
+
+  it("gives the replay a signal, so it can time out and be cancelled", async () => {
+    fetchMock
+      .mockResolvedValueOnce(jsonResponse(401, { detail: "Expired" }))
+      .mockResolvedValueOnce(jsonResponse(200, { access_token: "fresh" }))
+      .mockResolvedValueOnce(jsonResponse(200, { ok: true }));
+
+    await client.api.get("/projects", { retries: 0 });
+
+    // Previously this was a bare fetch with no signal at all.
+    const [, replayInit] = fetchMock.mock.calls[2];
+    expect(replayInit.signal).toBeInstanceOf(AbortSignal);
+  });
+
+  it("surfaces a network failure during the replay as an ApiError", async () => {
+    // The replay sits outside the retry loop, so nothing there would convert a
+    // raw fetch rejection. Callers only ever catch ApiError.
+    fetchMock
+      .mockResolvedValueOnce(jsonResponse(401, { detail: "Expired" }))
+      .mockResolvedValueOnce(jsonResponse(200, { access_token: "fresh" }))
+      .mockRejectedValueOnce(networkError());
+
+    await expect(client.api.get("/projects", { retries: 0 })).rejects.toBeInstanceOf(
+      client.ApiError,
+    );
+  });
+
+  it("surfaces an abort during the replay as a 408", async () => {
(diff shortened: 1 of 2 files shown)


MOHITKOURAV01:
Self-review follow-up pushed.

The replay after a token refresh sits **outside** the retry loop, so nothing there was converting a raw `fetch` rejection — a connection drop during that one request escaped as a bare `TypeError`, and every caller only catches `ApiError`. Now wrapped.

Writing the test for it turned up something worth flagging: the refresh path had **no coverage at all, and could not have**. `refreshAccessToken` short-circuits on `isBackendConfigured()`, and `VITE_API_BASE_URL` is unset under vitest, so any test written against the default module import passes without exercising a single line of it. My first attempt at these two tests passed for exactly that reason before I noticed. The new cases load a fresh copy of the module with the env stubbed, and cover the replay carrying the refreshed token, the replay getting a signal (it previously had none), network and abort failures during it, and the refresh itself being rejected.

Suite is 519 green.

nensii21:
Good catch. The missing refresh-path coverage and the env short-circuit could have made the tests falsely pass. The updated tests covering the refreshed-token replay, signal, network/abort failures, and refresh rejection look solid. 519 green is good.

PR #1108, opened 2026-08-12
fix(pagination): stop repeating the boundary row on every page (fixes #1103)

### LINKED ISSUE #1103: Cursor pagination repeats the boundary row on every page
## Description

`paginate_query` in `backend/app/core/pagination.py` builds its keyset filter with inclusive comparisons:

```python
if is_desc:
    query = query.filter(sort_column <= c_val)
else:
    query = query.filter(sort_column >= c_val)
```

and the cursor it hands out is built from the **last item of the page it just returned**:

```python
last_item = items[-1]
next_data["val"] = str(val) if val is not None else None
```

So the row that ends page N is `<=` the cursor value and comes back as the first row of page N+1. Every page after the first repeats one item.

## Reproduction

With `limit=3` over five rows sorted by `created_at` descending (E, D, C, B, A):

- page 1 → `[E, D, C]`, `next_cursor` encodes C
- page 2 → filter is `created_at <= C` → `[C, B, A]`

C is rendered twice. In an infinite-scroll list that means a visibly duplicated card, and in React a duplicate `key` warning.

It gets worse when `sort_column` is not unique. If several rows share a `created_at` (bulk inserts and seeded data do this constantly), `<=` on the timestamp alone re-emits *all* of the tied rows on the next page, and there is no tiebreaker to make the order stable between the two queries.

## Second problem: the cursor value is stringified

```python
next_data["val"] = str(val) if val is not None else None
```

`val` is usually a `datetime` or a `UUID`. It goes into the cursor as `'2026-08-11 10:54:43.123456+00:00'` and comes back out of `decode_cursor` as a `str`, which is then compared against a `DateTime` column. On Postgres that either raises or silently does a text comparison with different ordering semantics than the index; on SQLite it compares lexicographically and quietly returns the wrong rows. There is no round-trip back to the original type.

## Third problem: `prev_cursor` is offset-shaped

`next_cursor` is keyset-shaped (`{"id": ..., "val": ...}`) but `prev_cursor` is always offset-shaped:

```python
prev_offset = max(0, effective_offset - limit)
prev_cursor = encode_cursor({"offset": prev_offset})
```

When you paginate forward with keyset cursors, `effective_offset` stays `0` the whole way, so `prev_cursor` is `{"offset": 0}` on every page — it always sends you back to page 1 rather than one page back. `has_prev` is also `True` as soon as any cursor is present, including a cursor that decoded to `None` because it was malformed.

## Suggested fix

- Use a strict comparison (`<` / `>`) with a `(sort_column, id_column)` composite tiebreaker so tied sort values are still totally ordered.
- Keep the cursor's type information (encode the value plus a small type tag) and coerce it back before it reaches the filter, instead of `str()`.
- Either make `prev_cursor` a real reverse keyset cursor or drop it for keyset mode rather than returning something that does not work.
- Treat a cursor that fails to decode as a client error instead of silently falling through to page 1.

Add tests for the boundary case, ties on the sort column, and the round-trip of a datetime/UUID cursor — there is no coverage of `paginate_query` today.

MOHITKOURAV01:
# Pull Request

## Summary

`paginate_query` built its keyset filter with inclusive comparisons (`<=` / `>=`) but generated the cursor from the last row of the page it had just returned, so that row was re-emitted as the first row of the next page. This makes the boundary strict, gives the sort a tiebreaker so it is a total order, preserves cursor value types, and replaces the broken `prev_cursor`.

---

## Related Issue

Closes #1103

---

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Refactoring
- [ ] UI/UX enhancement
- [x] Tests
- [ ] Other

---

### 1. The boundary is strict, and the order is total

With `limit=3` over five rows sorted descending (E, D, C, B, A), page 1 was `[E, D, C]` and page 2 was `[C, B, A]`.

The filter is now the standard row-value comparison, written out longhand:

```
sort < :sort  OR  (sort = :sort  AND  id < :id)
```

The tiebreaker matters as much as the strictness. `created_at` is not unique — rows written in the same transaction share a timestamp — so ordering by it alone leaves tied rows in whatever order the planner picks, and it is free to answer differently for the two queries that make up a page boundary. Carrying the primary key makes the order total, so the boundary is stable. `ORDER BY` gained the same tiebreaker.

I wrote the comparison out longhand rather than using `tuple_(a, b) < tuple_(x, y)` because SQLite (which the test suite runs on) does not put an index behind row-value comparisons, and the longhand form is what the query planner wants on Postgres anyway.

### 2. Cursor values keep their type

Values were serialised with `str(val)`, so a `datetime` went into the cursor as `'2026-08-11 10:54:43.123456+00:00'` and came back out of `decode_cursor` as a `str`, which was then compared against a `DateTime` column. Postgres raises on that; SQLite compares lexicographically and quietly returns different rows than the index order implies.

Values now carry a short type tag and are rebuilt on the way back in — `datetime`, `date`, `time`, `UUID`, `Decimal`, `int`, `float`, `bool`, `str`, `None`. Two ordering details that are easy to get wrong and are covered by tests: `bool` is checked before `int` (it is a subclass, and would otherwise round-trip as `0`/`1`) and `datetime` before `date` (same reason).

### 3. `prev_cursor` actually goes back one page

`next_cursor` was keyset-shaped but `prev_cursor` was always `{"offset": ...}` — and since `effective_offset` stays `0` for the whole of a forward keyset walk, it was `{"offset": 0}` on every page. "Previous" always meant "page 1".

It is now a real reverse keyset cursor carrying a direction flag. Walking backwards runs the query in the opposite order, then flips the page back before returning it, so an ascending list stays ascending regardless of which way you arrived.

### 4. A bad cursor is an error

A cursor that fails to decode used to fall through to page 1. That is the worst possible behaviour for an infinite scroll — it restarts from the top and loops. It now raises `InvalidCursor` (a `ValueError` subclass).

Worth flagging for reviewers: this is a behaviour change, but nothing outside the tests calls `paginate_query` today, so there is no blast radius. I have deliberately **not** touched `main.py` to register a 400 handler for it — that file is contended by several open PRs right now and I did not want to add a conflict for a one-line registration. Happy to do it in a follow-up, or here if you would rather.

### 5. `paginate_and_respond()`

Added, because every router that uses this wants `paginate_query` immediately followed by `build_paginated_response`, and threading six positional results between them by hand is how the two halves drift apart.

---

## Testing

- [x] Tested locally
- [x] Existing tests pass
- [x] Added new tests

`backend/tests/test_keyset_pagination.py`, 39 tests. The existing `test_pagination.py` only ever asked for page 1, which is precisely why this bug survived — it is invisible until you follow a cursor. So most of the new tests walk the pages and assert on the concatenation:

- next page does not repeat the last row (the direct regression test)
- full forward walks ascending and descending return every row exactly once, including when `limit` exactly divides the row count
- rows with identical `created_at` are not duplicated across pages, and partial ties keep the overall order
- `prev_cursor` returns the immediately preceding page, in forward order, with correct `has_next`/`has_prev`
- cursor values round-trip with their type for 13 value kinds; `bool`/`int` and `datetime`/`date` are not conflated
- a datetime cursor survives a real page walk against a `DateTime` column
- malformed cursors are rejected: undecodable, no row reference, unknown direction, untagged value (the old format), value that does not match its tag
- offset mode still works, including past-the-end and negative offsets
- `total` counts the whole set on every page, empty result sets, `limit < 1`, and `paginate_and_respond` agreeing with `paginate_query`

`pytest tests/test_pagination.py tests/test_keyset_pagination.py` → 44 passed. Formatted with `black`.

### DIFF
--- a/backend/app/core/pagination.py
+++ b/backend/app/core/pagination.py
@@ -1,10 +1,261 @@
-from typing import Any, List, Optional, Tuple, TypeVar
-from sqlalchemy.orm import Query
-from sqlalchemy import desc, asc, inspect
+"""
+Keyset ("cursor") pagination for SQLAlchemy queries.
 
-from app.schemas.pagination import PaginatedResponse, encode_cursor, decode_cursor
+The rule that makes keyset pagination work is that the cursor must describe a
+*strict* boundary over a *total* order. Get either half wrong and pages overlap
+or drop rows:
+
+* **Strict.** The cursor points at the last row of the page you just got. The
+  next page must be the rows strictly after it. An inclusive ``<=`` returns
+  that row again as the first row of the next page.
+* **Total.** ``created_at`` is not unique. Two rows written in the same
+  transaction share a timestamp, so ordering by it alone leaves their relative
+  order up to the planner, and it is free to answer differently between two
+  queries. Every comparison here carries the primary key as a tiebreaker, so
+  the order is total even when the sort column is not.
+
+Cursors are opaque to clients but not encrypted -- they are base64 of a small
+JSON object, and a client that decodes one learns only the sort value of a row
+it was already shown. They *are* type-tagged: a cursor over a ``DateTime``
+column has to come back out of the cursor as a ``datetime``, not as the string
+``'2026-08-11 10:54:43.123456+00:00'``, or the comparison against the column
+either raises or silently degrades to a text comparison with different ordering
+than the index.
+"""
+
+from __future__ import annotations
+
+import datetime as dt
+import uuid
+from decimal import Decimal
+from typing import Any, List, Literal, Optional, Tuple, TypeVar
+
+from sqlalchemy import and_, asc, desc, or_
+from sqlalchemy.orm import Query
+
+from app.schemas.pagination import PaginatedResponse, decode_cursor, encode_cursor
 
 T = TypeVar("T")
+
+Direction = Literal["next", "prev"]
+
+
+class InvalidCursor(ValueError):
+    """
+    Raised when a cursor cannot be decoded or does not describe this query.
+
+    This is a client error, not a server error: a caller that sends a cursor
+    from a different sort order, or a truncated one, has asked for something we
+    cannot answer. Silently falling back to page 1 is worse -- an infinite
+    scroll that hits it restarts from the top and loops forever.
+    """
+
+
+# ---------------------------------------------------------------------------
+# Cursor value encoding
+# ---------------------------------------------------------------------------
+#
+# JSON has no datetime and no UUID, so a tag travels with the value and is used
+# to rebuild the original type on the way back in.
+
+_TAG_NULL = "null"
+_TAG_STR = "str"
+_TAG_INT = "int"
+_TAG_FLOAT = "float"
+_TAG_BOOL = "bool"
+_TAG_DECIMAL = "dec"
+_TAG_UUID = "uuid"
+_TAG_DATETIME = "dt"
+_TAG_DATE = "date"
+_TAG_TIME = "time"
+
+
+def encode_cursor_value(value: Any) -> dict:
+    """Tag a Python value so it survives a JSON round trip with its type."""
+    if value is None:
+        return {"t": _TAG_NULL, "v": None}
+    # bool before int: bool is a subclass of int and would otherwise be tagged
+    # as one and come back as 0/1.
+    if isinstance(value, bool):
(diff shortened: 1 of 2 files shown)


nensii21:
The pagination logic looks well covered. One concern is the intentional cursor-format change: existing clients may still hold old cursors when this deploys. How will those cursors be handled at the API boundary, and should the InvalidCursor → 400 handler be included in this PR rather than deferred?

PR #1109, opened 2026-08-12
fix(security): pin link preview fetches to the address that passed validation (fixes #1104)

### LINKED ISSUE #1104: Link preview SSRF guard can be defeated by DNS rebinding; normalise_url also corrupts IPv6 hosts
## Description

`backend/app/utils/url_safety.py` resolves a hostname and checks every address it gets back, and `link_preview_service._fetch_html` correctly re-validates each redirect hop by hand instead of letting `httpx` follow them. That part is good.

The gap is what happens between the check and the connect. `validate_outbound_url` returns the addresses it verified:

```python
return SafeTarget(
    url=url,
    scheme=scheme,
    host=host,
    port=port,
    addresses=addresses,
)
```

and the fetch throws them away:

```python
target = validate_outbound_url(url)
...
fetched = self._fetch_html(target.url)   # only the URL string survives
```

`_fetch_html` hands the URL string to `httpx`, which does its **own** DNS lookup. Nothing ties the socket to the address we approved.

## Why this is exploitable

This is textbook DNS rebinding. An attacker controls `evil.example.com` and serves it with a 0-second TTL:

1. We call `getaddrinfo("evil.example.com")` → `203.0.113.10`. Public, passes.
2. `httpx` calls `getaddrinfo("evil.example.com")` a few milliseconds later → `169.254.169.254`.
3. We fetch the cloud metadata endpoint and put whatever comes back into a preview card.

The attacker does not even need precise timing — they can just alternate answers round-robin and retry the paste until it lands. The module docstring already anticipates this ("It exposes the resolved addresses to the caller, so a fetch can be re-checked at every redirect hop") but no caller consumes `addresses`, so the resolution is validated and then discarded.

The same applies to the redirect loop: `validate_outbound_url(current)` re-resolves, then `client.stream("GET", current)` resolves a third time.

## Second problem: `normalise_url` corrupts IPv6 hosts

```python
netloc = host
if parsed.port and parsed.port != DEFAULT_PORTS.get(scheme):
    netloc = f"{host}:{parsed.port}"
```

`parsed.hostname` strips the brackets from an IPv6 literal, and they are never put back:

```
>>> normalise_url("http://[::1]:8080/x")
'http://::1:8080/x'
>>> urlparse('http://::1:8080/x').hostname
None
```

The result is not a parseable URL. Since `normalise_url` is what builds the link-preview cache key, every IPv6 URL produces a mangled key, and `http://[::1]/a` and `http://[::2]/a` are not reliably distinguishable once mangled.

## Third problem: query-order-sensitive cache keys

`normalise_url` is documented as producing "a stable spelling of a URL, for use as a cache key", but it passes `parsed.query` through untouched. `?a=1&b=2` and `?b=2&a=1` are the same page and get two cache entries, two outbound fetches, and two rate-limit hits.

## Suggested fix

- Have `validate_outbound_url` return a target the caller can actually pin to, and give the fetch path a transport that connects to the approved address while still sending the original `Host` / SNI. Re-check the address on every hop.
- Bracket IPv6 hosts in `normalise_url` and add a test for the literal-host case.
- Sort query parameters (and drop an empty query) so the cache key is order-insensitive.
- Cover all of it in `backend/tests/test_link_previews.py`; the rebinding case can be tested with a resolver stub that returns a different answer on the second call, so no network is needed.

MOHITKOURAV01:
# Pull Request

## Summary

The SSRF guard resolved a hostname and checked every address it got back, then handed the *hostname* to `httpx`, which resolved it again. Nothing tied the connection to the address that passed validation, which leaves the guard open to DNS rebinding. This pins each fetch to a verified address, and fixes two `normalise_url` bugs found along the way.

---

## Related Issue

Closes #1104

---

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Refactoring
- [ ] UI/UX enhancement
- [x] Tests
- [ ] Other

---

### 1. Pin the connection to the verified address

`SafeTarget` already carried `addresses`, and the module docstring already said they were exposed so a fetch could be re-checked — but nothing consumed them:

```python
target = validate_outbound_url(url)
...
fetched = self._fetch_html(target.url)   # only the URL string survives
```

The attack is textbook rebinding. An attacker serves `evil.example.com` with a 0-second TTL:

1. `getaddrinfo("evil.example.com")` → `203.0.113.10`. Public, passes.
2. `httpx` resolves the same name a few milliseconds later → `169.254.169.254`.
3. The metadata response gets rendered into a preview card.

No precise timing needed — alternating answers round-robin and retrying the paste gets there eventually.

Added `pin_target()`, which turns a validated target into three things:

- a **URL with the literal IP** in place of the hostname, so there is no second lookup to poison
- a **`Host` header** carrying the original hostname (with the port when it is not the scheme default), so name-based virtual hosting still resolves to the right site rather than a 421
- an **`sni_hostname` extension**, so TLS still presents and validates the right certificate — without it the handshake uses the IP as the server name and every certificate on the internet fails to verify

`pin_target` refuses an address that is not in `target.addresses`, so pinning somewhere unverified is an error rather than a quiet hole.

`_fetch_html` now pins **every hop**, not just the first. Redirects were already re-validated by hand; they are now re-pinned too, so hop two cannot be the metadata service either. Relative redirects resolve against the logical URL rather than the pinned one, and the reported `final_url` stays logical — an IP has no business appearing in a preview card.

### 2. `normalise_url` corrupted IPv6 hosts

`urlparse` strips the brackets from an IPv6 literal in `.hostname`, and they were never put back:

```
>>> normalise_url("http://[::1]:8080/x")
'http://::1:8080/x'
>>> urlparse('http://::1:8080/x').hostname
None
```

Not a parseable URL. Since `normalise_url` builds the link-preview cache key, every IPv6 URL produced a mangled key. Added `format_host_for_url()`, which brackets IPv6 and leaves everything else alone.

### 3. `normalise_url` produced order-sensitive cache keys

It is documented as "a stable spelling of a URL, for use as a cache key" but passed `parsed.query` through untouched, so `?a=1&b=2` and `?b=2&a=1` were two entries, two outbound fetches, and two rate-limit slots for the same page.

Parameters are now sorted — **by key only, with a stable sort**, so repeated keys keep their relative order. `?tag=a&tag=b` stays distinct from `?tag=b&tag=a`, since plenty of APIs treat repeated parameters as an ordered list and collapsing them would be a cache collision between two different pages.

---

## Testing

- [x] Tested locally
- [x] Existing tests pass
- [x] Added new tests

`backend/tests/test_url_pinning.py`, 36 tests. No DNS and no network — the resolver and the transport are both injected, and the tests assert on what the HTTP client was actually asked to connect to.

- `pin_target`: URL carries the address not the hostname; `Host` header and SNI preserved; path/query survive; pathless URL pins to root; non-default port lands in both URL and `Host`; IPv6 bracketed and parseable by `httpx.URL`; a specific verified address can be chosen; an unverified address raises; choice is deterministic
- the fetch path: the client connects to the verified address with the right `Host` and SNI; the rebinding resolver (public on the first lookup, metadata on every one after) never gets a second bite; each redirect hop is pinned to its own address; a redirect to a private address is still refused after exactly one connection; relative redirects resolve against the logical URL; `final_url` is logical, not pinned; non-HTML content types are not parsed; the redirect budget is bounded
- `normalise_url`: scheme/host lowercasing, default-port dropping, fragment dropping, root path, IPv6 brackets on both default and explicit ports, query-order insensitivity, repeated-key order preservation, blank values, empty query, idempotence

Worth flagging: `_fetch_html` had **no** test coverage before this, so the hand-rolled redirect handling was untested as well. That is now covered.

`pytest tests/test_url_pinning.py tests/test_link_previews.py` → 83 passed (47 of those pre-existing). Formatted with `black`.

### DIFF
--- a/backend/app/services/link_preview_service.py
+++ b/backend/app/services/link_preview_service.py
@@ -11,6 +11,8 @@ The rules this module enforces on every fetch:
 * the destination is validated before we connect (see ``utils.url_safety``),
   and re-validated at every redirect hop -- following redirects automatically
   would let a public URL bounce us to ``169.254.169.254``
+* the connection is pinned to the address that validation approved, so DNS
+  cannot answer differently between the check and the connect
 * the body is streamed and abandoned once it exceeds a byte cap, so a hostile
   server cannot make us buffer a gigabyte
 * only HTML content types are parsed
@@ -34,8 +36,10 @@ import httpx
 from app.core.cache import cache_manager
 from app.core.config import settings
 from app.utils.url_safety import (
+    SafeTarget,
     UnsafeURL,
     normalise_url,
+    pin_target,
     validate_outbound_url,
 )
 
@@ -127,7 +131,7 @@ class LinkPreviewService:
             if cached is not None:
                 return LinkPreview(**cached)
 
-        fetched = self._fetch_html(target.url)
+        fetched = self._fetch_html(target)
         if fetched is None:
             cache_manager.set(
                 cache_key,
@@ -169,15 +173,26 @@ class LinkPreviewService:
     # Fetching
     # ------------------------------------------------------------------
 
-    def _fetch_html(self, url: str) -> Optional[tuple[str, str]]:
+    def _fetch_html(self, target: SafeTarget) -> Optional[tuple[str, str]]:
         """
-        Fetch a URL and return ``(body, final_url)``, or ``None`` on failure.
+        Fetch a validated target and return ``(body, final_url)``, or ``None``.
 
-        Redirects are followed by hand so that every hop goes back through
-        :func:`validate_outbound_url`. ``httpx``'s own redirect following would
-        happily walk us onto the metadata service.
+        Two things happen by hand here rather than being left to ``httpx``:
+
+        * **Redirects.** Every hop goes back through
+          :func:`validate_outbound_url`. ``httpx``'s own redirect following
+          would happily walk us onto the metadata service.
+        * **Address resolution.** Each hop connects to the address that hop's
+          validation actually approved, via :func:`pin_target`. Handing the
+          hostname to ``httpx`` would let it resolve a second time, and a
+          short-TTL record that answers public-then-private turns the whole
+          guard into decoration.
+
+        ``final_url`` is the logical URL, not the pinned one -- the IP is a
+        transport detail and has no business ending up in a preview card.
         """
-        current = url
+        current_target = target
+        current_url = target.url
 
         try:
             with httpx.Client(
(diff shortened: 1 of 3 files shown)


nensii21:
Since the security fix depends on ensuring the HTTP client never resolves the hostname again, can you confirm that the pinned IP + Host + SNI path is used for every redirect hop and that httpx cannot perform a fresh DNS lookup anywhere in that flow? This is the critical invariant preventing DNS rebinding from bypassing the SSRF validation.

PR #1110, opened 2026-08-12
fix(feed): stop the activity feed loading every page at once (fixes #1105)

### LINKED ISSUE #1105: useIntersectionObserver rebuilds its observer on every fetch, so the activity feed loads all pages at once
## Description

`useIntersectionObserver` tears down and rebuilds its observer whenever either of its two arguments changes identity, and `IntersectionObserver` fires its callback immediately on `observe()` for an element that is already in view. Together those two facts make the activity feed pull every page it has as soon as the sentinel scrolls into view once, without any further scrolling.

## The loop

`ActivityFeed.tsx` wires it up like this:

```tsx
const handleIntersect = useCallback(() => {
  if (hasNextPage && !isFetchingNextPage) {
    fetchNextPage();
  }
}, [hasNextPage, isFetchingNextPage, fetchNextPage]);

const sentinelRef = useIntersectionObserver(
  handleIntersect,
  Boolean(hasNextPage && !isFetchingNextPage),
);
```

Both `handleIntersect` and `enabled` change every time `isFetchingNextPage` flips, and the hook memoises its ref callback on `[enabled, onIntersect]`. So:

1. Sentinel enters the viewport → `onIntersect()` → `fetchNextPage()`.
2. `isFetchingNextPage` becomes `true` → `enabled` becomes `false` → new ref callback → React calls the old one with `null` and the new one with the node → `disconnect()`, then early `return` because it is disabled. Observer gone.
3. Fetch resolves → `isFetchingNextPage` becomes `false` → `enabled` becomes `true` → new ref callback again → a **fresh** `IntersectionObserver` is constructed and `observe(node)` is called.
4. The sentinel is still in the viewport, so the brand-new observer fires straight away with `isIntersecting: true` → `onIntersect()` → step 1.

It only stops when `hasNextPage` goes false. On a feed with a lot of history that is every page, back to back, and the user never scrolled.

Even with a stable callback the hook has the plain version of this problem: while the sentinel stays intersecting, `isIntersecting` is reported again on any change to the intersection ratio, and there is no edge-detection, so scrolling slowly inside the 200px margin fires repeatedly.

## Other problems

- **No SSR/environment guard.** `new IntersectionObserver(...)` is called unconditionally. This is an SSR app (`src/server.ts`), and the constructor does not exist in jsdom either, so any test that renders a component using this hook throws `IntersectionObserver is not defined`. That is presumably why there is no test file for it.
- **`rootMargin` is hardcoded** to `"200px"` with no way to pass `root`, `threshold`, or a different margin, so the hook cannot be used inside a scroll container.
- **Stale-closure footgun.** Because the observer is rebuilt to pick up a new `onIntersect`, callers are forced to `useCallback` their handler or eat a rebuild on every render. The hook should hold the latest callback in a ref instead.

## Suggested fix

Keep the same ref-callback shape so `ActivityFeed` does not have to change much, but:

- Store `onIntersect` in a ref that is updated on every render, and drop it from the memo deps so the observer is not rebuilt when the handler identity changes.
- Only call the handler on a **false → true** transition, so a sentinel that stays in view does not refire.
- Accept `root`, `rootMargin`, `threshold`, and a `once` option.
- Guard on `typeof IntersectionObserver === "undefined"` and no-op.
- Disconnect on unmount and when the node is detached.

Worth adding a test file for this too — with a small `IntersectionObserver` stub the transition behaviour is straightforward to assert.

MOHITKOURAV01:
# Pull Request

## Summary

`useIntersectionObserver` rebuilt its observer whenever either of its arguments changed identity, and `IntersectionObserver` fires its callback immediately on `observe()` for an element already in view. In `ActivityFeed` those two facts formed a loop that pulled every page of the feed as soon as the sentinel first became visible.

---

## Related Issue

Closes #1105

---

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation update
- [x] Performance improvement
- [ ] Refactoring
- [ ] UI/UX enhancement
- [x] Tests
- [ ] Other

---

### The loop

```
sentinel enters view      → onIntersect() → fetchNextPage()
isFetchingNextPage: true  → enabled: false → new ref callback → observer disconnected
fetch resolves            → enabled: true  → new ref callback → fresh observer
fresh observer observes an element that is still in view → immediate report → back to line 1
```

It only stopped when `hasNextPage` went false. On a feed with any real history that is every page, back to back, and the user never scrolled.

### What changed in the hook

- **The handler lives in a ref** and is no longer an observer dependency, so a handler whose identity changes (an inline arrow, or a `useCallback` whose deps moved) no longer costs a rebuild.
- **Fires only on a false → true transition.** Even with a stable handler, a sentinel that stays in view is reported again on every change to the intersection ratio, so scrolling slowly inside the 200px margin was firing repeatedly.
- **One effect owns the observer's whole lifecycle**, with the node kept in state. My first pass at this kept the node in a ref and had both the ref callback and an effect call `observe()` — which meant the node was observed twice and the initial report arrived twice. Two of the tests below caught that, so they are worth keeping.
- **The returned ref callback has a stable identity**, so React stops calling it with `null` and then the node again on every render where it changed.
- **Guards on `typeof IntersectionObserver === "undefined"`.** This is an SSR app, and jsdom does not implement the API either — which is why there was no test file for this hook before.
- **Accepts `root`, `rootMargin`, `threshold` and `once`**, so it can be used inside a scroll container. `threshold` is compared by value, since it is almost always passed as an inline array and would otherwise force a rebuild every render.

### What changed in `ActivityFeed`

`enabled` no longer includes `isFetchingNextPage`:

```diff
-const sentinelRef = useIntersectionObserver(
-  handleIntersect,
-  Boolean(hasNextPage && !isFetchingNextPage),
-);
+const sentinelRef = useIntersectionObserver(handleIntersect, Boolean(hasNextPage));
```

`handleIntersect` already refuses to stack fetches, so there is nothing for the toggle to protect against — all it did was tear the observer down and rebuild it into another immediate report.

Behaviour worth calling out for reviewers: after this, if a loaded page does not push the sentinel out of view, the feed waits for the user to scroll rather than chaining straight into the next page. That is the intended trade — the alternative is the runaway loop — but if you would rather it kept filling until the sentinel clears, say so and I will add an explicit "keep going while still visible" check instead of getting it as an accident of observer churn.

---

## Testing

- [x] Tested locally
- [x] Existing tests pass
- [x] Added new tests

`frontend/src/hooks/__tests__/useIntersectionObserver.test.tsx`, 23 tests, driven by a stub that reproduces the behaviour that actually matters here — reporting the target's state as soon as `observe()` is called.

- firing: fires on entry, not while out of view, not repeatedly while it stays in view, again after leaving and re-entering, and not at all when disabled
- the loop: no rebuild when the handler identity changes; the newest handler is always the one called; the `ActivityFeed` scenario end to end (a fetch toggling `enabled`) yields one page rather than running away; a re-render handing back the same node does not refire
- options: default 200px margin, custom margin, threshold, a scroll container as root, an inline threshold array not counting as a change, and a real margin change rebuilding
- `once`: fires a single time and disconnects, and does not fire again on re-entry
- teardown: disconnects on unmount, leaves nothing live when disabled, never more than one live observer across ten enable/disable flips, and re-observes when enabled again
- no `IntersectionObserver`: renders and unmounts without throwing

Full suite: 505 tests pass. `tsc --noEmit` and `eslint` clean.

### DIFF
--- a/frontend/src/features/activities/components/ActivityFeed.tsx
+++ b/frontend/src/features/activities/components/ActivityFeed.tsx
@@ -102,10 +102,12 @@ export function ActivityFeed({ actorId, targetId, targetType }: ActivityFeedProp
     }
   }, [hasNextPage, isFetchingNextPage, fetchNextPage]);
 
-  const sentinelRef = useIntersectionObserver(
-    handleIntersect,
-    Boolean(hasNextPage && !isFetchingNextPage),
-  );
+  // `enabled` deliberately leaves out `isFetchingNextPage`. Toggling it off
+  // while a page loads tears the observer down, and the fresh one built when
+  // the fetch finishes reports the still-visible sentinel immediately, which
+  // asks for the next page, which toggles it again. `handleIntersect` already
+  // refuses to stack fetches, so the observer can just stay up.
+  const sentinelRef = useIntersectionObserver(handleIntersect, Boolean(hasNextPage));
 
   const activities = data?.pages.flatMap((page) => page) ?? [];
 
--- a/frontend/src/features/activities/components/ActivityFeed.tsx
+++ b/frontend/src/features/activities/components/ActivityFeed.tsx
@@ -162,9 +162,11 @@ export function ActivityFeed({ actorId, targetId, targetType }: ActivityFeedProp
               </div>
             )}
 
-            {hasNextPage && !isFetchingNextPage && (
-              <div ref={sentinelRef} className="h-1" aria-hidden="true" />
-            )}
+            {/* Kept mounted across a fetch. Unmounting it hands the hook a null
+                node and then a brand new one, and a brand new observer reports
+                an already-visible sentinel immediately -- which is the other
+                half of the runaway pagination. */}
+            {hasNextPage && <div ref={sentinelRef} className="h-1" aria-hidden="true" />}
 
             {!hasNextPage && activities.length > 0 && (
               <p className="py-4 text-center text-sm text-gray-400">
--- a/frontend/src/hooks/__tests__/useIntersectionObserver.test.tsx
+++ b/frontend/src/hooks/__tests__/useIntersectionObserver.test.tsx
@@ -1,0 +1,367 @@
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+import { act, render } from "@testing-library/react";
+import { useCallback, useState } from "react";
+
+import { useIntersectionObserver } from "../useIntersectionObserver";
+
+/**
+ * A stand-in for the real thing, which jsdom does not implement.
+ *
+ * The important behaviour to reproduce is that a real `IntersectionObserver`
+ * reports the target's current state as soon as `observe()` is called — that
+ * initial report is what turned an observer rebuild into a fetch loop.
+ */
+class FakeIntersectionObserver implements IntersectionObserver {
+  static instances: FakeIntersectionObserver[] = [];
+  /** Whether a freshly observed target is treated as already in view. */
+  static initiallyIntersecting = false;
+
+  readonly root: Element | Document | null;
+  readonly rootMargin: string;
+  readonly thresholds: ReadonlyArray<number>;
+
+  targets: Element[] = [];
+  disconnected = false;
+
+  private callback: IntersectionObserverCallback;
+
+  constructor(callback: IntersectionObserverCallback, options?: IntersectionObserverInit) {
+    this.callback = callback;
+    this.root = (options?.root as Element | null) ?? null;
+    this.rootMargin = options?.rootMargin ?? "0px";
+    this.thresholds = options?.threshold === undefined ? [0] : [options.threshold].flat();
+    FakeIntersectionObserver.instances.push(this);
+  }
+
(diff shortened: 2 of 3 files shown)


MOHITKOURAV01:
Pushed a follow-up (3deff88) — I had missed half of the loop on the first pass.

`ActivityFeed` rendered the sentinel as `{hasNextPage && !isFetchingNextPage && <div ref={sentinelRef} />}`, so the element was **unmounted for the duration of every fetch** and remounted afterwards. A remounted node is a genuinely new node, so no amount of deduping inside the hook can catch it — the fresh observer sees an already-visible sentinel, reports immediately, and asks for another page. Changing `enabled` alone did not close it.

The sentinel is now kept mounted whenever `hasNextPage`, and there is a test that pins the cost of getting it wrong: one extra fire per remount, and still only one live observer either way.

nensii21:
Good catch. Keeping the sentinel mounted is the right fix, and the added test covers the remount regression properly. The observer cleanup and single-live-observer behavior also look good. 3deff88 looks solid.

PR #1111, opened 2026-08-12
fix(moderation): stop blocking posts that contain long numbers (fixes #1106)

### LINKED ISSUE #1106: Spam filter blocks posts containing long numbers and misses trivially obfuscated spam
## Description

`analyzeSpam` in `frontend/src/lib/validation/spamDetection.ts` is a hard gate on posting. `postsApi.create` and `postsApi.comment` both throw before the request is sent:

```ts
const spamCheck = analyzeSpam(body.content);
if (spamCheck.isSpam) {
  throw new Error(`Post rejected by AI Spam Filter: ${spamCheck.reasons.join(". ")}`);
}
```

So a false positive is not a warning — the user simply cannot post, and the message they get blames them for "suspicious link shorteners or contact numbers".

Given that, the current heuristics are both too loose in one direction and trivially evadable in the other.

## False positives

**Any long run of digits is treated as a phone number.**

```ts
const PHISHING_PATTERNS = [
  ...
  /\+?[0-9]{10,12}/, // Raw phone numbers
];
```

This is unanchored and has no boundary, so it matches 10–12 digits *inside* a longer number too. It fires on a Unix timestamp (`1754985600`), a snowflake/ID pasted into a bug report, a checksum, a large numeric literal in a code sample. On its own that is `+0.4`; add a single keyword match (`+0.35`) and the post is blocked at `0.75`.

**Keywords match on substrings, not words.** `lowerText.includes(keyword)` means `"casino"` matches inside `casinos`, and more importantly single-word entries fire inside unrelated words. There is no word boundary anywhere in the keyword pass.

**The caps rule punishes code.** `text.length > 20` and `capsCount / letterCount > 0.6` flags a post that is mostly constants — `MAX_RETRIES`, `API_BASE_URL`, an SQL snippet, a stack trace with class names. This is a developer network; posts are full of that.

**The repetition rule punishes code too.** `wordSet.size / words.length < 0.4` on anything over 8 words flags a diff, a config block, or any snippet with repeated identifiers.

## False negatives

Every check runs against the raw string, so the cheapest possible obfuscation walks straight through. All of these score `0`:

| Input | Why it passes |
| --- | --- |
| `fr33 m0n3y` | leetspeak, no keyword match |
| `b u y  n o w` | spacing |
| `buy&#8203;now` (zero-width space) | invisible separator |
| `ＢＵＹ ＮＯＷ` (full-width) | different code points, and `toLowerCase()` does not fold them to ASCII |
| `t․me/foo` (U+2024 one-dot leader) | homoglyph defeats the `/t\.me\//i` pattern |
| `𝐛𝐮𝐲 𝐧𝐨𝐰` (mathematical bold) | different code points |

The link-shortener list is also a fixed set of five, missing the ones that actually show up (`is.gd`, `cutt.ly`, `rb.gy`, `shorturl.at`, `discord.gg`, and so on).

## Other notes

- The scores are additive with no cap per category, so three keyword hits alone (`3 × 0.35 = 1.05`) max out the score regardless of context or length. A long, obviously legitimate post that happens to mention "discount code" twice is treated the same as a two-line advert.
- Link density is an absolute count (`urls.length > 3`) rather than a ratio, so a well-sourced 2000-word write-up with five references is penalised the same as a five-line link dump.
- There is no test file for this module at all, which is uncomfortable for something that can silently block posting.

## Suggested fix

- Normalise before matching: NFKC fold (handles full-width and mathematical alphanumerics), strip zero-width and combining marks, map common homoglyphs, collapse leetspeak digit substitutions and separator runs. Match against the normalised form.
- Match keywords on word boundaries, and use the normalised text so `fr33 m0n3y` and `b u y  n o w` are caught.
- Replace the digit-run regex with something that actually looks like a phone number (separators, country code, plausible grouping) and skip candidates that sit inside a longer digit run or inside a code span.
- Skip fenced/inline code when applying the caps and repetition rules, and require a minimum word count before either can fire.
- Cap each signal's contribution and make link density a ratio of the post length.
- Return the normalised text alongside the result so a reviewer can see what was matched, and add a test file covering both directions.

MOHITKOURAV01:
# Pull Request

## Summary

`analyzeSpam` gates posting — `postsApi.create` and `postsApi.comment` both throw before the request is sent — so a false positive does not warn anyone, it stops them posting. The phone-number pattern was `/\+?[0-9]{10,12}/`, which fires on any Unix timestamp or long id, while every check ran against the raw string so the cheapest obfuscation scored zero. This fixes both directions.

---

## Related Issue

Closes #1106

---

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Refactoring
- [ ] UI/UX enhancement
- [x] Tests
- [ ] Other

---

### Fewer false positives

**Phone numbers.** `/\+?[0-9]{10,12}/` is unanchored and has no boundary, so it matched ten to twelve digits *inside* a longer run too — a Unix timestamp (`1754985600`), a job id, a checksum, a large numeric literal in a snippet. On its own that is `+0.4`; one keyword match on top and the post is refused at `0.75`. Detection now wants a country code or separator grouping, which is how a spam post actually writes a number. A number on its own also no longer blocks a post — plenty of legitimate posts share one.

**Keywords** match on word boundaries, so `betting` no longer fires inside `bettingham`.

**Shortener hosts** are matched as hosts (and subdomains), so an article whose *path* mentions `t.me` is not treated as a chat invite.

**Code is removed before the shape checks.** Fenced blocks, indented blocks, inline spans and URLs are stripped, and both the shouting and repetition rules additionally skip anything that reads like code. `MAX_RETRIES = 3; API_BASE_URL = ...` was being flagged as shouting, and a snippet with repeated identifiers as repetitive prose. This is a developer network; posts are full of both.

**Link density is a ratio**, not `urls.length > 3`, so a 2000-word write-up with five references is not scored like a five-line link dump.

**Every signal is capped**, so no single rule can carry a post over the line alone. Three keyword hits used to be `3 × 0.35 = 1.05` regardless of context or length.

### Fewer false negatives

Matching now runs after a normalisation pass — NFKD, combining marks stripped, NFKC recompose, homoglyphs mapped, invisible characters removed, leetspeak decoded, spelled-o-u-t words closed up. That closes all six evasions from the issue in one place:

| Input | Before | After |
| --- | --- | --- |
| `fr33 m0n3y` | 0 | caught |
| `b u y  n o w` | 0 | caught |
| `buy<ZWSP>now` | 0 | caught |
| `ＢＵＹ ＮＯＷ` | 0 | caught |
| `𝐛𝐮𝐲 𝐧𝐨𝐰` | 0 | caught |
| `t․me/foo` (U+2024) | 0 | caught |

The shortener list also picked up the hosts that actually turn up — `is.gd`, `cutt.ly`, `rb.gy`, `shorturl.at`, `rebrand.ly`, `discord.gg` and friends.

### Two things the tests caught while writing this

Both are the kind of thing that looks obviously right and is not, so they have named regression tests:

- **Decoding `!` as `i`** turns `free money!` into `free moneyi`, which then matches *nothing*. It loses far more than the handful of `cl!ck` it catches, so `!` is out of the substitution map entirely, and `@`/`$` are only decoded away from a token's edges (`price$` is a price, `@alice` is a mention).
- **Collapsing spaced-out letters has to treat a wider gap as a word break.** Collapsing greedily turns `b u y  n o w` into `buynow`, which matches nothing either. The separator is a single space, and the double space survives as the word break.

### Result shape

`SpamCheckResult` gained `normalizedText` and a `signals` array (`id`, `reason`, `weight`), so a disputed block can be explained rather than argued about. `reasons` is unchanged, so `postsApi`'s error message still works as-is. `SPAM_THRESHOLD` is exported rather than being a bare `0.5` in the middle of the function.

---

## Testing

- [x] Tested locally
- [x] Existing tests pass
- [x] Added new tests

`frontend/src/lib/validation/__tests__/spamDetection.test.ts`, 70 tests, in a module that had none — which is uncomfortable for something that can silently stop people posting.

- **normalisation helpers**: fullwidth, mathematical alphanumerics, zero-width, soft hyphens, combining marks, one-dot leader, Cyrillic lookalikes; leetspeak decoding scoped to mixed tokens with the edge cases above; spaced-letter collapsing and its word-break behaviour
- **legitimate posts are not blocked**, as a table: a Unix timestamp, a long numeric id, a version plus checksum, screaming constants, a fenced code block, a stack trace, a well-sourced long write-up with five links, a plain question, a single marketing-ish word, empty and whitespace-only
- **obvious spam is blocked**: plain advert, chat invite, shortener plus keywords, contact number plus keywords
- **each of the six obfuscations** from the issue
- **phone numbers**: international, grouped, parenthesised; *not* a semver string or a bare digit run; and not enough on its own to block
- **scoring**: keyword cap, never above 1, never below 0, agreement with `SPAM_THRESHOLD`, one reason per signal, normalised text exposed, empty input
- **shape rules**: shouting flagged in prose but not in code and not for a short `LGTM`; stretched characters; repetition flagged in prose but not in a short post

Full suite: 548 tests pass. `tsc --noEmit` and `eslint` clean.

---

## Notes for reviewers

The keyword list, the host list and the weights are all judgement calls, and I have kept them close to the originals rather than expanding scope — the point of this PR is the matching and the scoring, not the vocabulary. If any threshold looks wrong, `CAPS` and `SPAM_THRESHOLD` are now in one place at the top of the file and the `signals` breakdown makes it easy to see what a given post actually scored.

Worth a second opinion on one judgement in particular: I made a phone number on its own non-blocking. That is a deliberate loosening — sharing a number is not spam by itself — but if moderation would rather keep it strict, it is a one-line change to `CAPS.phone`.

### DIFF
--- a/frontend/src/lib/validation/__tests__/spamDetection.test.ts
+++ b/frontend/src/lib/validation/__tests__/spamDetection.test.ts
@@ -1,0 +1,332 @@
+import { describe, expect, it } from "vitest";
+
+import {
+  SPAM_THRESHOLD,
+  analyzeSpam,
+  collapseSpacedLetters,
+  decodeLeetspeak,
+  normalizeForMatching,
+} from "../spamDetection";
+
+/** `analyzeSpam` gates posting, so "blocked" is the thing worth asserting on. */
+function isBlocked(text: string): boolean {
+  return analyzeSpam(text).isSpam;
+}
+
+function signalIds(text: string): string[] {
+  return analyzeSpam(text).signals.map((s) => s.id);
+}
+
+describe("normalizeForMatching", () => {
+  it("folds fullwidth characters to ASCII", () => {
+    expect(normalizeForMatching("ＢＵＹ ＮＯＷ")).toBe("buy now");
+  });
+
+  it("folds mathematical alphanumerics to ASCII", () => {
+    expect(normalizeForMatching("𝐛𝐮𝐲 𝐧𝐨𝐰")).toBe("buy now");
+  });
+
+  it("strips zero-width characters", () => {
+    expect(normalizeForMatching("buy​now")).toBe("buynow");
+  });
+
+  it("strips soft hyphens", () => {
+    expect(normalizeForMatching("buy­now")).toBe("buynow");
+  });
+
+  it("strips combining marks", () => {
+    expect(normalizeForMatching("b̶u̶y̶")).toBe("buy");
+  });
+
+  it("maps a one-dot leader to a full stop", () => {
+    expect(normalizeForMatching("t․me")).toBe("t.me");
+  });
+
+  it("maps cyrillic lookalikes to latin", () => {
+    // Cyrillic es, a and o in place of the latin c, a and o.
+    expect(normalizeForMatching("саsinо")).toBe("casino");
+  });
+
+  it("leaves ordinary text alone apart from casing", () => {
+    expect(normalizeForMatching("Hello World")).toBe("hello world");
+  });
+});
+
+describe("decodeLeetspeak", () => {
+  it("decodes digits inside words", () => {
+    expect(decodeLeetspeak("fr33 m0n3y")).toBe("free money");
+  });
+
+  it("decodes symbol substitutions", () => {
+    expect(decodeLeetspeak("c@sino")).toBe("casino");
+  });
+
+  it("leaves a pure number alone", () => {
+    // The whole reason to scope this: a timestamp must not become a word.
+    expect(decodeLeetspeak("1754985600")).toBe("1754985600");
+  });
+
+  it("leaves a word with no substitutions alone", () => {
+    expect(decodeLeetspeak("hello world")).toBe("hello world");
+  });
+
+  it("does not touch numbers next to words", () => {
+    expect(decodeLeetspeak("issue 1234567890 is open")).toContain("1234567890");
+  });
+
+  it("leaves a trailing exclamation mark as punctuation", () => {
+    // Decoding `!` to `i` turns "free money!" into "free moneyi", which then
+    // matches no keyword at all — it loses far more than it catches.
+    expect(decodeLeetspeak("free money!")).toBe("free money!");
+  });
+
+  it("leaves a trailing dollar sign as a currency symbol", () => {
+    expect(decodeLeetspeak("price$")).toBe("price$");
+  });
+
+  it("leaves a leading at sign as a mention", () => {
+    expect(decodeLeetspeak("@alice")).toBe("@alice");
+  });
+
+  it("still decodes a symbol in the middle of a word", () => {
+    expect(decodeLeetspeak("ca$ino")).toBe("casino");
+  });
+});
+
+describe("collapseSpacedLetters", () => {
+  it("closes up a spelled-out word", () => {
+    expect(collapseSpacedLetters("b u y  n o w")).toBe("buy now");
+  });
+
+  it("leaves short sequences alone", () => {
+    // Two single letters is ordinary prose, not obfuscation.
+    expect(collapseSpacedLetters("a b test")).toBe("a b test");
+  });
+
+  it("leaves normal words alone", () => {
+    expect(collapseSpacedLetters("this is a normal sentence")).toBe("this is a normal sentence");
+  });
+
+  it("treats a wider gap as a word break", () => {
+    // Collapsing across the double space would give "buynow", which matches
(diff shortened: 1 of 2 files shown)


nensii21:
Because this is a hard gate on both posts and comments, false positives have a direct user-facing impact. Do we have an intended recovery path for a legitimate post that crosses SPAM_THRESHOLD, or is the submission simply rejected with no override/report mechanism?

PR #1174, opened 2026-08-14
fix(rate-limit): key on the real client and share counters across workers

### LINKED ISSUE #1169: Rate limiter keys on the proxy address and counts per process, so limits are global and multiplied by replica count
### ✦ Describe the Bug

`backend/app/middleware/rate_limit.py` builds the global limiter like this:

```python
limiter = Limiter(
    key_func=get_remote_address,
    default_limits=[settings.DEFAULT_RATE_LIMIT],
    enabled=settings.ENABLE_RATE_LIMIT,
    headers_enabled=True,
)
```

Two things are wrong with that in the way this application is actually deployed.

**1. `get_remote_address` returns the peer socket address.** In production the peer is the ingress controller, not the client — this repo ships `k8s/` manifests and a `docker-compose` stack that both put a proxy in front of the app. Every request therefore hashes to the *same* rate-limit key.

**2. There is no `storage_uri`.** SlowAPI defaults to an in-memory store, which is per worker process. `REDIS_URL` is already configured and used by `app/core/cache.py`, but the limiter does not use it.

### ⟡ Steps to Reproduce

1. Run the app behind any reverse proxy (the provided `docker-compose.dev.yml` or the `k8s/` ingress).
2. From two different client IPs, send 6 requests to `POST /api/auth/login` (`LOGIN_RATE_LIMIT = "5/minute"`).
3. The sixth request is refused regardless of which client sent it — both clients share one bucket.
4. Now scale the deployment to 3 replicas and repeat: roughly 15 attempts get through before any are refused.

### ✦ Expected Behavior

- The limit applies per real client, using the forwarded client address when the request came from a proxy we trust.
- Counters are shared across workers and replicas, so `5/minute` means five, not five times the replica count.

### ⟡ Root Cause / Impact

The two defects push in opposite directions and both are bad:

- **Shared bucket:** `LOGIN_RATE_LIMIT = "5/minute"` and `REGISTER_RATE_LIMIT = "3/hour"` become global caps. Three registrations *per hour across the entire site* is a trivial denial of service — one client can lock out every other user by exhausting the shared counter. Auth limits are the most restrictive in the config, so they break first and hardest.
- **Per-process counters:** the effective limit is multiplied by the number of workers and replicas, and every counter resets on deploy or restart. The brute-force protection the auth limits are there to provide is proportionally weaker, and unpredictably so.

`X-Forwarded-For` must not be trusted unconditionally — a client can send that header directly and forge any address, which turns a per-IP limit into no limit at all. The header is only meaningful when the request genuinely arrived from a known proxy, so the fix needs a trusted-proxy setting rather than a blind `request.headers["x-forwarded-for"]`.

### ✅ Acceptance Criteria

- [ ] Add settings for trusted proxy CIDRs and the number of trusted proxy hops, defaulting to trusting nothing
- [ ] Derive the client address from `X-Forwarded-For` only when the immediate peer is a trusted proxy; otherwise use the peer address
- [ ] Ignore private/invalid entries when walking the forwarded chain
- [ ] Point the limiter at `REDIS_URL` so counters are shared, falling back to in-memory when Redis is unreachable rather than failing to boot
- [ ] Key authenticated requests by user id where available, so a shared NAT egress does not put unrelated users in one bucket
- [ ] Tests: forged `X-Forwarded-For` from an untrusted peer is ignored; a trusted proxy's header is honoured; multi-hop chains pick the right hop

MOHITKOURAV01:
# Pull Request

## Summary

The rate limiter keyed on the peer socket address and kept its counters in the worker process. Behind a reverse proxy that means one shared bucket for the entire internet; across N workers it means N times the configured limit. Both are fixed here, along with two other places that were reading the client-controlled end of `X-Forwarded-For`.

---

## Related Issue

Closes #1169

---

## Type of Change

- [x] Bug fix
- [ ] New feature
- [x] Documentation update
- [ ] Performance improvement
- [ ] Refactoring
- [ ] UI/UX enhancement
- [x] Tests
- [ ] Other

---

### The key function

`slowapi.util.get_remote_address` returns `request.client.host` — the peer on the socket. In every deployment this repo describes (`k8s/` ingress, `docker-compose.dev.yml`) the peer is the proxy, so every request hashed to one key.

The tightest limits break the hardest. `REGISTER_RATE_LIMIT = "3/hour"` as a *global* cap is three registrations per hour across the whole site, and one client can exhaust it and lock everyone else out. `LOGIN_RATE_LIMIT = "5/minute"` is five logins per minute site-wide.

Reading `X-Forwarded-For` to fix that would be worse, not better. It is an ordinary request header — a client can send one and claim any address it likes, and a limit keyed on an address the caller picks is not a limit at all; the attacker just varies the header.

New `app/core/client_address.py`:

- **Trust gate first.** The header is only evidence when the immediate peer is inside `TRUSTED_PROXY_CIDRS`. From anywhere else it is ignored and the peer address is used.
- **Walk the chain from the right.** Skip hops that are our own proxies; stop at the first that is not. The left-hand entries are the ones a client can fabricate — a client sending `X-Forwarded-For: 1.2.3.4` gets its real address *appended* by the proxy, so the rightmost untrusted entry is the real one.
- **An unparseable hop stops the walk** rather than being skipped, so a client cannot hide a real hop behind junk.
- **Fail closed.** `TRUSTED_PROXY_CIDRS` is empty by default, meaning no header is trusted — identical behaviour to before this PR. A malformed CIDR entry is dropped rather than raising, because taking the application down over a fat-fingered environment variable is worse than running with one fewer trusted proxy, and dropping it fails closed.
- **Address shapes.** Handles `1.2.3.4:5678`, `[2001:db8::1]:443` and bare IPv6. A naive `split(":")[0]` turns `2001:db8::1` into `2001`; only a single colon means "address with port".
- **No peer at all** reports the constant `"unknown"` rather than `None`, since `None` keys collapse together in exactly the way this module exists to prevent.

### The storage backend

No `storage_uri` meant SlowAPI's in-memory default: counters per worker process, so the effective limit was multiplied by the worker and replica count, unpredictably, and reset on every deploy. `RATE_LIMIT_STORAGE_URI` points it at Redis.

It is a **new setting rather than reusing `REDIS_URL`** so a developer without Redis running gets working limits instead of a boot failure. When it is unset outside of tests, the module logs a warning — a limiter that silently counts per process looks fine right up until it matters. When it is set but unreachable, the limiter degrades to in-memory and logs at error level rather than refusing to boot.

### Keying authenticated requests by user

Address-only keying puts everyone behind a single NAT egress — an office, a campus, a mobile carrier — into one bucket, where one heavy user throttles the rest. Authenticated requests key on `sub` instead. Junk or expired credentials fall through to the address, so sending a garbage `Authorization` header is not a way to dodge the limit.

The token is read directly rather than via the auth dependency because SlowAPI runs the key function before dependency resolution.

### Two more sites with the same root cause

`audit_context.py` and `structured_logging.py` both did:

```python
ip = forwarded_for.split(",")[0].strip()
```

That is the **leftmost** entry — the one the client fully controls. Every audit record and request log carried whatever IP the caller claimed, which makes the audit trail forgeable. Both now go through the same resolution.

### Also

`rate_limit.py` had the `is_testing and not force_rate_limits` ternary copied twelve times. It is one `_limit()` helper now; the values it produces are unchanged.

---

## Testing

- [x] Tested locally
- [x] Existing tests pass
- [x] Added new tests
- [ ] UI verified

`backend/tests/test_client_address.py` — 27 tests, weighted toward the cases where the header must be *ignored*:

- a forged header from an untrusted peer is ignored; ten different forged values from one peer all resolve to the same key, so they all count against the same bucket
- with nothing configured, the header is ignored entirely (i.e. the default is the old behaviour)
- a genuine proxy's header is honoured; multi-hop chains pick the rightmost untrusted hop; a client-supplied prefix is discarded; an all-trusted chain falls back to the peer; junk stops the walk
- ports, IPv6 brackets and bare IPv6 parse correctly
- `X-Real-IP` is used only when `X-Forwarded-For` is absent, and only from a trusted peer
- malformed CIDRs are skipped, not fatal; a bare host is accepted as a CIDR
- rate-limit keys: anonymous by address, authenticated by user, two users never share a key, junk credentials fall back, and two untrusted clients get distinct keys (the regression)

```
backend $ python -m pytest tests/test_client_address.py -q
27 passed
```

Full backend suite before and after this branch: **the same 86 failures, an identical set** — all pre-existing on `main`, none introduced here.

---

## Checklist

---

## Additional Notes

**Deployments need to set `TRUSTED_PROXY_CIDRS` to get the benefit.** Both new settings are documented in `backend/.env.example` with the values for a Kubernetes ingress, a Compose bridge network and a single host proxy. Left unset, behaviour is exactly what it is on `main` today — this PR cannot make anything worse than it already is, but it does need one line of configuration to make it better.

I have not touched the `request.client.host` reads in the routers (`auth.py`, `projects.py`, `mfa.py` and others). Those record an IP alongside a login or a project event and deserve the same treatment, but they are a wider change and I would rather land the security-relevant path first. Happy to follow up if reviewers want them in scope.


nensii21:

PR #1175, opened 2026-08-14
fix(cache): key cached responses by caller and bound the L1 map

### LINKED ISSUE #1170: @cached omits the caller from the cache key, so per-user responses leak between accounts
### ✦ Describe the Bug

The `@cached` decorator in `backend/app/core/cache.py` builds its cache key by dropping the request-scoped dependencies:

```python
for k, v in kwargs.items():
    if k in ["db", "request", "response", "current_user"]:
        continue
```

`current_user` is discarded, so two different users calling the same endpoint with the same query parameters compute the **same cache key** and share the same cached response.

### ⟡ Steps to Reproduce

1. Apply `@cached(ttl=300)` to any endpoint whose result depends on `current_user` — a personalised feed, recommendations, "my projects", unread counts.
2. Sign in as user A and call it. The response is stored under a key that does not mention A.
3. Sign in as user B and call it with the same arguments within the TTL.
4. User B receives user A's response.

### ✦ Expected Behavior

A cache key must include every input the cached value depends on. If a function's result varies by caller, the caller has to be part of the key — or the function must be explicitly declared as returning caller-independent data.

### ⟡ Root Cause / Impact

The docstring calls this behaviour correct — *"Correctly ignores FastAPI dependencies (Session, Request, Response, User) when generating cache keys"* — which is true for `db`, `request` and `response` (they carry no semantic input) but not for `current_user`, which is precisely a semantic input. Excluding it silently converts any per-user cached endpoint into a cross-account data leak, and the decorator gives the caller no way to opt in to per-user keying.

Three further problems in the same module:

**L1 never evicts.** `self._l1_cache` only ever shrinks when an expired key is read again (`get`) or explicitly deleted. A key written once and never re-read stays in the dict forever. Because the key embeds stringified arguments, a decorated endpoint with a free-text or paginated parameter mints a fresh key per distinct argument — unbounded growth in a long-lived process.

**The dict is mutated without a lock.** Sync FastAPI endpoints run in a thread pool, so `get` and `set` are called concurrently. `del self._l1_cache[key]` in `get` can raise `KeyError` when two threads expire the same key at once.

**There is no way to invalidate.** `delete(key)` needs the exact key, but keys are generated inside the decorator from stringified arguments, so a caller who mutates the underlying data has no way to construct the keys it should invalidate. Cached data can only be waited out.

Minor: `disconnect()` closes the Redis client but leaves `self._redis_client` set, so later calls use a closed connection; and the L1 rehydration on an L2 hit hardcodes a 60-second TTL rather than reading the remaining TTL from Redis, so L1 can outlive the L2 entry it was derived from.

### ✅ Acceptance Criteria

- [ ] `@cached` includes the caller's identity in the key by default
- [ ] An explicit opt-out exists for genuinely public, caller-independent responses
- [ ] Cache keys are hashed to a bounded length rather than embedding raw stringified arguments
- [ ] L1 is bounded — a maximum entry count with LRU eviction, plus removal of expired entries
- [ ] L1 mutations are guarded by a lock
- [ ] A prefix/namespace invalidation path exists so writes can clear what they invalidated
- [ ] `disconnect()` clears the client reference
- [ ] L1 rehydration honours the remaining L2 TTL
- [ ] Tests: two users do not share a cached value; eviction bounds the map; concurrent access does not raise

MOHITKOURAV01:
# Pull Request

## Summary

The `@cached` decorator dropped `current_user` from the cache key, so any endpoint whose response varies by caller served the first caller's response to everyone else for the length of the TTL. This makes the caller part of the key by default, and fixes four other problems in the same module — an L1 map that never evicted, unguarded concurrent mutation, no way to invalidate anything, and unbounded key length.

---

## Related Issue

Closes #1170

---

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation update
- [x] Performance improvement
- [ ] Refactoring
- [ ] UI/UX enhancement
- [x] Tests
- [ ] Other

---

### The caller belongs in the key

```python
for k, v in kwargs.items():
    if k in ["db", "request", "response", "current_user"]:
        continue
```

The docstring called this correct — *"Correctly ignores FastAPI dependencies (Session, Request, Response, User)"* — and it is, for three of the four. `db`, `request` and `response` carry no semantic input; including them would give every request a unique key and disable caching entirely. `current_user` is the opposite: it is *nothing but* semantic input.

`per_user` now defaults to `True`. `per_user=False` is the opt-out, and the four existing call sites are audited and marked explicitly with a one-line justification each. That is deliberately more typing than the old behaviour: it makes the claim reviewable per endpoint, rather than the decorator making it silently for all of them at once.

An anonymous caller resolves to a shared `anon` identity, which is correct — they are indistinguishable to the endpoint. A `current_user` with no `id` raises rather than falling back to `str(user)`, which would produce `<User object at 0x...>` — unique per instance, so caching would silently stop working instead of failing visibly.

### L1 never evicted

It only shrank when an expired key happened to be read again. A key written once and never re-read stayed for the life of the process, and since keys embed the call's arguments, a paginated or free-text endpoint mints a new one per distinct argument. It is now an `OrderedDict` used as an LRU, bounded at 1000 entries — a bound on entry *count*, not bytes, because measuring arbitrary payloads costs more than it saves and the count is what keeps "grows forever" off the table.

### Unguarded concurrent mutation

Sync FastAPI endpoints run in a thread pool, so every L1 mutation is concurrent. `del self._l1_cache[key]` could raise `KeyError` when two threads expired the same key at once. All mutations now take a lock, and removal is `pop(key, None)`.

### Nothing could be invalidated

`delete(key)` needs the exact key, but the decorator builds keys internally, so a caller that mutated the underlying data had no way to reconstruct what to drop. Stale data could only be waited out. `delete_prefix()` sweeps a namespace, using `SCAN` rather than `KEYS` — `KEYS` blocks the Redis event loop for the whole sweep, which on a shared instance stalls every other client.

### Unbounded key length

Keys interpolated the stringified arguments, so a long search term produced a proportionally long key in both levels. The arguments are now a fixed-width SHA-256 prefix, and kwargs are sorted so the same call spelled two ways shares one entry instead of getting two.

### Smaller ones

- `disconnect()` closed the Redis client but kept the reference, so anything reaching the cache after shutdown used a closed connection.
- An L2 hit promoted into L1 with a flat 60s TTL, which could outlive the L2 entry it came from — leaving one process serving a value the rest of the fleet had dropped. Promotion honours the remaining TTL, capped at 60s.
- Promotion is now best-effort in its own guard. Folded into the read as it was, a failure there discarded a value that had already been decoded and was about to be returned.

---

## Testing

- [x] Tested locally
- [x] Existing tests pass
- [x] Added new tests
- [ ] UI verified

`backend/tests/test_cache.py` — 26 tests. Redis is not assumed to be running; `_redis_client` stays `None` (the documented degraded mode) and the L2 paths are driven with a small fake.

- **The leak:** two users do not share a value; the same user is served from cache; `per_user=False` shares one entry; anonymous callers share one; a user without an `id` raises with an actionable message
- **Keys:** prefix/function/caller structure, distinct arguments distinct keys, kwarg order irrelevant, plumbing excluded, length bounded regardless of input size
- **Eviction:** L1 is bounded at its limit under 50 writes; the least recently used entry goes first; expired entries are not returned and are dropped on read
- **Invalidation:** `delete`, `delete_prefix` across a namespace, deleting an absent key, `clear_l1`
- **Concurrency:** 8 threads × 200 mixed set/get/delete raises nothing; 8 threads reading the same expired key all hit the removal path
- **Redis:** promotion respects the remaining TTL and is capped; errors degrade to L1; `disconnect` clears the reference; `delete_prefix` reaches L2

```
backend $ python -m pytest tests/test_cache.py tests/test_api_caching.py -q
37 passed
```

Full backend suite before and after this branch: **the same 86 failures, an identical set** — all pre-existing on `main`, none introduced here.

### One existing test changed

`test_api_caching.py::test_caching_decorator_ignores_dependencies` asserted `"hello" in key`, i.e. that the argument's text appears verbatim in the key. Hashing the fingerprint deliberately changes that. Its assertions now cover the same intent against the new format: plumbing still excluded, the key scoped to endpoint and caller, and distinct arguments producing distinct entries.

Two of its sibling tests were also failing on this branch before I fixed two real bugs they caught — `current_user` arriving as a plain `dict` rather than a model, and a `MagicMock` Redis client whose `ttl()` return value is not numeric. Both are handled now (mappings are supported; a non-numeric TTL falls back to the default rather than discarding the value).

---

## Checklist

---

## Additional Notes

**Cache keys change shape, so existing entries are orphaned by this deploy.** They expire on their own TTL (60–300s for the current call sites) and nothing reads them in the meantime, so the effect is one cold cache period. No migration needed.

**A related problem I did not fix here:** `get_project` and `get_project_by_slug` call `ProjectAnalyticsService.record_view` *inside* a cached function, so the view is only recorded on a cache miss — project view counts are silently under-counting by the cache hit rate. Caching a function with a side effect is the actual bug there, and the fix is to move `record_view` out of the cached call rather than to change the decorator. I left it alone to keep this PR to one subject; happy to open a separate issue for it if reviewers agree it is worth tracking.

### DIFF
--- a/backend/app/core/cache.py
+++ b/backend/app/core/cache.py
@@ -1,8 +1,34 @@
+"""
+Two-level cache: an in-process L1 in front of a shared Redis L2.
+
+Three rules shape everything here.
+
+**A cache key must name every input the value depends on.** The obvious ones
+are the function's arguments. The one that is easy to miss is *who is asking*
+-- and missing it is not a stale-data bug, it is one user being served
+another user's response. ``current_user`` used to be dropped from the key
+along with ``db`` and ``request``, on the grounds that all three are FastAPI
+dependencies. But ``db`` and ``request`` carry no semantic input, and
+``current_user`` is nothing but semantic input. See :func:`cached`.
+
+**A cache that cannot forget is a memory leak.** L1 keys embed the call's
+arguments, so an endpoint with a free-text or paginated parameter mints a new
+key per distinct argument. Without a bound and an eviction policy the dict
+grows for the lifetime of the process.
+
+**Sync FastAPI endpoints run in a thread pool.** Every mutation of the L1
+mapping is therefore concurrent and has to be guarded.
+"""
+
+import hashlib
 import json
 import logging
+import threading
 import time
+from collections import OrderedDict
+from collections.abc import Mapping
 from functools import wraps
-from typing import Any, Callable, Dict, Optional, Tuple
+from typing import Any, Callable, Optional, Tuple
 
 import redis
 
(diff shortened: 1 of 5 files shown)


nensii21:

PR #1176, opened 2026-08-14
fix(ws): stop carrying rooms and identity across a session boundary

### LINKED ISSUE #1171: WebSocket client re-joins the previous session's project rooms and revives itself after logout
### ✦ Describe the Bug

`frontend/src/api/ws.ts` keeps a module-level singleton `WsClient` whose lifecycle does not survive a sign-out. Four distinct problems, all in the same object:

**1. Project rooms outlive the session.** `joinedProjects` is only ever added to by `joinProject` and removed from by `leaveProject`. `disconnect()` does not clear it. The socket re-joins on open:

```ts
this.socket.onopen = () => {
  this.reconnectAttempts = 0;
  for (const projectId of this.joinedProjects) {
    this.send({ type: "join", project_id: projectId });
  }
};
```

Sign out and sign in as a different user in the same tab, and the new session immediately joins every project room the previous user was in.

**2. `on()` revives a socket that was deliberately closed.** `disconnect()` sets `manualClose = true` and nulls the socket, but `on()` reconnects on a null socket and `connect()` resets `manualClose = false`:

```ts
on(handler: Handler): () => void {
  this.handlers.add(handler);
  if (!this.socket) this.connect();   // ← after logout, this reopens
  ...
}
```

Any component that mounts and subscribes after logout re-opens the connection.

**3. The reconnect backoff has no jitter.** `Math.min(30_000, 1000 * 2 ** attempts)` is deterministic, so every tab that was connected when the server restarted reconnects at the same millisecond, and keeps doing so in lockstep. `frontend/src/api/client.ts` already treats this as a bug worth solving for HTTP and applies full jitter, with a comment explaining why — the WS path never got the same treatment.

**4. Reconnection is unbounded and offline-blind.** There is no attempt ceiling and no check of `navigator.onLine`, so a backgrounded tab with no network retries every 30s indefinitely.

### ⟡ Steps to Reproduce

For the room leak:

1. Sign in as user A, open a project page (`ws.joinProject(...)` runs).
2. Sign out. `tokenStore.subscribe` fires `ws.disconnect()` — `joinedProjects` still holds the project id.
3. Sign in as user B in the same tab. `ws.connect()` runs and `onopen` sends `{type: "join", project_id: <A's project>}`.
4. User B is now in a room they never asked to join, and receives its events for as long as the server allows it.

### ✦ Expected Behavior

- Disconnecting clears the room set; a new session starts with no rooms.
- A manual disconnect stays disconnected until something explicitly reconnects.
- Reconnect delays are jittered.
- Reconnection gives up after a bounded number of attempts and resumes when the browser reports it is back online.

### ⟡ Root Cause / Impact

The client was written as a process-lifetime singleton, but sign-in and sign-out change identity underneath it. Whether the stale `join` actually delivers another user's events depends on how strictly the server authorises room membership — but the client should not be sending the request at all, and relying on the server to reject it is not a defence we should be depending on.

### ✅ Acceptance Criteria

- [ ] `disconnect()` clears `joinedProjects` and the pending reconnect timer
- [ ] A manually closed client is not revived by `on()`; reconnection requires an explicit `connect()`
- [ ] Reconnect backoff applies full jitter, consistent with `client.ts`
- [ ] A maximum reconnect attempt count, after which the client stops and reports state
- [ ] Reconnection pauses while `navigator.onLine` is false and resumes on `online`
- [ ] The socket is not opened at all when there is no access token
- [ ] Tests covering: rooms cleared across a sign-out/sign-in cycle; no revive after manual close; jitter applied; attempts bounded

MOHITKOURAV01:
# Pull Request

## Summary

`WsClient` is a module-level singleton, but sign-in and sign-out change identity underneath it. The client carried project rooms, an open socket and an authentication token across that boundary. This makes the session boundary actually reset the client, and brings the reconnect loop in line with what `client.ts` already does for HTTP.

---

## Related Issue

Closes #1171

---

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Refactoring
- [ ] UI/UX enhancement
- [x] Tests
- [ ] Other

---

### 1. Rooms outlived the session

`joinedProjects` was only ever emptied by `leaveProject`. `disconnect()` left it populated, and the socket re-joins on open:

```ts
this.socket.onopen = () => {
  for (const projectId of this.joinedProjects) {
    this.send({ type: "join", project_id: projectId });
  }
};
```

Sign out, sign in as somebody else in the same tab, and the new session's first `onopen` sent a `join` for every project the previous user had open. `disconnect()` now clears the set.

Whether that `join` actually delivers another user's events depends on how strictly the server authorises room membership — but the client should not be sending it, and relying on the server to reject it is not a defence worth depending on.

### 2. A deliberate close did not stay closed

```ts
on(handler) {
  this.handlers.add(handler);
  if (!this.socket) this.connect();   // and connect() resets manualClose
}
```

Any component mounting and subscribing after a sign-out reopened the connection. Subscribing now connects only when the client was not deliberately closed; `retry()` is the explicit way back.

### 3. The socket kept whatever token it was opened with

A WebSocket authenticates once, at the handshake, from the token in its URL — there is no way to re-authenticate an open one. `tokenStore.subscribe` called `ws.connect()` on every token change, but `connect()` returned early whenever a socket was already open, so the socket kept its original identity through a refresh and through a different user signing in.

The token the socket was opened with is now remembered, and `connect()` replaces the socket when it changes.

### 4. Reconnection was deterministic and unbounded

`Math.min(30_000, 1000 * 2 ** attempts)` has no jitter, so every tab that was connected when the server restarted reconnects on the same tick and keeps doing so in lockstep — the load pattern a recovering server least needs. `client.ts` already applies full jitter to HTTP retries, with a comment explaining why; the socket never got the same treatment. It does now.

Attempts are capped at 10 (roughly four minutes of real outage), reconnection pauses while `navigator.onLine` is `false` rather than burning the budget against a disconnected network, and an `online` listener resets the budget and resumes. Without that last part a tab that exhausted its attempts during an outage would stay dead until reloaded.

### Also

- The socket is no longer opened without an access token. Doing so only produced a handshake the server rejects, followed by the retry loop trying it again.
- Status is observable through `onStatus` / `getStatus` (`closed` / `connecting` / `open` / `reconnecting` / `gave-up`), so a caller can surface "gave up" rather than silently sitting disconnected.
- `closeSocket()` detaches the handlers before closing, so a deliberate replacement does not run the reconnect path meant for a dropped connection.

---

## Testing

- [x] Tested locally
- [x] Existing tests pass
- [x] Added new tests
- [ ] UI verified

`frontend/src/api/__tests__/ws.test.ts` — 27 tests against a fake `WebSocket` that lets a test complete a handshake, drop a connection, deliver a message, or fail to construct:

- **Rooms:** no re-join after a sign-out/sign-in cycle (the regression); own rooms re-joined after a drop; a room that was left is not re-joined; explicit disconnect clears them
- **Manual close:** not revived by subscribing; schedules no reconnect; reopens on `retry()`
- **Tokens:** socket opened with the current token; replaced token reconnects; the same token does not; no socket without a token
- **Reconnection:** jitter applied and capped; two clients dropping together get different delays; reconnects after a drop; gives up at the limit and stays given up; the budget resets on a successful open; construction failure schedules a retry and recovers; offline does not burn attempts; `online` resumes
- **Events:** parsed delivery, malformed messages ignored, unsubscribe stops delivery, no send on a socket that is not open
- **Status:** current status replayed to a new subscriber, transitions, unsubscribe

```
frontend $ npx vitest run
Test Files  41 passed (41)
     Tests  636 passed (636)

frontend $ npx tsc --noEmit
(no errors)
```

---

## Checklist

---

## Additional Notes

**Nothing imports `api/ws.ts` yet.** `useChatWebSocket` and `useCollaborativeWorkspace` open raw `WebSocket`s of their own and are untouched by this. So this is not fixing a bug users are hitting today — it is fixing one that lands the moment somebody wires this module up, which is a much cheaper time to fix it. I would rather say that plainly than imply a live incident.

The two hooks that do use raw sockets have their own reconnect loops with the same missing jitter. Folding them onto this client would be the right end state, but it is a behavioural change to two working features and does not belong in the same PR. Happy to open a follow-up issue.

### DIFF
--- a/frontend/src/api/__tests__/ws.test.ts
+++ b/frontend/src/api/__tests__/ws.test.ts
@@ -1,0 +1,478 @@
+/**
+ * WebSocket client lifecycle.
+ *
+ * The client is a module-level singleton, but sign-in and sign-out change
+ * identity underneath it. Most of what is asserted here is about what the
+ * client does *across* that boundary: which rooms it re-joins, whether a
+ * deliberate close stays closed, and whether the socket keeps the identity it
+ * was opened with after the token changes.
+ *
+ * `VITE_API_BASE_URL` has to be set before `client.ts` is imported, since it
+ * reads the env var at module scope. Hence the dynamic imports.
+ */
+
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+
+// ── A WebSocket good enough to assert against ────────────────────────────────
+
+class FakeWebSocket {
+  static CONNECTING = 0;
+  static OPEN = 1;
+  static CLOSING = 2;
+  static CLOSED = 3;
+
+  /** Every socket constructed during a test, in order. */
+  static instances: FakeWebSocket[] = [];
+
+  /** Constructor throws once, for the "cannot open" path. */
+  static throwOnNextConstruct = false;
+
+  readyState = FakeWebSocket.CONNECTING;
+  sent: string[] = [];
+  closeCalls = 0;
+
+  onopen: (() => void) | null = null;
+  onmessage: ((e: { data: string }) => void) | null = null;
+  onclose: (() => void) | null = null;
+  onerror: (() => void) | null = null;
+
+  constructor(public url: string) {
+    if (FakeWebSocket.throwOnNextConstruct) {
+      FakeWebSocket.throwOnNextConstruct = false;
+      throw new Error("cannot open");
+    }
+    FakeWebSocket.instances.push(this);
+  }
+
+  send(data: string) {
+    this.sent.push(data);
+  }
+
+  close() {
+    this.closeCalls += 1;
+    const wasOpen = this.readyState !== FakeWebSocket.CLOSED;
+    this.readyState = FakeWebSocket.CLOSED;
+    if (wasOpen) this.onclose?.();
+  }
+
+  // -- test helpers --
+
+  /** Complete the handshake. */
+  open() {
+    this.readyState = FakeWebSocket.OPEN;
+    this.onopen?.();
+  }
+
+  /** Drop the connection the way a server restart would. */
+  dropped() {
+    this.readyState = FakeWebSocket.CLOSED;
+    this.onclose?.();
+  }
+
+  deliver(payload: unknown) {
+    this.onmessage?.({ data: JSON.stringify(payload) });
+  }
+
+  /** The messages sent on this socket, parsed. */
+  get messages(): Array<Record<string, unknown>> {
+    return this.sent.map((raw) => JSON.parse(raw));
+  }
+
+  static get latest(): FakeWebSocket {
+    return FakeWebSocket.instances[FakeWebSocket.instances.length - 1];
+  }
+
+  static reset() {
+    FakeWebSocket.instances = [];
+    FakeWebSocket.throwOnNextConstruct = false;
+  }
+}
+
+// ── Harness ──────────────────────────────────────────────────────────────────
+
+let ws: typeof import("../ws").ws;
+let reconnectDelay: typeof import("../ws").reconnectDelay;
+let tokenStore: typeof import("../tokens").tokenStore;
+
+async function loadModules() {
+  vi.resetModules();
+  vi.stubEnv("VITE_API_BASE_URL", "http://api.test");
+
+  const tokensModule = await import("../tokens");
+  tokenStore = tokensModule.tokenStore;
+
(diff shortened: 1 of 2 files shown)


nensii21:

PR #1177, opened 2026-08-14
fix(auth): propagate sign-out and session changes across tabs

### LINKED ISSUE #1172: Signing out in one tab leaves other tabs authenticated; no cross-tab session sync
### ✦ Describe the Bug

`frontend/src/api/tokens.ts` splits the session across three places with different scopes:

| Where | What | Scope |
|---|---|---|
| module variable `accessToken` | access token | per JS context (per tab) |
| `sessionStorage` | access token | per tab |
| `localStorage` | refresh token | **shared across all tabs** |

`localStorage` is shared and emits a `storage` event to other tabs when it changes — but nothing in the frontend listens for it. `grep -rn '"storage"' frontend/src` returns nothing.

So a session change in one tab is invisible to every other tab.

### ⟡ Steps to Reproduce

**Sign-out does not propagate:**

1. Sign in, then open the app in a second tab.
2. Sign out in tab A. `tokenStore.clear()` removes the refresh token from `localStorage` and the access token from tab A.
3. Tab B still holds its own `accessToken` in memory and in its own `sessionStorage`. It stays fully signed in and keeps making authorised requests until the access token expires.
4. When it does expire, `refreshAccessToken()` reads `localStorage`, finds nothing, and returns `null` — so tab B finally drops out, at a moment unrelated to the logout, with whatever request happened to be in flight failing.

**Identity can diverge:**

1. Sign in as user A in tab A and tab B.
2. In tab A, sign out and sign in as user B. `localStorage` now holds user B's refresh token.
3. Tab B still shows user A's UI and still holds user A's access token.
4. When tab B's access token expires it refreshes against user B's refresh token and silently becomes user B — with a page still rendered from user A's data.

### ✦ Expected Behavior

- Signing out in one tab signs out every tab, promptly.
- Signing in as a different user in one tab does not leave another tab rendering the previous user's data.

### ⟡ Root Cause / Impact

The comment at the top of the file explains the design — access in memory to limit XSS exposure, refresh in `localStorage` so the session survives a reload — and that reasoning is sound. The gap is that `localStorage` was chosen for its cross-tab persistence without wiring up the cross-tab *notification* that comes with it.

The security consequence is the one users assume is covered: "log out everywhere on this machine" is what people expect logout to mean, and on a shared computer the second tab remains a live authenticated session after the user believes they have signed out. The correctness consequence is the identity divergence above, where a tab keeps rendering one user's data while authenticating as another.

Note that `storage` events do not fire in the tab that made the change, so the handler naturally only affects other tabs. Care is needed not to treat an unrelated `localStorage` write as a session change, and not to loop by writing back to storage from inside the handler.

### ✅ Acceptance Criteria

- [ ] A `storage` listener watches the refresh-token key and reacts to changes from other tabs
- [ ] Refresh token removed elsewhere → this tab clears its in-memory and `sessionStorage` access token and notifies subscribers
- [ ] Refresh token replaced with a different value elsewhere → this tab drops its stale access token rather than pairing it with another user's refresh token
- [ ] Writes originating in this tab do not re-enter the handler or loop
- [ ] The listener is registered once, is SSR-safe, and can be torn down
- [ ] Existing `subscribe()` consumers (including the `ws` auto-connect wiring) see the change through the normal notification path
- [ ] Tests simulating a `storage` event for: cleared token, replaced token, unrelated key

MOHITKOURAV01:
# Pull Request

## Summary

The refresh token lives in `localStorage`, which is shared across every tab on the origin and fires a `storage` event to the others when it changes. Nothing listened for it, so a session change in one tab was invisible to the rest — signing out in one tab left the others fully authenticated.

---

## Related Issue

Closes #1172

---

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Refactoring
- [ ] UI/UX enhancement
- [x] Tests
- [ ] Other

---

### The gap

The session lives in three places with three different scopes:

| Where | What | Scope |
|---|---|---|
| module variable `accessToken` | access token | per tab |
| `sessionStorage` | access token | per tab |
| `localStorage` | refresh token | **shared across all tabs** |

The comment at the top of `tokens.ts` explains the design — access in memory to limit XSS exposure, refresh in `localStorage` so the session survives a reload — and that reasoning is sound. The gap is that `localStorage` was chosen for its cross-tab *persistence* without wiring up the cross-tab *notification* that comes with it. `grep -rn '"storage"' frontend/src` returned nothing.

**Sign-out did not propagate.** Tab A calls `tokenStore.clear()`, which removes the refresh token from shared storage. Tab B still holds its own access token in memory and in its own `sessionStorage`, so it stays signed in and keeps making authorised requests until that token expires — at which point it drops out at a moment unrelated to the logout, with whatever request was in flight failing. On a shared machine, "log out" did not mean what the user assumed.

**Identity could diverge.** Sign out in tab A and sign in as a different user. Tab B still renders user A and still holds A's access token. When it expires, `refreshAccessToken()` reads `localStorage`, finds **B's** refresh token, and tab B silently becomes user B — with a page still full of A's data.

### The fix

`tokens.ts` listens for `storage` on the refresh key:

- **Removed elsewhere** (or `localStorage.clear()`, which arrives as `key === null`) → clear this tab's access token and notify subscribers.
- **Replaced with a different value** → clear the stale access token but leave the new refresh token alone. It belongs to the new session and is valid; only this tab's access token is stale. This tab becomes unauthenticated and re-establishes identity from the new refresh token rather than inheriting one from the previous session. Dropping it is the better of the two failure modes — the alternative is the silent identity swap above.
- **A different key, or a write that changed nothing** → ignored, so an unrelated `localStorage` write does not sign the tab out.

`storage` does not fire in the tab that made the change, so a local `set()` notifies exactly once and the listener cannot loop. There is a test for that specifically.

Registration is idempotent, SSR-safe, and returns a teardown function.

### `useAuth` had to be wired up too

```ts
useEffect(() => {
  if (!tokenStore.getAccess()) { setUser(null); return; }
  api.get<AuthUser>("/api/users/me")...
}, []);   // ← empty deps
```

It read the token once on mount and never subscribed, so none of the above would have reached the UI. It now re-runs the lookup when the token changes. That also fixes the same-tab case, which was already broken independently: a token refresh or a sign-in went unnoticed by this hook too.

### Also

`subscribe()` returned `listeners.delete(l)` directly, i.e. a boolean. An unsubscribe function that returns a value is easy to misread as meaningful and misbehaves where a cleanup returning `void` is expected — a React effect, for one. It returns `void` now.

---

## Testing

- [x] Tested locally
- [x] Existing tests pass
- [x] Added new tests
- [ ] UI verified

`frontend/src/api/__tests__/tokens.test.ts` — 20 tests. A real two-tab scenario cannot be reproduced in one jsdom window, so the tests write the value and then dispatch the `storage` event the browser would have delivered — the same order and the same resulting state a second tab leaves behind.

- **Storage:** the access token stays out of `localStorage`; it survives a reload via `sessionStorage`; an omitted refresh argument leaves the refresh token alone; `clear` removes both; subscribers notified on set and clear; unsubscribe returns nothing and stops delivery
- **Sign-out elsewhere:** this tab's access token is cleared (the regression); subscribers notified; `localStorage.clear()` counts; already-signed-out is a no-op
- **A different session elsewhere:** the stale access token is dropped, the new refresh token is left in place, subscribers told this tab is unauthenticated
- **Must not disturb the session:** a different key, a no-change write, this tab's own writes (notified exactly once, not twice), and a subscriber that reads back during notification does not loop
- **Lifecycle:** teardown stops reacting; registering twice does not double-handle an event

```
frontend $ npx vitest run
Test Files  41 passed (41)
     Tests  629 passed (629)

frontend $ npx tsc --noEmit
(no errors)

frontend $ npx eslint src/api/tokens.ts src/api/__tests__/tokens.test.ts src/contexts/auth-context.tsx
(clean)
```

---

## Checklist

---

## Additional Notes

**On the "replaced" case being a sign-out locally.** The alternative is to keep the tab signed in and let it silently adopt the new identity. I went with signing it out because the tab's rendered content belongs to the previous user, and adopting an identity the user did not choose *in that tab* is the more surprising outcome — it is the exact failure described in the issue. The tab re-authenticates from the new refresh token on its next request, so a user who genuinely wants both tabs on the new account gets there with one reload rather than seeing a mismatched page.

**Still not fixed, and out of scope here:** logout is local-only — it drops the tokens but does not tell the server, so a refresh token that leaked before the logout stays valid until it expires. That needs a server-side revocation list keyed on the refresh token's `jti` (which `create_refresh_token` already mints). Worth its own issue.

### DIFF
--- a/frontend/src/api/__tests__/tokens.test.ts
+++ b/frontend/src/api/__tests__/tokens.test.ts
@@ -1,0 +1,290 @@
+/**
+ * Token storage and cross-tab session sync.
+ *
+ * `storage` events do not fire in the tab that made the change, so a real
+ * two-tab scenario cannot be reproduced in one jsdom window. These tests
+ * dispatch the event the browser would have delivered, which is the same
+ * thing from the listener's point of view — and they assert that the local
+ * writes this tab makes itself do *not* go through that path, since a
+ * listener that reacted to its own writes would loop.
+ */
+
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+
+const REFRESH_KEY = "devlink.refresh";
+const ACCESS_KEY = "devlink.access";
+
+let tokenStore: typeof import("../tokens").tokenStore;
+let stopSessionSync: typeof import("../tokens").stopSessionSync;
+
+/**
+ * Fire the `storage` event another tab's write would have produced.
+ *
+ * jsdom does not propagate storage writes into events, so the value is
+ * written first and the event dispatched second — which is the order and the
+ * state a real second tab would leave behind.
+ */
+function storageEventFromAnotherTab(
+  key: string | null,
+  newValue: string | null,
+  oldValue: string | null,
+) {
+  if (key === null) {
+    window.localStorage.clear();
+  } else if (newValue === null) {
+    window.localStorage.removeItem(key);
+  } else {
+    window.localStorage.setItem(key, newValue);
+  }
+
+  window.dispatchEvent(
+    new StorageEvent("storage", {
+      key,
+      newValue,
+      oldValue,
+      storageArea: window.localStorage,
+    }),
+  );
+}
+
+beforeEach(async () => {
+  vi.resetModules();
+  window.localStorage.clear();
+  window.sessionStorage.clear();
+
+  const module = await import("../tokens");
+  tokenStore = module.tokenStore;
+  stopSessionSync = module.stopSessionSync;
+});
+
+afterEach(() => {
+  stopSessionSync?.();
+});
+
+// ── Basic storage ────────────────────────────────────────────────────────────
+
+describe("storage", () => {
+  it("keeps the access token out of localStorage", () => {
+    tokenStore.set("access", "refresh");
+
+    expect(window.sessionStorage.getItem(ACCESS_KEY)).toBe("access");
+    expect(window.localStorage.getItem(ACCESS_KEY)).toBeNull();
+    expect(window.localStorage.getItem(REFRESH_KEY)).toBe("refresh");
+  });
+
+  it("reads the access token back from sessionStorage after a reload", async () => {
+    window.sessionStorage.setItem(ACCESS_KEY, "survived");
+    vi.resetModules();
+
+    const { tokenStore: reloaded } = await import("../tokens");
+
+    expect(reloaded.getAccess()).toBe("survived");
+  });
+
+  it("leaves the refresh token alone when it is not passed", () => {
+    tokenStore.set("access-1", "refresh-1");
+    tokenStore.set("access-2");
+
+    expect(tokenStore.getRefresh()).toBe("refresh-1");
+    expect(tokenStore.getAccess()).toBe("access-2");
+  });
+
+  it("clear removes both", () => {
+    tokenStore.set("access", "refresh");
+    tokenStore.clear();
+
+    expect(tokenStore.getAccess()).toBeNull();
+    expect(tokenStore.getRefresh()).toBeNull();
+  });
+
+  it("notifies subscribers on set and clear", () => {
+    const seen: (string | null)[] = [];
+    const off = tokenStore.subscribe((t) => seen.push(t));
+
+    tokenStore.set("access", "refresh");
+    tokenStore.clear();
+
+    expect(seen).toEqual(["access", null]);
+    off();
+  });
+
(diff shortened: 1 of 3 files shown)


nensii21:

Issue #1197, opened 2026-08-16
RBAC defines its project permission table twice, and system maintainers get a blanket grant over every org and project


MOHITKOURAV01:
### ✦ Describe the Bug

`backend/app/core/rbac.py` defines `PROJECT_ROLE_PERMISSIONS` **twice**, back to back:

```python
PROJECT_ROLE_PERMISSIONS: dict[MemberRole, frozenset[str]] = {   # line 189
    MemberRole.OWNER: frozenset({PROJECT_UPDATE, PROJECT_DELETE, ...}),
    ...
}

PROJECT_ROLE_PERMISSIONS = {                                      # line 214
    MemberRole.OWNER: {PROJECT_UPDATE, PROJECT_DELETE, ..., PROJECT_MANAGE_ROLES, ...},
    ...
}
```

The second binding overwrites the first. Twenty-four lines of typed, annotated permission table at line 189 are dead — never read, never executed, and never able to disagree with the live table loudly enough for anyone to notice. The two are already out of sync: the dead table gives `MemberRole.MAINTAINER` three permissions, the live one gives it seven, and the live one adds `CONTRIBUTOR`, `REVIEWER` and `VIEWER` roles that the dead one does not know about. Whichever was intended, the file currently claims both.

This is a maintenance trap: the next person to grant a permission has a 50% chance of editing the block that does nothing.

### ⟡ System `MAINTAINER` is a blanket grant over every org and every project

`has_org_permission` and `has_project_permission` both short-circuit here:

```python
if has_system_role(db, user_id, SystemRole.ADMIN, SystemRole.MAINTAINER):
    return True
```

`return True` — for *any* `permission` argument. The system-level table says a `MAINTAINER` holds exactly `{manage_content, view_analytics, contribute}`. But because these two functions never consult that table, a maintainer can pass `ORG_DELETE` and `PROJECT_DELETE` and be granted. The module docstring advertises the intent as *"platform staff can manage any org"*; what is implemented is "platform staff can delete any org". A content moderator is one call away from destroying an organisation.

The same shape appears at the bottom of `has_project_permission`:

```python
if org_member and org_member.role in (OrgMemberRole.OWNER, OrgMemberRole.ADMIN):
    return True
```

An org admin inherits *every* project permission in the org, including `project:transfer_ownership` on a project they do not own.

### ✦ `get_user_permissions` returns a set with no scope, and the UI trusts it

```python
for pm in project_members:
[...]

PR #914, opened 2026-08-06
Feature/context aware quick actions

dristicg:
**Title:** feat(ui): implement context-aware quick actions

**Description:**
This PR replaces the static quick action cards on the dashboard with dynamic, personalized actions, improving the user experience with context-aware recommendations.

**Changes:**
- Added a `QuickAction` interface and mocked personalized actions (e.g., "Continue current project", "Review applications", "Publish flare", "Invite recommended builder") in the frontend seed data.
- Updated `dashboardService` to expose a `quickActions()` API function with a fallback mechanism.
- Refactored the `QuickActions` component in `sections.tsx` to fetch dynamic data and dynamically map Lucide icons (e.g., `FolderPlus`, `Users2`, `UserPlus`, `Flame`).

**Screenshot**
image
image

Closes #746

PR #1199, opened 2026-08-16
fix(backend): repair merge-corrupted modules so the app imports again

### LINKED ISSUE #1194: Backend cannot start: three modules on main have Python syntax errors after the voice/video introduction merge
### ✦ Describe the Bug

`main` does not compile. Three backend modules contain Python syntax errors, and one of them — `app/routers/users.py` — is registered on the v1 router, so the application cannot import, let alone serve a request.

```
$ python -m compileall -q backend/app
backend/app/routers/users.py:525    SyntaxError: invalid syntax
backend/app/schemas/user.py:96      IndentationError: unexpected indent
backend/app/utils/uploads.py:117    SyntaxError: '(' was never closed
```

All three are the same accident: the voice-introduction feature (#977, merged as #1164) and the video-introduction feature (#978, merged as #1183) touch the same functions, and the merge in `e2be7a58` ("Merge branch 'main' into feat/video-introduction-978") kept **both sides of every hunk, interleaved**, instead of resolving them. No conflict markers were left behind, which is why nothing flagged it — the result looks like ordinary code until you try to parse it.

### ⟡ Steps to Reproduce

```
git checkout main
cd backend && python -c "import app.main"
```

```
File "backend/app/routers/users.py", line 525
    try:
    ^^^
SyntaxError: invalid syntax
```

Or just start the server — `uvicorn app.main:app` fails on import.

### ✦ What the corruption looks like

**`app/routers/users.py`** — `upload_voice_introduction` lost its entire body, and `upload_video_introduction` has two function bodies spliced into one:

```python
async def upload_voice_introduction(...):
    if not file.filename:
        raise HTTPException(status_code=400, detail="No file provided")
                                    # <- body ends here; nothing is saved

@router.post("/me/video-introduction", ...)
async def upload_video_introduction(...):
    ...
    contents = await file.read()

    try:
        validate_video_introduction_upload(     # <- call never closed
    contents = await file.read()                # <- second copy of the read

    try:
        validate_voice_introduction_upload(
            ...
        )
    ...
    video_url = save_video_introduction_upload(  # <- call never closed
    voice_url = save_voice_introduction_upload(
        ...
    )

    full_video_url = ...
    return UserService.update_video_introduction_url(
        db,
        current_user,
        full_video_url,                          # <- call never closed
    full_voice_url = ...
    return UserService.update_voice_introduction_url(...)
```

**`app/utils/uploads.py`** — `save_video_introduction_upload` was consumed by the following `def`:

```python
def save_video_introduction_upload(
def validate_voice_introduction_upload(
    filename: str | None, content_type: str | None, size_bytes: int
) -> None:
```

and further down, two directories and two returns stacked on top of each other, so the second of each pair is dead:

```python
    upload_dir = Path(settings.UPLOAD_DIR) / "video_introductions"
    upload_dir = Path(settings.UPLOAD_DIR) / "voice_introductions"
    ...
    return f"/uploads/video_introductions/{safe_name}"
    return f"/uploads/voice_introductions/{safe_name}"   # unreachable
```

**`app/schemas/user.py`** — a field indented one level too deep inside `UserBase`:

```python
    resume_url: ValidURL | None = None
      voice_introduction_url: ValidURL | None = None
    portfolio_url: ValidURL | None = None
```

### ⟡ Impact

- The backend does not start. Every endpoint is down, not just the two upload routes.
- `pytest` collects nothing that transitively imports `app.main`, so CI's "tests passed" is reporting on a suite that never ran the API. This is the same class of false-green described in #1090.
- `save_video_introduction_upload` no longer exists as a callable, and `UserBase.voice_introduction_url` never made it into the schema.

Both features are effectively unshipped even though both PRs are merged.

### ✦ Expected Behavior

`main` imports, the server boots, and both upload endpoints work independently:

- `POST /api/v1/users/me/voice-introduction` validates and stores an audio file under `uploads/voice_introductions/`
- `POST /api/v1/users/me/video-introduction` validates and stores a video file under `uploads/video_introductions/`

### ⟡ Also worth fixing while we are here

`check_pr.py` at the repository root is stored as UTF-16 (it starts with a `FF FE` BOM), so `python check_pr.py` dies with *"source code string cannot contain null bytes"*. It is a stray developer script rather than application code, but it is committed and it is broken.

### ✅ Acceptance Criteria

- [ ] `python -m compileall backend/app` exits clean
- [ ] `uvicorn app.main:app` starts
- [ ] `upload_voice_introduction` and `upload_video_introduction` are separate handlers, each validating and saving through its own helper
- [ ] `save_video_introduction_upload` is restored; both save helpers write to their own directory and return their own URL
- [ ] `UserBase.voice_introduction_url` is correctly indented and present on the schema
- [ ] `check_pr.py` is re-encoded as UTF-8 (or removed)
- [ ] A test asserts that **every** module under `backend/app` imports, so the next bad merge fails CI instead of shipping

MOHITKOURAV01:
Fixes #1194

## What was wrong

`main` does not compile. Three modules have syntax errors, and one of them is registered on the v1 router, so the application cannot import and every endpoint is down:

```
$ python -m compileall -q backend/app
backend/app/routers/users.py:525    SyntaxError: invalid syntax
backend/app/schemas/user.py:96      IndentationError: unexpected indent
backend/app/utils/uploads.py:117    SyntaxError: '(' was never closed
```

All three come from `e2be7a58`, which merged `feat/video-introduction-978` into `main` on top of the already-merged `feat/voice-introduction-977`. The two branches edited the same three functions, and the merge kept **both sides of every hunk, interleaved**, without leaving a conflict marker. The result reads like ordinary code until the parser gets to it.

While fixing that I ran a parse over the whole backend, which turned up four more places with the same signature — a name bound twice, the second binding silently discarding the first. Two of them are live bugs.

### The three syntax errors

**`app/routers/users.py`** — `upload_voice_introduction` had lost its entire body (it validated the filename and then fell straight into the next `@router.post`), and `upload_video_introduction` had two function bodies spliced together with three unclosed call expressions. Split back into two handlers, each calling its own validator and save helper. The router was also importing only the voice helpers, so the video ones are now imported too.

**`app/utils/uploads.py`** — `save_video_introduction_upload` had been swallowed by the following `def`. Restored it. `save_voice_introduction_upload` had two `upload_dir` assignments and two `return`s stacked; it was writing to `voice_introductions/` and returning a `video_introductions/` URL, so every saved clip 404'd.

**`app/schemas/user.py`** — `voice_introduction_url` was indented one level too deep, in two places.

### The four duplicate definitions

**`app/routers/auth.py`** — `/refresh` and `/logout` were each declared twice, ~350 lines apart. Both decorators run at import, so both routes were registered and Starlette served whichever it matched first — the earlier one. The later pair was dead except in the OpenAPI schema, where they appeared as duplicate operations.

I kept the live handlers and folded in the two things the dead copies did better: the explicit `decode_token(..., expected_type=REFRESH)` pre-check on `/refresh`, which rejects a non-refresh token before touching the database, and `AUTH_LIMIT` in place of a hardcoded `"10/minute"`. The dead copies are removed. Net behaviour change: `/refresh` now 401s on a wrong-typed token slightly earlier. Everything else is identical.

**`app/services/recommendation_service.py`** — two different services were both named `RecommendationService` in the same module. The second replaced the first, taking `get_recommended_projects` with it — and `app/routers/recommendations.py:69` still calls that method, so `GET /api/v1/recommendations/projects` raised `AttributeError` and answered 500.

They are genuinely different services with different scoring models, so rather than merge them the first is renamed to `ProjectAffinityRecommendations` and `RecommendationService` now inherits from it. Both call sites resolve; no method body changed.

**`app/services/auth_service.py`** — `github_login` defined twice on the same class. Two different implementations of the OAuth flow, only the second of which ever ran. Removed the dead first one (84 lines).

**`tests/test_projects.py`, `tests/test_users.py`** — `test_get_project_not_found` and `test_get_user_not_found` each declared twice, identical apart from a local variable name. Each file reported one more test than it ran. Removed the shadowed copies.

### Supporting

- **`backend/alembic/versions/video_intro_url_001_...`** — `video_introduction_url` and `video_introduction_thumbnail_url` have been on the ORM model since the video feature landed but never got a migration; only the voice column did. On a database built from migration history rather than `create_all`, selecting a user raises `UndefinedColumn`. Chained off `voice_intro_url_001` so it does not add another head to a history that already has more than one (#926). Guarded with an inspector check so it is safe against databases that already have the columns.

- **`UserService.update_video_introduction_url`** — re-indented from its 7-space block, and it no longer clears `video_introduction_thumbnail_url` when the caller does not pass one. Re-uploading a video through the endpoint (which generates no thumbnail) was wiping a thumbnail that had been set previously.

- **`save_resume_upload`** hardcoded `.pdf`, so every DOCX resume was stored under a name claiming to be a PDF and would not open for whoever downloaded it. It now follows the validated extension.

- **Video size limit** — video introductions were sharing `MAX_UPLOAD_SIZE_MB`, which is the *image* limit. A real 30-second clip was rejected as too large. Videos get their own budget.

- **`check_pr.py`** — removed. It is a two-line debugging one-liner hardcoded to PR #914, committed in UTF-16 (so `python check_pr.py` dies with "source code string cannot contain null bytes"), with escaping mangled badly enough that it cannot be recovered into something meaningful. Nothing references it. Happy to restore it as UTF-8 instead if anyone wants it.

- Stripped a UTF-8 BOM from `add_voice_introduction_url.py`.

## Tests

**`tests/test_source_integrity.py`** (new) — this is the part I actually care about, because none of the above should have been able to reach `main`.

It walks every `.py` file under `backend/` and asserts, without importing anything:

- it parses (`ast.parse`) — the check `main` failed
- it contains no merge conflict markers — the ordinary version of the same accident
- it is UTF-8, no BOM, no null bytes — how `check_pr.py` broke
- no name is bound twice at module scope, and no class defines a method twice — the quiet version, where the file parses fine and half of it is dead

No imports means no database, no settings, no third-party packages: it runs in a bare checkout in about a second. That matters, because the tests that *would* have caught the original breakage all import `app.main` transitively, so they failed at **collection** — and a collection error reads like a broken environment rather than broken code.

The duplicate-definition check carries a small allowlist, `KNOWN_DUPLICATE_DEFINITIONS`, holding the one case not fixed here: `PROJECT_ROLE_PERMISSIONS` in `app/core/rbac.py`, which is #1197 and has its own PR. A companion test asserts every allowlist entry is *still* a duplicate, so an allowance cannot outlive the bug it covers.

**`tests/test_media_introduction_uploads.py`** (new) — 30 tests over the voice and video paths. The one that matters most is `test_voice_and_video_do_not_share_a_directory`: the merge damage was only possible because nothing asserted that a voice upload lands somewhere different from a video upload, so a function doing half of each still looked correct.

## Verification

```
$ python -m compileall -q backend/app        # clean
$ python -c "import app.main"                # imports
$ pytest tests/test_source_integrity.py tests/test_media_introduction_uploads.py
```

Full suite run locally before and after.

## Review notes

- The auth change is the only one with any behavioural delta, and it is small. Worth a careful look at whether folding the type pre-check into the live `/refresh` is what we want, or whether it should stay purely a deletion.
- `ProjectAffinityRecommendations` is not a great name. Open to better ones — the constraint was keeping `RecommendationService.get_recommended_projects` resolvable without touching the router.
- I removed `check_pr.py` rather than repairing it. Say the word and I will restore it as UTF-8 instead.
- There are other stray artifacts at the repo root (`google_patch.diff`, `ms_patch.diff`, `pr_info.json`, `upstream_sections.tsx`) that look accidentally committed. Left alone here — out of scope, but probably worth a separate cleanup.


MOHITKOURAV01:
Pushed a follow-up: the `Revision graph` and `Upgrade and downgrade` jobs were failing, so I chased it down.

They were not caused by this PR — `main` already has **five** Alembic heads, and `alembic upgrade head` cannot resolve a target while more than one exists. The migration I added here replaced `voice_intro_url_001` as a head rather than adding a sixth, so the count was unchanged at 5. What this PR did was *trigger* the jobs, since they are path-filtered to alembic changes and nothing had touched a migration in a while.

This is #926 regressed. It was closed after collapsing eleven heads, and it came back the same way it happened the first time: feature branches each chained off whatever was current when they branched, so merging them produced siblings rather than a chain.

`zzzz00000001_merge_all_heads.py` is a merge revision over the five. I used a merge rather than rewriting anyone's `down_revision` because the five are genuinely independent — different tables, any order — and rewriting history that has already been applied would strand every environment that has run one of them. It creates and drops nothing.

Graph is now 1 head, 1 base, 57 revisions, no duplicate ids.

Happy to pull both the new migration and this merge revision out into their own PR if you would rather keep this one to the compile fix — say the word and I will split it.

PR #1238, opened 2026-08-20
fix(api): make the OpenAPI document generatable again

### LINKED ISSUE #1236: /openapi.json, /docs and /redoc all return 500: Optional is used in routers/projects.py but never imported
`/openapi.json`, `/docs` and `/redoc` all return 500: `Optional` is used in `routers/projects.py` but never imported

## Summary

`backend/app/routers/projects.py` starts with `from __future__ import annotations` and then uses `Optional[str]` and `Optional[uuid.UUID]` in `get_project_audit_trail` — but `Optional` is never imported in that module.

Because of the postponed-evaluation import the annotations are plain strings at import time, so the app starts cleanly and every existing test passes. The name is only resolved when FastAPI builds the OpenAPI schema, and at that point it blows up:

```
pydantic.errors.PydanticUserError: `TypeAdapter[typing.Annotated[ForwardRef('Optional[str]'), FieldInfo(..., alias='event_type',
description='Filter by event type substring', ...)]]` is not fully defined; you should define
`typing.Annotated[ForwardRef('Optional[str]'), ...]` and all referenced types, then call `.rebuild()` on the instance.
```

## Reproduction

```python
from fastapi.testclient import TestClient
from app.main import app

TestClient(app).get("/openapi.json")   # 500
```

Or against a running server:

```bash
curl -o /dev/null -w '%{http_code}\n' http://localhost:8000/openapi.json   # 500
```

`/docs` and `/redoc` serve their HTML shell fine and then fail as soon as the browser fetches `/openapi.json`, so both are blank/errored in practice.

## Where

`backend/app/routers/projects.py:451`

```python
@router.get(
    "/{project_id}/audit-trail",
    response_model=PaginatedProjectAuditLogsResponse,
    ...
)
def get_project_audit_trail(
    project_id: uuid.UUID,
    event_type: Optional[str] = Query(              # <- Optional not imported
        None, description="Filter by event type substring"
    ),
    user_id: Optional[uuid.UUID] = Query(           # <- same
        None, description="Filter by actor or target user ID"
    ),
    ...
```

Those are the only two uses of `Optional` in the file, and `typing` is not imported anywhere in it. Line 1 is `from __future__ import annotations`, which is exactly what turns a `NameError` at import time into a 500 at schema time.

## Impact

- The interactive API docs — the main reference for everyone building against this backend — are down.
- `GET /openapi.json` is a 500, so anything that consumes the spec is down with it. That includes the `sdk-generation` workflow and any client codegen.
- `GET /api/projects/{project_id}/audit-trail` cannot resolve its own signature, so the endpoint itself is unusable.

Nothing in CI covers this: no test requests `/openapi.json`, and `app.openapi()` is never called during the test run.

## Expected

- `GET /openapi.json` returns 200 and a valid schema.
- `/docs` and `/redoc` render.
- The project audit-trail endpoint works.

## Suggested fix

Import `Optional` in `routers/projects.py` (or drop it for `str | None`, which is what the rest of the newer routers use), and add a test that calls `app.openapi()` / requests `/openapi.json` and asserts 200 — a schema that cannot be generated should be a red test, not a surprise in the browser.

MOHITKOURAV01:
Fixes #1236.

## The bug

Two route signatures use `Optional` in modules that never import it:

- `app/routers/projects.py` — `get_project_audit_trail(event_type: Optional[str], user_id: Optional[uuid.UUID])`
- `app/routers/analytics.py` — `current_user: Annotated[Optional[User], Depends(get_optional_current_user)]`

Both files start with `from __future__ import annotations`, which is why this was never a `NameError`. The annotations stay strings at import time; the name is only resolved when FastAPI walks the routes to build the schema, and that is where it fails:

```
pydantic.errors.PydanticUserError: `TypeAdapter[typing.Annotated[ForwardRef('Optional[str]'), ...]]`
is not fully defined; you should define ... and all referenced types, then call `.rebuild()`.
```

The visible symptoms are `GET /openapi.json` → 500, and `/docs` / `/redoc` blank because that is what they fetch.

The issue only mentioned `projects.py`. The second one in `analytics.py` fell out of writing the test — worth noting since it means this was not a one-off typo.

## The fix

Both switched to `X | None`, which is the spelling the surrounding code already uses (`projects.py` has `language: str | None = Query(None)` a couple of hundred lines up), rather than adding a `typing` import for two parameters.

```diff
-    event_type: Optional[str] = Query(
+    event_type: str | None = Query(
         None, description="Filter by event type substring"
     ),
-    user_id: Optional[uuid.UUID] = Query(
+    user_id: uuid.UUID | None = Query(
         None, description="Filter by actor or target user ID"
     ),
```

## Why nothing caught it

`app.openapi()` was never called anywhere in the test suite, and no test requested `/openapi.json`. The schema is not exercised by endpoint tests at all — a route can be completely undescribable and still serve traffic.

`tests/test_openapi_schema.py` closes that gap:

| Test | What it catches |
|---|---|
| `test_openapi_schema_can_be_generated` | the failure in this issue |
| `test_openapi_endpoint_returns_200` | generation succeeding but the response failing to serialise |
| `test_docs_pages_are_served` | `/docs` and `/redoc` returning non-200 for some other reason |
| `test_schema_documents_the_whole_api` | a router silently not included — the document generates fine and is just missing a chunk |
| `test_operation_ids_are_unique` | colliding `operationId`s, which break generated SDKs; this app mounts some routers under two prefixes deliberately |
| `test_no_route_signature_uses_an_unimported_name` | the same class of bug, but reported as "module.function: name 'Optional' is not defined" instead of a wall of Pydantic internals |

Verified they actually fail on the broken code — reverting just the two annotation changes turns the file red:

```
2 failed, 1 passed, 3 errors
FAILED test_openapi_endpoint_returns_200
FAILED test_no_route_signature_uses_an_unimported_name
```

## One thing worth flagging for review

Walking the routes is not as simple as it used to be. Recent FastAPI does not flatten `include_router` eagerly — `app.routes` holds `_IncludedRouter` placeholders that keep the real router on `original_router` and only resolve it while matching a request. A naive walk over `.routes` finds **5** routes for this app; following `original_router` as well finds **547**. If that internal ever changes, `test_schema_documents_the_whole_api` (which asserts > 500 documented paths, independently of the walker) still holds the line.

## Verification

```
$ pytest tests/test_openapi_schema.py -q
6 passed
```

`/openapi.json` now returns 200 with 722 documented paths.

### DIFF
--- a/backend/app/routers/analytics.py
+++ b/backend/app/routers/analytics.py
@@ -178,7 +178,7 @@ def get_my_profile_analytics(
 def track_profile_click(
     request: TrackClickRequest,
     db: Annotated[Session, Depends(get_db)],
-    current_user: Annotated[Optional[User], Depends(get_optional_current_user)] = None,
+    current_user: Annotated[User | None, Depends(get_optional_current_user)] = None,
 ):
     AnalyticsService.log_profile_click(
         db=db,
--- a/backend/app/routers/projects.py
+++ b/backend/app/routers/projects.py
@@ -456,10 +456,10 @@ def update_project(
 )
 def get_project_audit_trail(
     project_id: uuid.UUID,
-    event_type: Optional[str] = Query(
+    event_type: str | None = Query(
         None, description="Filter by event type substring"
     ),
-    user_id: Optional[uuid.UUID] = Query(
+    user_id: uuid.UUID | None = Query(
         None, description="Filter by actor or target user ID"
     ),
     page: int = Query(1, ge=1),
--- a/backend/tests/test_openapi_schema.py
+++ b/backend/tests/test_openapi_schema.py
@@ -1,0 +1,188 @@
+"""The OpenAPI document has to be buildable.
+
+`app.openapi()` was never called anywhere in the suite, so a route whose
+signature could not be resolved was invisible: the app imported, every
+endpoint test passed, and `/openapi.json` returned a 500 in the browser.
+
+That is exactly the failure mode `from __future__ import annotations` invites.
+Annotations become strings at import time, so an unimported name in a route
+signature is not a `NameError` when the module loads -- it is a
+`PydanticUserError` much later, when FastAPI finally tries to resolve the
+forward reference to build a schema for it.
+
+These tests build the document once and then assert things about it, so any
+route that cannot describe itself fails here rather than in a client
+generator or a docs page.
+"""
+
+from __future__ import annotations
+
+from collections import Counter
+
+import pytest
+from fastapi.routing import APIRoute
+from fastapi.testclient import TestClient
+
+from app.main import app
+
+
+@pytest.fixture(scope="module")
+def schema() -> dict:
+    """The generated document.
+
+    Built once per module: generation walks every route and is the expensive
+    part of these tests. `app.openapi_schema` is cleared first so a document
+    cached by an earlier test cannot mask a regression here.
+    """
+    app.openapi_schema = None
+    return app.openapi()
+
+
+def test_openapi_schema_can_be_generated(schema: dict) -> None:
+    """The headline check: building the document does not raise.
+
+    A route signature that references a name the module never imported only
+    fails at this point, which is why this assertion is worth its own test
+    even though every other test in the file depends on the same fixture.
+    """
+    assert schema["openapi"].startswith("3.")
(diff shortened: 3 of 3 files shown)


nensii21:

PR #1239, opened 2026-08-20
fix(api): one spelling for the error code, and finish wiring the envelope

### LINKED ISSUE #1237: Error envelope was half-renamed: responses carry error.error_code, everything else still expects error.code
Error envelope was half-renamed: responses carry `error.error_code`, everything else still expects `error.code`

## Summary

The standardized error envelope introduced in #572 emits its machine-readable code under the key `error_code`, while the *input* side of the very same handler, the handler's own docstring example, and all 17 tests written for it use `code`. The rename landed on the output half only.

`backend/app/core/error_handlers.py:67`

```python
error_dict: Dict[str, Any] = {
    "error_code": code,          # <- emitted
    "message": message,
    "timestamp": ...,
    "request_id": ...,
}
```

`backend/app/core/error_handlers.py:88` — the same module, reading a route's structured detail:

```python
if isinstance(exc.detail, dict):
    code = exc.detail.get("code") or DEFAULT_STATUS_CODES.get(status_code, "ERROR")
```

So a route raises `HTTPException(detail={"code": "PROJECT_NOT_FOUND", ...})` and the client receives `{"error": {"error_code": "PROJECT_NOT_FOUND", ...}}`. Write `code`, read `error_code`.

`git log -p` on the file shows it in one hunk of `95ead353` ("feat: Standardize API error responses (#572)"):

```
-        "code": code,
+        "error_code": code,
```

with the docstring example above it left on `"code": "PROJECT_NOT_FOUND"`, and no corresponding change in the tests.

## Reproduction

Any error response:

```bash
$ curl -s http://localhost:8000/api/messages/conversation/00000000-0000-0000-0000-000000000000/nope
{"error":{"error_code":"NOT_FOUND","message":"Not Found","timestamp":"...","request_id":"..."},"detail":"Not Found"}
```

## Failing tests on `main`

17 tests across three files fail with `KeyError: 'code'`:

```
$ pytest tests/test_error_responses.py tests/test_global_error_handler.py tests/test_request_validation.py -q
...
    def test_http_exception_404_format(client):
        response = client.get("/test/not-found")
        assert response.status_code == 404
        data = response.json()
        assert "error" in data
>       assert data["error"]["code"] == "PROJECT_NOT_FOUND"
E       KeyError: 'code'

7 failed  tests/test_error_responses.py
8 failed  tests/test_global_error_handler.py
2 failed  tests/test_request_validation.py
```

These are not stale tests chasing a feature that never shipped — they are the tests that came with #572, testing the contract #572 documented.

## Why it matters beyond the red CI

`error.code` is the field a client is supposed to branch on; `message` is for humans and `detail` is the FastAPI compatibility shim. Right now:

- the tests that define the contract are red, so the envelope is effectively unverified;
- a route that raises a structured `detail={"code": ...}` and a route that raises a plain string detail are indistinguishable to a caller trying to read `code`;
- #1067 ("Standardize API Response Schemas") is still open and will have to pick one of these two spellings anyway.

## Expected

One key, consistently. `code` is the one the rest of the codebase, the docstring, and the tests already use:

```json
{
  "error": {
    "code": "PROJECT_NOT_FOUND",
    "message": "Project not found.",
    "timestamp": "2026-08-20T15:32:23.934285+00:00",
    "request_id": "eb13279e-d0b9-48e9-94ef-ea1612da2d11"
  },
  "detail": "Project not found."
}
```

## Suggested fix

Emit `code` from `format_error_response`, and keep `error_code` alongside it as a deprecated alias for one release so any consumer that started reading the current spelling does not break. Cover every handler (`http_exception_handler`, `validation_exception_handler`, `rate_limit_exception_handler`, `integrity_error_handler`, `global_exception_handler`) with a test that asserts both keys are present and equal, and write the envelope down somewhere in `docs/` so the next rename has something to contradict.

MOHITKOURAV01:
Fixes #1237.

## The reported bug

`format_error_response` emitted `error_code`; the docstring above it, the input side of the same handler, and every test used `code`.

```python
# error_handlers.py — output
"error_code": code,

# error_handlers.py, 20 lines down — input
code = exc.detail.get("code") or DEFAULT_STATUS_CODES.get(status_code, "ERROR")
```

`code` wins — it is what the rest of the codebase already uses. `error_code` stays as an alias carrying the identical value, because responses have shipped with it long enough that something outside this repo may read it, and swapping one silent break for another is not a fix. Both are named constants (`ERROR_CODE_KEY`, `DEPRECATED_ERROR_CODE_KEY`) so the next rename has to trip over the fact that there are two.

## What else turned up

Writing the tests made it clear the key was the visible half of a larger gap.

**`AppException` had no handler.** `app/core/exceptions.py` defines `AppException`, `AuthException`, `DatabaseException` and `ValidationException`, each carrying its own `status_code`, `code` and `details` — and none of it reached the client. With no registration they fell through to `global_exception_handler` and came back as a flat `500 INTERNAL_SERVER_ERROR`. A `401 TOKEN_EXPIRED` was indistinguishable from "the server is broken". That is almost certainly why nothing in the codebase raises them: they didn't work. Now registered, so the hierarchy is usable.

**Generic `SQLAlchemyError` had no handler.** Only `IntegrityError` did. Everything else — connection failures, operational errors — went out through the catch-all, and the driver's message can contain the statement, its bound parameters, and the host it failed to reach. Now a `500 DATABASE_ERROR` with the real text logged.

**`integrity_error_handler` leaked the row back to the caller.** `details` was the raw driver string:

```
duplicate key value violates unique constraint "ix_users_email"
DETAIL:  Key (email)=(victim@example.com) already exists.
```

That names the physical index and, worse, **echoes the value that collided** — so a signup form hitting a duplicate email replied with the address already on the account. It now returns `{"duplicate_field": "email"}` and logs the rest. This is the one change here with a security flavour; flagging it explicitly rather than burying it in the diff.

Separately, the handlers that were building a payload without setting `detail` now set it, so the FastAPI compatibility mirror is on every path out instead of most.

## Registration order

`IntegrityError` is a subclass of `SQLAlchemyError`, and `AppException` is a subclass of `Exception`, so both new registrations sit "under" existing ones. Starlette resolves handlers by walking the raised exception's MRO and picking the most specific class registered, so the specific ones still win. The block is written most-specific-first anyway and says so in a comment — order is documentation here, not precedence.

## Tests

`backend/tests/test_error_envelope.py` (31 tests) covers the seam that actually broke, rather than more per-handler code assertions:

- `code` present on **every** registered handler's output — parametrised, because the break was one handler's worth of inconsistency and a spot check is what missed it;
- the deprecated alias *equals* `code` rather than merely existing — two keys that can disagree would be worse than one wrong key;
- `message` / `timestamp` / `request_id` always present, and `detail` mirrors `message`;
- a route's `code` survives the round trip unchanged;
- none of the three leak paths above put server-side detail in the body (asserted against the response text, not just the parsed field).

## The one test edit

`test_unhandled_value_error_handling` could never have passed. `add_exception_handler(Exception, ...)` is installed on Starlette's `ServerErrorMiddleware`, which builds the response and then re-raises so the server still logs the fault — so the response is unreachable through a default `TestClient`. Its sibling in `test_error_responses.py` already used `raise_server_exceptions=False` for exactly this; this now does the same, with a comment saying why.

## Docs

`docs/api_error_envelope.md` — the field table, where `code` comes from, the deprecation, what never appears in an error body and why, and the handler table. There was no written spec for this before, which is part of how a rename could half-land.

## Verification

Full backend suite, before and after, comparing the failure sets:

```
before: 96 failing    after: 79 failing
regressions: (none)
fixed:       17
```

The 17 are all of `test_error_responses.py`, `test_global_error_handler.py` and `test_request_validation.py`. The remaining 79 are pre-existing failures in unrelated areas (websockets, RBAC test fixtures, stale org paths) and are untouched by this PR.

### DIFF
--- a/backend/app/core/error_handlers.py
+++ b/backend/app/core/error_handlers.py
@@ -2,7 +2,7 @@ from __future__ import annotations
 
 import logging
 import re
-from typing import Any, Dict
+from typing import Any, Dict, Final
 from datetime import datetime, timezone
 from app.core.tracing import get_request_id
 
@@ -11,7 +11,9 @@ from fastapi.exceptions import HTTPException, RequestValidationError
 from fastapi.responses import JSONResponse
 from slowapi.errors import RateLimitExceeded
 from starlette.exceptions import HTTPException as StarletteHTTPException
-from sqlalchemy.exc import IntegrityError
+from sqlalchemy.exc import IntegrityError, SQLAlchemyError
+
+from app.core.exceptions import AppException
 
 logger = logging.getLogger("devlink")
 
@@ -51,6 +53,27 @@ def generate_error_code(status_code: int, message: str | None) -> str:
     return "_".join(words).upper()
 
 
+#: The key clients are meant to branch on.
+#:
+#: This used to be spelled two ways at once. `format_error_response` emitted
+#: `error_code`, while the docstring example, the *input* side of
+#: `http_exception_handler` (`exc.detail.get("code")`) and every test written
+#: for the envelope used `code`. A route would raise
+#: `HTTPException(detail={"code": "PROJECT_NOT_FOUND"})` and the caller would
+#: receive `{"error": {"error_code": "PROJECT_NOT_FOUND"}}`: write `code`,
+#: read `error_code`.
+ERROR_CODE_KEY: Final[str] = "code"
+
+#: The spelling that shipped by accident, kept as an alias.
+#:
+#: Nothing in this repository reads it, but the responses have been going out
+#: with it for a while and something outside may have started depending on it.
+#: Removing it in the same change that fixes the contract would trade one
+#: silent break for another, so it stays -- carrying the same value -- until a
+#: release that says it is going.
+DEPRECATED_ERROR_CODE_KEY: Final[str] = "error_code"
+
+
 def format_error_response(
     code: str,
     message: str,
@@ -59,12 +82,31 @@ def format_error_response(
 ) -> Dict[str, Any]:
     """
     Build standardized JSON response payload.
+
+    The envelope is::
+
+        {
+          "error": {
+            "code": "PROJECT_NOT_FOUND",
+            "error_code": "PROJECT_NOT_FOUND",   # deprecated alias
+            "message": "Project not found.",
+            "timestamp": "2026-08-20T15:32:23.934285+00:00",
+            "request_id": "eb13279e-d0b9-48e9-94ef-ea1612da2d11",
+            "details": ...                       # only when there are any
+          },
+          "detail": "Project not found."         # FastAPI compatibility
+        }
+
+    `code` is machine-readable and stable; `message` is for humans and may be
+    reworded; `detail` exists so callers written against plain FastAPI keep
+    working.
     """
     request_id = getattr(request.state, "request_id", None) if request else None
     request_id = request_id or get_request_id() or "unknown"
 
     error_dict: Dict[str, Any] = {
-        "error_code": code,
+        ERROR_CODE_KEY: code,
+        DEPRECATED_ERROR_CODE_KEY: code,
         "message": message,
         "timestamp": datetime.now(timezone.utc).isoformat(),
         "request_id": request_id,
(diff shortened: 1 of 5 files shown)


nensii21:

PR #1240, opened 2026-08-20
fix(messages): require conversation membership on every message route

### LINKED ISSUE #1234: Private conversations are readable without a token: the messages API never checks conversation membership
Private conversations are readable without a token: the messages API never checks conversation membership

## Summary

`backend/app/routers/messages.py` declares `router = APIRouter(tags=["Messages"])` with **no router-level dependency**, and then several endpoints take neither `current_user` nor any membership check. The result:

- three message endpoints are reachable **with no credentials at all** and return message bodies;
- every other message endpoint authenticates the caller but never asks whether that caller is a member of the conversation they are addressing.

Knowing (or enumerating) a conversation UUID is enough to read a private DM thread, search it, send into it, and pin messages in it.

## Reproduction

Alice creates a conversation and posts a message. Eve is a registered user who is not a member. Run against the app with the standard test fixtures:

```
CREATE CONV     201
SEND (alice)    201   "salary is 250k"

EVE LIST        200
NO-TOKEN LIST   200   [{"content":"salary is 250k", ...}]
NO-TOKEN COUNT  200
NO-TOKEN SEARCH 200   [{"content":"salary is 250k", ...}]
EVE PINNED      200
EVE PIN         200
EVE SEND        201   "eve was here"
```

`NO-TOKEN LIST` / `NO-TOKEN COUNT` / `NO-TOKEN SEARCH` are plain `curl`s with no `Authorization` header:

```
curl http://localhost:8000/api/messages/conversation/<uuid>
curl http://localhost:8000/api/messages/conversation/<uuid>/count
curl "http://localhost:8000/api/messages/search/<uuid>?keyword=salary"
```

## Affected endpoints

Unauthenticated — no `get_current_user` dependency at all:

| Endpoint | Line |
|---|---|
| `GET /messages/search/{conversation_id}` | `messages.py:170` |
| `GET /messages/conversation/{conversation_id}/count` | `messages.py:188` |
| `GET /messages/conversation/{conversation_id}` | `messages.py:222` |

Authenticated but not membership-checked:

| Endpoint | Line | Effect |
|---|---|---|
| `POST /messages/` | `messages.py:46` | send into any conversation |
| `GET /messages/conversation/{id}/pinned` | `messages.py:207` | read any conversation's pinned messages |
| `PATCH /messages/{id}/pin` | `messages.py:387` | pin anyone's message |
| `PATCH /messages/{id}/unpin` | `messages.py:413` | unpin anyone's message |
| `GET /messages/{message_id}` | `messages.py:319` | read any single message |
| `POST/DELETE /messages/typing/...` | `messages.py:246`, `293` | inject and observe typing state |
| `POST /messages/conversation/{id}/read` and the bulk read/deliver routes | `messages.py:493`, `526` | mutate read state in a foreign thread |

The service layer does not compensate: `MessageService.list_pinned_messages(db, conversation_id)` and `MessageService.list_conversation_messages(db, conversation_id, limit)` filter only on `conversation_id`. Note that `MessageService.search_all_messages` (the newer global search added with #973) *does* scope to `ConversationMember` — so the pattern is understood, it was simply never applied to the per-conversation routes.

The `/pin`, `/unpin` and `/pinned` routes are new in #1205 (`feat/advanced-messaging-973`), but the unauthenticated list/search/count routes predate it — this has been open for a while and got wider, not newer.

## Expected

- Every `/messages` route requires an authenticated caller.
- Every route that names a `conversation_id` (directly, via `MessageCreate.conversation_id`, or via the message it addresses) rejects callers who are not a `ConversationMember` of that conversation — `404` rather than `403`, so conversation ids are not confirmed to strangers.
- Pin/unpin require membership; edit/delete/restore continue to require authorship.

## Suggested fix

A single `require_conversation_member` dependency in the messages router, plus a `_member_message_or_404` helper for the routes that start from a `message_id`, applied uniformly — and regression tests that assert a non-member and an anonymous caller both get `404` on each route.

MOHITKOURAV01:
Fixes #1234.

## What was wrong

Two distinct problems, same root cause — `router = APIRouter(tags=["Messages"])` with no dependency, and per-route checks that were never added.

**Three routes had no authentication at all.** `GET /messages/conversation/{id}`, `GET /messages/conversation/{id}/count` and `GET /messages/search/{id}` took no `current_user`, so there was nothing to check. A plain `curl` with a conversation id returned message bodies.

**The rest never checked membership.** Authenticated, but a conversation id was the only thing between a stranger and a private thread: list it, send into it, read and set its pinned messages, watch and forge its typing indicators, read any single message.

## The check already existed

Worth saying plainly, because it changes what this PR is: the read-receipt routes (`mark_as_read`, `mark_as_delivered`, `mark_conversation_as_read`, the bulk variants) have been checking `ConversationMember` in the service layer since they were written, and return `403` when it fails. Nothing new was invented here — this applies an existing convention to the routes that never adopted it.

## Three shapes, because the id arrives three ways

**Path parameter → `require_conversation_member`**

```python
def list_conversation_messages(
    conversation_id: uuid.UUID,
    current_user: User = Depends(require_conversation_member),
): ...
```

The dependency authenticates, checks membership, and *returns the caller*. A route that wants `current_user` gets it from the check, so there is no way to take the user and forget the check. That coupling is deliberate — the original gap is exactly what "remember to also add the check" looks like across 20 routes.

**Request body → `_assert_conversation_member`**

`POST /messages/`, `/read/bulk` and `/bulk-deliver` carry the conversation id in the body where a path-parameter dependency cannot see it. Explicit call, before the write rather than after.

**Message id → `_member_message_or_404`**

`GET /messages/{id}`, `/pin`, `/unpin`, and the edit/delete/restore family reach the conversation only through the message. Resolve the message (404 if absent), then check membership of its conversation.

## Membership is not authorship

Edit/delete/restore keep the ownership check and gain the membership one. Not redundant in either direction:

- a member who did not write a message still cannot edit it;
- `sender_id` survives someone leaving a conversation, so an ownership-only check let a **removed member go on editing and deleting inside a thread they had left**. There is a test for this specifically.

## Two decisions worth reviewing

**403, not the 404 the issue suggested.** I changed my mind writing it. 404 avoids confirming a conversation exists, but the read-receipt routes already answer 403 and their tests assert it — 404 here would mean two different answers to the same question depending on which route you asked. Conversation ids are UUIDs, so enumeration is not the threat model, and consistency is worth more. Happy to flip both if you'd rather.

**Bulk-by-message-id is scoped, not refused.** `/read/bulk` and `/bulk-deliver` accept either a `conversation_id` or a list of `message_ids`. A caller may legitimately pass ids spanning several of their own conversations, so the id form stays *filtered* to conversations they belong to (already the behaviour in `MessageService.bulk_mark_as_read`) rather than rejecting the whole request. A stranger's id updates nothing — asserted in `test_bulk_read_by_message_id_ignores_foreign_messages`.

## Enforcement point

The router, not the service. `MessageService.list_conversation_messages` and friends still filter only on `conversation_id`. I left them alone deliberately — adding a redundant membership query inside each would double the round trips on the hot read path, and the dependency makes the router check structurally hard to skip. Flagging it since "the service does not compensate" was in the issue and this PR consciously does not change that.

## Tests

`backend/tests/test_message_authorization.py` — 18 tests, every route three ways:

| | |
|---|---|
| anonymous | the routes that used to answer 200 with no header |
| authenticated non-member | parametrised across all 8 conversation-scoped routes, plus send / read-one / pin / unpin / read-state / edit / delete |
| **member** | every route still works, listing still returns content, sending still works, pinning still works, the author can still edit and delete |

The member cases matter as much as the other two — an authorization change that locks out the people who are allowed in is not a fix.

Assertions check the response *text* for the secret string too, not just the status code, so a handler that 500s after serialising the messages would still fail.

## Verification

Reverting only the router change turns 9 of the 18 red:

```
FAILED test_unauthenticated_read_routes_are_closed[GET-/api/messages/conversation/{cid}]
FAILED test_unauthenticated_read_routes_are_closed[GET-/api/messages/conversation/{cid}/count]
FAILED test_unauthenticated_read_routes_are_closed[GET-/api/messages/search/{cid}?keyword=offer]
FAILED test_non_member_is_refused_every_conversation_route
FAILED test_non_member_cannot_send_into_the_conversation
FAILED test_non_member_cannot_read_a_single_message
FAILED test_non_member_cannot_pin_or_unpin
FAILED test_removed_member_loses_access_to_their_own_messages
FAILED test_unknown_conversation_is_refused_not_leaked
9 failed, 9 passed
```

Full backend suite: 96 failures before and after — identical sets, no regressions. (Those 96 are pre-existing failures in unrelated areas on `main`.)

## Scope

`/api/conversations` has its own routes for creating conversations and adding/removing members; they are not touched here. `docs/messaging_authorization.md` writes the rule down.

### DIFF
--- a/backend/app/routers/messages.py
+++ b/backend/app/routers/messages.py
@@ -25,7 +25,7 @@ from app.schemas.message import (
 from app.services.message_service import MessageService
 
 # pyrefly: ignore [missing-import]
-from sqlalchemy import select
+from sqlalchemy import func, select
 
 from app.models.conversation_member import ConversationMember
 from app.models.message import Message
@@ -35,6 +35,84 @@ from app.services.notification_service import NotificationService
 router = APIRouter(
     tags=["Messages"],
 )
+
+
+# ------------------------------------------------------------------
+# Conversation membership  (issue #1234)
+# ------------------------------------------------------------------
+#
+# Every route below addresses a conversation: by path parameter, through the
+# body of the request, or indirectly through the message it names. None of
+# them may act on a conversation the caller does not belong to.
+#
+# The read-receipt routes already got this right at the service layer
+# (`MessageService.mark_as_read` and friends check `ConversationMember` before
+# touching anything), so the shape below is deliberately the same check and
+# the same 403 -- this is applying an existing convention to the routes that
+# never adopted it, not inventing a second one.
+
+
+def _assert_conversation_member(
+    db: Session,
+    conversation_id: uuid.UUID,
+    user_id: uuid.UUID,
+) -> None:
+    """Raise 403 unless ``user_id`` belongs to ``conversation_id``.
+
+    A count rather than a fetch: nothing here needs the membership row, only
+    the yes/no, and the composite index on (conversation_id, user_id) answers
+    it without loading anything.
+    """
+    is_member = db.scalar(
+        select(func.count(ConversationMember.id)).where(
+            ConversationMember.conversation_id == conversation_id,
+            ConversationMember.user_id == user_id,
+        )
+    )
+
+    if not is_member:
+        raise HTTPException(
+            status_code=status.HTTP_403_FORBIDDEN,
+            detail="You are not a member of this conversation",
+        )
+
+
+def require_conversation_member(
+    conversation_id: uuid.UUID,
+    db: Session = Depends(get_database),
+    current_user: User = Depends(get_current_user),
+) -> User:
+    """Dependency for routes with a ``conversation_id`` path parameter.
+
+    Returns the caller, so a route declares
+    ``current_user: User = Depends(require_conversation_member)`` and gets
+    both the authentication and the membership check from one dependency --
+    there is no way to take the user and forget the check.
+    """
+    _assert_conversation_member(db, conversation_id, current_user.id)
+    return current_user
+
+
+def _member_message_or_404(
+    db: Session,
+    message_id: uuid.UUID,
+    current_user: User,
+) -> Message:
+    """Fetch a message the caller is entitled to see.
+
+    For routes keyed on a ``message_id``, where the conversation is only
+    reachable through the message itself.
+    """
+    db_message = MessageService.get_message(db, message_id)
+
+    if db_message is None:
+        raise HTTPException(
+            status_code=status.HTTP_404_NOT_FOUND,
+            detail="Message not found",
+        )
+
+    _assert_conversation_member(db, db_message.conversation_id, current_user.id)
+    return db_message
 
 
 @router.post(
@@ -49,6 +127,11 @@ def send_message(
     db: Session = Depends(get_database),
     current_user: User = Depends(get_current_user),
 ):
+
+    # The conversation id is in the body, out of reach of
+    # `require_conversation_member`, so the check is explicit here -- and
+    # before the write, not after it.
+    _assert_conversation_member(db, message.conversation_id, current_user.id)
 
     sent = MessageService.send_message(
         db=db,
@@ -172,7 +255,9 @@ def search_messages(
     conversation_id: uuid.UUID,
     keyword: str = Query(...),
     db: Session = Depends(get_database),
+    current_user: User = Depends(require_conversation_member),
 ):
+    """Search one conversation the caller belongs to."""
 
     return MessageService.search_messages(
         db,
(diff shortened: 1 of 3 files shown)


nensii21:

PR #1222, opened 2026-08-18
feat/added empty state ui

anshika-guleria:
# Pull Request

## Summary

Improves empty states across profile and key user-facing sections with consistent illustrated UI, clearer guidance, and contextual calls to action.

---

## Related Issue

Closes #726

---

## Type of Change

Please check the relevant option(s):

- [ ] Bug fix
- [ ] New feature
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Refactoring
- [x] UI/UX enhancement
- [ ] Tests
- [ ] Other

---

## Changes Made

- Replaced plain-text empty messages with consistent icon-based empty-state panels.
- Added helpful descriptions and contextual CTAs for actions such as creating projects, hackathons, issues, and templates.
- Improved empty states across profile sections, activity feed, applications, dashboard projects, flares, leaderboard, templates, hackathons, and project issues.
- Added offline-safe empty states for pinned projects and contribution activity.
- Corrected profile project filtering so profiles only display their own projects.

---

## Screenshots (if applicable)

Screenshot (1956)
Screenshot (1957)
Screenshot (1958)
Screenshot (1959)
Screenshot (1960)

---

## Testing

Describe how you tested your changes.

- [x] Tested locally
- [ ] Existing tests pass
- [ ] Added new tests
- [x] UI verified
- [ ] Cross-browser tested (if applicable)

Additional testing notes:

- Verified the frontend in mock/offline mode without the backend running.
- `npm run typecheck` is currently blocked by pre-existing errors in `src/routes/verify-email.tsx` related to a missing `authApi.verifyEmail` method.

---

## Checklist

Before submitting this pull request, confirm the following:

- [x] My code follows the project's coding standards.
- [x] I have tested my changes locally.
- [ ] I have updated the documentation where necessary.
- [x] My changes do not introduce new warnings or errors.
- [x] I have reviewed my own code.
- [x] This pull request focuses on a single feature or fix.
- [x] I have linked the related issue.

---

## Additional Notes
none

PR #1188, opened 2026-08-16
feat(chat): implement compact messaging widget on dashboard

Babin123456:
## Description
Implements a space-efficient Compact Messaging Widget on the dashboard sidebar for quick chat previews and messaging access.

### Key Changes
- **Compact Avatar Layout**: Streamlined 32px builder avatars with live green status indicators.
- **Animated Typing Indicator**: Integrated animated three-dot `TypingIndicator` for conversations with active typing.
- **Unread Badge & Timestamps**: Displays unread count pills and relative time elapsed (`5m`, `25m`, `2h`).
- **Direct Navigation**: Clicking any conversation navigates directly to `/messages/$conversationId`.
- **Dashboard Sidebar Integration**: Added `CompactMessagingWidget` to `frontend/src/routes/_app.dashboard.tsx` sidebar alongside notifications and upcoming events.

## Related Issue
Closes #741

## Checklist
- [x] Seamless layout integration on dashboard sidebar.
- [x] Type-safe implementation with TanStack Query integration.
- [x] Committed and pushed to `feat/compact-messaging-widget-741`.

## Screenshot

image

PR #1187, opened 2026-08-16
Feat/ai recommendation panel

Babin123456:
## Description
Enhances the dashboard's AI Recommendation Panel with actionable collaborator insights, match scores, and direct interaction buttons.

### Key Changes
- **Match Score Badges**: Displays high-contrast match percentage badges (e.g., `94% Match`) calculated from recommendation scores.
- **Skills Gap Analysis**: Displays matched skills (`✓ Skill`) and missing skills needed for team completion (`+ Skill needed`).
- **Recommended Collaborator Profiles**: Shows builder avatars, full names, roles, and headline bios.
- **Suggested Actions**: Added direct action buttons ("Invite to Team", "Connect", "View Profile") linking to builder profiles and directory.
- **Empty & Loading States**: Clean skeleton loaders and empty state guidance when profile data needs completion.

## Related Issue
Closes #738

## Checklist
- [x] Connects to `recommendationsApi.builders()` with fallback mock data.
- [x] No `any` type usage, conforming to repository ESLint rules.
- [x] Committed and pushed to `feat/ai-recommendation-panel-738`.

PR #1186, opened 2026-08-16
Feat/project overview cards

Babin123456:
## Description
Redesigns the project overview cards across the dashboard and projects explorer to present richer, more actionable project information at a glance.

### Key Changes
- **Reusable `ProjectOverviewCard`**: Created `frontend/src/components/projects/ProjectOverviewCard.tsx` featuring:
  - Visual completion percentage bar with dynamic color milestones.
  - Team size metrics (current vs max capacity) with builder count.
  - Status badges (`Recruiting`, `In Progress`, `Completed`, `Archived`).
  - Deadline / due date badges.
  - Tech stack tags with overflow indicators and difficulty level pill.
  - Quick action link to full project details.
- **Dashboard Integration**: Updated `CurrentProjects` in `frontend/src/features/dashboard/sections.tsx` with completion metrics, team count, and dynamic API/fallback integration.
- **Projects Grid Integration**: Integrated `ProjectOverviewCard` into `frontend/src/routes/_app.projects.tsx` grid view alongside project insight cards.

## Related Issue
Closes #737

## Checklist
- [x] Responsive layout across mobile, tablet, and desktop viewports.
- [x] Type-safe TypeScript implementation with zero ESLint warnings.
- [x] Committed and pushed to `feat/project-overview-cards-737`.

PR #1241, opened 2026-08-20
fix(dashboard): repair the mis-merged project card and make the frontend build

### LINKED ISSUE #1233: `main` does not build: features/dashboard/sections.tsx has a half-merged project card block
`main` does not build: `features/dashboard/sections.tsx` has a half-merged project card block

## What happens

`main` (currently `ce32637c`) does not compile. Both `npm run typecheck` and `npm run build` fail in the frontend:

```
$ npx tsc --noEmit
src/features/dashboard/sections.tsx(135,8): error TS17008: JSX element 'div' has no corresponding closing tag.
src/features/dashboard/sections.tsx(219,13): error TS1128: Declaration or statement expected.
src/features/dashboard/sections.tsx(221,11): error TS1381: Unexpected token. Did you mean `{'}'}` or `&rbrace;`?
src/features/dashboard/sections.tsx(284,13): error TS1381: Unexpected token. Did you mean `{'}'}` or `&rbrace;`?
src/features/dashboard/sections.tsx(284,15): error TS1381: Unexpected token. Did you mean `{'}'}` or `&rbrace;`?
src/features/dashboard/sections.tsx(285,9): error TS17002: Expected corresponding JSX closing tag for 'Link'.
src/features/dashboard/sections.tsx(286,5): error TS1005: ')' expected.
```

```
$ npx vite build
✗ Build failed in 1.61s
error during build:
[builtin:vite-transform] Adjacent JSX elements must be wrapped in an enclosing tag.
  ╭─[ src/features/dashboard/sections.tsx:219:13 ]
```

This is not a lint nit — the file is not parseable, so nothing downstream of it builds. Every open PR's `typecheck` / `frontend-ci` job fails on it regardless of what that PR changed, and the app cannot be deployed from `main`.

## Where it came from

`sections.tsx` was edited by four feature branches in a row:

- #1186 `feat/project-overview-cards-737` (rewrote the project card)
- #1187 `feat/ai-recommendation-panel-738`
- #1188 `feat/compact-messaging-widget-741`
- #1222 `feat/add-empty-states-ui`

Two of those branches each carry a commit that already tried to repair this file (`a275a308 fix(dashboard): fix unclosed JSX tags and duplicate imports in sections`, `5a43aa9b fix(dashboard): fix JSX structure and remove duplicate useQuery in sections`), and then the "merge main into branch" commits (`6428e8aa`, `d7709cf9`, `68a4c62f`) reintroduced the broken state. What landed is **both** versions of the project card spliced together.

Concretely, in `CurrentProjectsSection`:

- the `<div className="flex-1 px-5 pb-5 pt-1 ...">` opened at line 135 is never closed;
- the old avatar-stack card body ends at line 219–221 with `))}`, closing the `.map()`;
- and then at line 222 a **second** `return (` for the new overview card starts, outside of any callback, followed by a `})}` at line 284 that closes a callback that is no longer open;
- line 200 has a bare `const statusBadge = statusMap[p.status] || statusMap["in-progress"];` sitting in the middle of JSX children.

## Expected

- `npm run typecheck` and `npm run build` both succeed on `main`.
- `CurrentProjectsSection` renders one coherent card — the overview card from #1186 (icon, description, completion bar, team size, deadline, status badge) — with the empty/loading/error states from #1222 preserved.

## Suggested fix

Rebuild `CurrentProjectsSection` from the intended final shape rather than trying to patch the splice: one project shape, one card renderer, loading / error / empty branches kept, and the dead duplicate block removed. It would also be worth making the frontend typecheck job a required check so a non-parsing file cannot reach `main` again.

MOHITKOURAV01:
Fixes #1233.

`main` does not build. `npm run typecheck` and `npm run build` both fail on `features/dashboard/sections.tsx`, so every open PR's frontend job is red regardless of what that PR changed. The `frontend` workflow is failing on `main` right now.

## What the file actually contained

`CurrentProjects` had **two versions of the same project list spliced into one return**:

- the old avatar-stack `<Link>` rows from before #1186, and
- the overview card that replaced them,

with the wrapping `<div>` never closed, a second `return (` sitting outside any callback, a `})}` closing a callback that was no longer open, and a bare `const statusBadge = ...` statement in the middle of JSX children.

Four branches edited this file in sequence — #1186, #1187, #1188, #1222. Two of them carry a commit that already tried to repair it (`a275a308`, `5a43aa9b`) and the later "merge main into branch" commits put the broken state back.

## The approach

Re-splicing the two versions would leave the thing that made them mergeable in the first place. `ProjectOverviewCard` already exists, already has a `compact` variant, and is already used by the projects index — the dashboard was carrying an **inline near-copy of it**. `CurrentProjects` now renders the shared component. The loading skeletons, error branch and empty state from #1222 are all kept.

Net: −236 / +116 in that file.

## Three things the repair surfaced

**`error` was never destructured.** The JSX read `error` in its failure branch, but `useQuery` was destructured as `{ data, isLoading }` only. The error branch referenced an undefined binding. Fixed.

**The component-level `fallbackProjects` array is removed.** This is the one behavioural decision here, so flagging it plainly. `projectsService.list` is already `withFallback(() => projectsApi.list(params), seed.projects)` — the demo data is handled at the service layer. A second fallback inside the component could therefore only fire when the API genuinely returned an empty list, which is exactly the case #1222 added the `EmptyState` for. Showing someone three invented projects instead of a "create project" prompt is the wrong answer, and having both is why the merge had two contradictory intentions to reconcile. If you'd rather keep the mock projects, say so and I'll put them back behind an explicit flag instead.

**Dead `projectsList` mock removed** — left over from before the API call was wired up, unreferenced.

## A second failure was hiding behind the first

`tsc` reports nothing else while a file in the program fails to parse. Fixing the parse error surfaced:

```
src/routes/verify-email.tsx(9,38): error TS2345: Argument of type '"/verify-email"'
is not assignable to parameter of type 'keyof FileRoutesByPath | undefined'.
```

`routeTree.gen.ts` is generated *and committed*, and it was missing the `/verify-email` route. Regenerated — the diff is only that entry (+21 lines, three interface tables and the import).

## Why this reached `main`

`.github/workflows/typecheck.yml`, in full, before this PR:

```yaml
name: Type Check
on: [pull_request]
jobs:
  typecheck:
    runs-on: ubuntu-latest
    steps:
      - uses: actions/checkout@v4
```

It checks out the repository and stops. No Node, no install, no `tsc`. It has been reporting success on every pull request while checking nothing — which is worse than not having the job, because `main` sat with a file that does not parse and a green "Type Check" beside it.

It now installs and runs `npm run typecheck`, and `frontend-ci` gains typecheck and test steps ahead of its build (cheapest first — a file that does not parse should fail in seconds, not after a full production build).

## Verification

```
$ npx tsc --noEmit          # clean
$ npx vitest run            # 47 files, 701 tests passed
$ npx vite build            # ✓ built
$ npx eslint src/features/dashboard/sections.tsx   # clean
```


nensii21:

Issue #1248, opened 2026-08-21
CI, Lint, Type Check and Security Scan check out the repository and stop — main has 11 type errors behind four green ticks


MOHITKOURAV01:
## Summary

Four workflows run on every pull request, report success, and check nothing. Here is `typecheck.yml` in full:

```yaml
name: Type Check

on: [pull_request]

jobs:
  typecheck:
    runs-on: ubuntu-latest

    steps:
      - uses: actions/checkout@v4
```

That is the whole file. No `setup-node`, no install, no `tsc`. It clones the repository and the job ends green.

`ci.yml`, `lint.yml` and `security.yml` are the same file with a different `name:`. `release-drafter.yml` is zero bytes.

## This is not theoretical

`main` does not type check. Run `npx tsc --noEmit` in `frontend/`:

```
src/routes/index.tsx(221,15): error TS2322: Type '"/showcase"' is not assignable to type ...
src/routes/index.tsx(309,17): error TS2322: Type '"/showcase"' is not assignable to type ...
src/routes/showcase.tsx(37,38): error TS2345: Argument of type '"/showcase"' is not assignable ...
src/routes/showcase.tsx(55,23): error TS2339: Property 'page' does not exist on type ...
src/routes/showcase.tsx(56,20): error TS2339: Property 'q' does not exist on type ...
src/routes/showcase.tsx(57,27): error TS2339: Property 'language' does not exist on type ...
src/routes/showcase.tsx(58,29): error TS2339: Property 'difficulty' does not exist on type ...
src/routes/showcase.tsx(59,28): error TS2339: Property 'tab' does not exist on type ...
src/routes/showcase.tsx(161,15): error TS2322: Type '"/showcase"' is not assignable to type ...
src/routes/verify-email.tsx(9,38): error TS2345: Argument of type '"/verify-email"' is not assignable ...
```

Eleven errors, behind a green "Type Check ✓" on every PR that introduced them.

The cause is that `src/routeTree.gen.ts` is generated but committed, and it is stale. It registers 50 routes; `src/routes/` holds 52 files calling `createFileRoute`. `showcase.tsx` (merged in #1228) and `verify-email.tsx` are both missing from it, so every `Link to="/showcase"` and every `Route.useSearch()` in those files is typed against a tree that does not contain them.

[...]

PR #1250, opened 2026-08-21
fix(export): call the builder-flare collector instead of an undefined name

### LINKED ISSUE #1245: Every data-export route is a 500: collect_user_data references an undefined `builder_flares`
## Summary

`ExportService.collect_user_data` ends with

```python
        return UserExportData(
            ...
            notifications=notifications,
            builder_flares=builder_flares,
        )
```

`builder_flares` is not a local, not a parameter, and not an import. It is a bare undefined name, so the call raises `NameError` before it can build the response. Every route that reaches this function is a 500.

The data it was meant to pass is already there: `ExportService._get_builder_flares(db, user_id)` is defined 400 lines further down (`export_service.py:469`) and is never called from anywhere.

## Reproduction

Register a user, then hit any export route:

```
POST /api/users/me/export                        -> NameError: name 'builder_flares' is not defined
GET  /api/users/me/portfolio/export?format=json      -> NameError: name 'builder_flares' is not defined
GET  /api/users/me/portfolio/export?format=markdown  -> NameError: name 'builder_flares' is not defined
GET  /api/users/me/portfolio/export?format=pdf       -> NameError: name 'builder_flares' is not defined
```

All four go through `collect_user_data`, directly or via `export_portfolio_markdown` / `export_portfolio_html`.

## What this takes down

| Caller | Effect |
|---|---|
| `routers/export.py:25` — `POST /me/export` | the "download everything we hold about you" route. Nothing comes back. |
| `routers/export.py:40,48,55` — `GET /me/portfolio/export` | all three formats: markdown, html/pdf, json |
| `services/backup_service.py:110` | account backups are built on `collect_user_data`, so backup creation dies with it |

The export route is the GDPR/DSAR path. It is the one route where a 500 is not just a broken feature.

## Why nothing caught it

`from __future__ import annotations` is at the top of the module, but that is not what hid this — annotations were never the issue, the name is used in an *expression*. What hid it is that there is no test that calls `collect_user_data`, and no test that requests any `/export` route. `grep -rn "collect_user_data\|portfolio/export" backend/tests/` returns nothing.

`ruff check --select F821` reports it in under a second:

```
app/services/export_service.py:65:28: F821 Undefined name `builder_flares`
```

Nothing runs that. See the CI issue filed alongside this one — `lint.yml` checks out the repository and stops.

## Expected

`collect_user_data` calls `_get_builder_flares` like it calls the eleven collectors above it, and the export routes return a document.

## Suggested fix

One line to call the collector that already exists, plus tests that actually request each export route and assert on the body — including a builder flare in the fixture so `builder_flares` is non-empty and a future regression cannot pass by returning `[]`.

The same sweep is worth doing for the other `F821`s in `app/services/` (`analytics_service.py`, `duplicate_detection_service.py`, `notification_service.py`, `project_service.py`, `user_service.py`), though those are all in annotations and are latent rather than live.

MOHITKOURAV01:
Fixes #1245.

## The bug

```python
return UserExportData(
    ...
    notifications=notifications,
    builder_flares=builder_flares,   # <- not a local, not a param, not an import
)
```

`NameError` before the response is built, so every route that reaches
`collect_user_data` is a 500. That is `POST /me/export`, all three formats of
`GET /me/portfolio/export`, and `BackupService.create_backup`, which builds on
the same function.

The data was already collected — `_get_builder_flares` is defined at
`export_service.py:469` and had no callers. One line:

```diff
 notifications = ExportService._get_notifications(db, user.id)
+builder_flares = ExportService._get_builder_flares(db, user.id)
 return UserExportData(
```

## The other five

`ruff check --select F821 app/` reported six undefined names, not one. The issue
called the other five latent, on the grounds that they are all in annotations
under `from __future__ import annotations`. That was right for four of them and
wrong for the fifth:

```python
# routers/websockets.py, inside the doc.update handler
doc_uuid = UUID(doc_id) if doc_id else None
```

`UUID` is never imported in that module. It is a call, not an annotation, so
every collaborative-document edit over the websocket raises `NameError`. I
found it while adding the guard below rather than by reading the handler, which
is the argument for the guard.

The four genuinely latent ones (`Any` in `notification_service` /
`project_service` / `duplicate_detection_service`, `Optional` + `uuid` +
`ProfileAnalyticsResponse` in `analytics_service`, and the two schema names in
`user_service`) are imported rather than annotated away. They break
`get_type_hints()` on those classes today and would break louder the moment one
of those return types is used as a response model.

`routers/analytics.py` and `routers/projects.py` still have an `Optional` each.
Those are #1236 and **#1238 already fixes them** — I have left them alone rather
than create a conflict, and the guard allowlists exactly those two by file *and*
message until that merges.

## Why nothing caught it

No test requested an export route. `grep -rn "portfolio/export\|collect_user_data" backend/tests/`
was empty before this change. The route was not weakly covered; it was
uncovered, and so was `BackupService`'s use of it.

## Tests

**`tests/test_export_routes.py`** — 15 tests, the first coverage these routes
have had.

The two things it does deliberately:

- **requests the routes** rather than calling the service. A service-level test
  would have caught the `NameError`, but not that `export_portfolio_html`
  reaches it by a different path than `export_portfolio_markdown`.
- **asserts on the body**. A fix that passed `builder_flares=[]` satisfies every
  status-code assertion perfectly, so the fixture creates a real flare and the
  test asserts its title, role and status come back.

Also covers the two directions that a one-line fix makes easy to get wrong:
`test_export_does_not_leak_another_users_flares` (the `created_by` filter, which
nothing exercised) and `test_export_of_a_brand_new_user_is_empty_but_valid`
(empty is `[]`, not an error and not a missing key).

**`tests/test_no_undefined_names.py`** — runs `ruff check --select F821 app/`
and fails on anything not on the two-entry allowlist.

Scoped to one rule on purpose. This is not "lint the backend" — `ruff check app/`
reports 3354 findings, almost all formatting, and turning that on is a decision
that needs a baseline. F821 is the subset that means *this line raises when
reached*, so it needs no baseline beyond the two sites with a fix already open.

## The guard is in the wrong place, and I put it there anyway

A linter belongs in `lint.yml`. `lint.yml` checks out the repository and stops
(#1248), and the `ruff` pre-commit hook only sees staged files in a tree where
someone ran `pre-commit install`. So it lives in pytest, where something
actually runs it, with a comment saying it should move when the lint workflow
becomes real.

Worth flagging that this is currently advisory rather than blocking:
`backend-ci.yml` runs `pytest` under `continue-on-error: true`, so a red test
does not fail the job. That is the same defect as #1248 and out of scope here,
but it means the guard warns rather than gates until that changes.

## Verification

Reverting only the one-line fix turns 13 of the 15 export tests red, and 2 of
the 3 guard tests:

```
FAILED test_export_routes_return_a_document[data-export]
FAILED test_export_routes_return_a_document[portfolio-json]
FAILED test_export_routes_return_a_document[portfolio-markdown]
FAILED test_export_routes_return_a_document[portfolio-html]
FAILED test_data_export_carries_every_section
FAILED test_builder_flares_are_populated_not_just_present
FAILED test_export_does_not_leak_another_users_flares
FAILED test_portfolio_json_is_the_same_document_as_the_data_export
FAILED test_portfolio_markdown_renders_the_users_content
FAILED test_portfolio_html_renders_the_users_content
FAILED test_collect_user_data_is_callable
FAILED test_collect_user_data_is_json_serialisable
FAILED test_export_of_a_brand_new_user_is_empty_but_valid
13 failed, 2 passed

E  app/services/export_service.py:65:28: Undefined name `builder_flares`
2 failed, 1 passed
```

The two that stay green are the auth check and the `?format=docx` rejection —
both return before reaching the service, which is the right reason to survive.

Full backend suite, run on this branch and on `main` in a clean worktree:
**98 failures, identical sets, no regressions.** Those 98 are pre-existing and
unrelated (`test_last_active`, `test_notifications_events` and friends, mostly
the `TestingSessionLocal` import errors ruff also flags).

## Scope

Not touched: `backup_service.py`, which is fixed by consequence rather than by
change — it calls `collect_user_data` and now gets a document. I have not added
backup tests here; that path deserves its own coverage and its own PR.

### DIFF
--- a/backend/app/routers/websockets.py
+++ b/backend/app/routers/websockets.py
@@ -35,6 +35,7 @@ import json
 import logging
 from datetime import datetime, timezone
 from typing import Any, Dict, List, Optional, Set
+from uuid import UUID
 
 from fastapi import APIRouter, Depends, WebSocket, WebSocketDisconnect, status
 from jose import JWTError, jwt
--- a/backend/app/services/analytics_service.py
+++ b/backend/app/services/analytics_service.py
@@ -1,7 +1,8 @@
 from __future__ import annotations
 
+import uuid
 from datetime import datetime, timedelta, timezone
-from typing import Dict, List
+from typing import Dict, List, Optional
 
 from sqlalchemy import func, select
 from sqlalchemy.orm import Session
@@ -18,6 +19,7 @@ from app.schemas.analytics import (
     ProjectGrowthMetric,
     RetentionMetric,
 )
+from app.schemas.profile_analytics import ProfileAnalyticsResponse
 
 
 class AnalyticsService:
--- a/backend/app/services/duplicate_detection_service.py
+++ b/backend/app/services/duplicate_detection_service.py
@@ -3,7 +3,7 @@ from __future__ import annotations
 import json
 import logging
 import re
-from typing import Optional
+from typing import Any, Optional
 from uuid import UUID
 
 from sqlalchemy import select
--- a/backend/app/services/export_service.py
+++ b/backend/app/services/export_service.py
@@ -49,6 +49,7 @@ class ExportService:
         organizations = ExportService._get_organizations(db, user.id)
         activities = ExportService._get_activities(db, user.id)
         notifications = ExportService._get_notifications(db, user.id)
+        builder_flares = ExportService._get_builder_flares(db, user.id)
         return UserExportData(
             exported_at=now,
             profile=profile,
--- a/backend/app/services/notification_service.py
+++ b/backend/app/services/notification_service.py
@@ -1,6 +1,8 @@
 from __future__ import annotations
 
 import uuid
+from typing import Any
+
 from app.utils.time import utcnow
 
 # pyrefly: ignore [missing-import]
--- a/backend/app/services/project_service.py
+++ b/backend/app/services/project_service.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import uuid
+from typing import Any
 
 from sqlalchemy import func, select
 from sqlalchemy.orm import Session
--- a/backend/app/services/user_service.py
+++ b/backend/app/services/user_service.py
@@ -7,7 +7,12 @@ from sqlalchemy.orm import Session
 
 from app.models.activity import ActivityType
 from app.models.user import User
-from app.schemas.user import UserCreate, UserUpdate
+from app.schemas.user import (
+    PrivacySettingsUpdate,
+    ProfileCompletionResponse,
+    UserCreate,
+    UserUpdate,
+)
 from app.services.activity_service import ActivityService
 from app.models.application import Application, ApplicationStatus
 from app.models.follower import Follower
--- a/backend/tests/test_export_routes.py
+++ b/backend/tests/test_export_routes.py
@@ -1,0 +1,308 @@
+"""
+Coverage for the account-export routes.
+
+There was none before. `ExportService.collect_user_data` referenced an
+undefined `builder_flares` and raised `NameError` on every call, and all four
+export routes went through it, so all four were 500s -- including
+`POST /me/export`, which is the "give me everything you hold about me" route.
+The suite stayed green because nothing here ever asked for an export.
+
+So these tests do two things that are easy to skip:
+
+* they **request the routes**, rather than calling the service directly. The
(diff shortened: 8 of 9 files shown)


nensii21:

PR #1251, opened 2026-08-21
fix(routing): stop mounting routers under a prefix they already declare

### LINKED ISSUE #1246: Eleven routers are mounted at a doubled prefix, so /api/bookmarks 404s and /api/bookmarks/bookmarks is the real path
## Summary

A router that declares its own `prefix=` and is then mounted with a `prefix=` that ends in the same segment gets both. `app/main.py` does this thirteen times, so the legacy `/api` surface serves

```
/api/bookmarks/bookmarks/...
/api/conversations/conversations/...
/api/organizations/organizations/...
/api/skills/skills/...
/api/ws/ws/...
```

and the paths everyone writes down — `/api/bookmarks`, `/api/conversations` — are 404.

## Reproduction

Authenticated, against the standard test fixtures:

```
GET /api/bookmarks                       -> 404   |  /api/bookmarks/bookmarks                       -> 200
GET /api/bookmark-collections            -> 404   |  /api/bookmark-collections/bookmark-collections -> 200
GET /api/conversations                   -> 404   |  /api/conversations/conversations               -> 200
GET /api/organizations                   -> 404   |  /api/organizations/organizations               -> 200
GET /api/skills                          -> 404   |  /api/skills/skills                            -> 200
```

## Full list

Legacy `/api` mounts in `app/main.py`, router's own prefix in the middle column:

| Router | own prefix | served at |
|---|---|---|
| `bookmarks` | `/bookmarks` | `/api/bookmarks/bookmarks` |
| `bookmark_collections` | `/bookmark-collections` | `/api/bookmark-collections/bookmark-collections` |
| `conversations` (mounted twice) | `/conversations` | `/api/conversations/conversations` |
| `profile_suggestions` (mounted twice) | `/profile-suggestions` | `/api/profile-suggestions/profile-suggestions` and `/api/users/me/profile-suggestions/profile-suggestions` |
| `repositories` | `/repositories` | `/api/repositories/repositories` |
| `organizations` | `/organizations` | `/api/organizations/organizations` |
| `skills` | `/skills` | `/api/skills/skills` |
| `websockets` | `/ws` | `/api/ws/ws` |
| `recommendations` | `/recommendations` | `/api/recommendations/recommendations` |
| `health` | `/health` | `/api/health/health` |
| `saved_searches` | `/api/saved-searches` | `/api/saved-searches/api/saved-searches` |

`saved_searches` is its own thing: the router hardcodes `/api/` into its *own* prefix (`app/routers/saved_searches.py`), which is wrong on both surfaces regardless of how it is mounted.

`websockets` matters more than the count suggests — the socket endpoint is at `/api/ws/ws`, and a client that cannot guess that gets no realtime at all on the legacy surface.

## The v1 surface is mostly, but not entirely, fine

`docs/environment-setup.md` tells you to point the frontend at `http://localhost:8000/api/v1`, and `app/api/v1/router.py` mounts most routers bare, letting each carry its own prefix. That is the correct pattern and it is why this is survivable.

Three leaks through anyway:

- `organizations` is included **twice** — once with `prefix="/organizations"` (doubled) and once bare (correct). Both are live.
- `profile_suggestions` is included twice, both doubled.
- `saved_searches` gives `/api/v1/api/saved-searches`, from the hardcoded prefix above.

## Why nothing caught it

Nothing asserts a route's *path*. The suite has 4822 tests and not one of them requests any of these eleven routers on the legacy surface — grepping `backend/tests/` for them turns up four string literals in total, none of which is an affected path. A router can be mounted at any URL at all and the suite stays green.

The frontend does not report it either. `services/index.ts` wraps calls in `withFallback`, which answers a 404 with canned data, so a dead endpoint renders as plausible content instead of an error. That is filed separately.

## Expected

Each router is reachable at exactly one path per surface, and that path contains no repeated segment. `/api/bookmarks` and `/api/v1/bookmarks` both work; `/api/bookmarks/bookmarks` does not exist.

## Suggested fix

Drop the redundant `prefix=` from the legacy mounts and let each router's own prefix stand, matching what `api/v1/router.py` already does. Remove the hardcoded `/api/` from `saved_searches`. Delete the duplicate `organizations` and `profile_suggestions` includes.

The part worth keeping is a test that walks `app.routes` and fails on any path with a repeated segment, or with `/api/` inside a `/api/v1/` path — this is the kind of mistake that comes back the next time someone adds a router, and a list of known-good paths in a test does not prevent it.

Renaming a live URL is a break, so the legacy paths probably need to keep answering for a release. Worth deciding in review whether the doubled paths get a deprecation shim or just go.

MOHITKOURAV01:
Fixes #1246.

## The bug

Two prefixes, both applied:

```python
# app/routers/bookmarks.py
router = APIRouter(prefix="/bookmarks")

# app/main.py
app.include_router(bookmarks.router, prefix="/api/bookmarks")
#                                    -> /api/bookmarks/bookmarks/...
```

Neither line is wrong on its own, which is why this survived eleven times. You
have to read them together, and they are in different files.

```
GET /api/bookmarks         -> 404   |  /api/bookmarks/bookmarks         -> 200
GET /api/conversations     -> 404   |  /api/conversations/conversations -> 200
GET /api/organizations     -> 404   |  /api/organizations/organizations -> 200
GET /api/skills            -> 404   |  /api/skills/skills               -> 200
```

`docs/api_versioning_and_migration.md` says unversioned `/api` routes map to
the current stable version. For these eleven they did not map to anything you
could find.

## The fix

Mounts under `/api` carry only `/api`, and each router's own prefix supplies
the segment — which is what `api/v1/router.py` already does for most of its
includes. The router file is the one that should name the resource, because it
is the file that gets mounted on two surfaces.

```diff
-app.include_router(bookmarks.router, prefix="/api/bookmarks", tags=["Bookmarks"])
+app.include_router(bookmarks.router, prefix="/api", tags=["Bookmarks"])
```

Eleven of those. Two things needed more than that:

**`saved_searches` had the mount in its own prefix.**

```diff
 router = APIRouter(
-    prefix="/api/saved-searches",
+    prefix="/saved-searches",
```

`/api` in a router's prefix is wrong on both surfaces — it gave
`/api/saved-searches/api/saved-searches` on the legacy mount and
`/api/v1/api/saved-searches` under v1. This is the one place a route moved
under `/api/v1` as well.

**Three routers were included twice.** `conversations` (`main.py`, six lines
apart, byte-identical), `project_milestones` (adjacent, byte-identical) and
`background_jobs` (`main.py`, fifteen lines apart, with the import repeated
too). Under v1, `organizations` and `profile_suggestions` were each included
twice — once doubled, once not. The dead registrations are removed.

`organizations` needed a moment's care: the doubled v1 include was the one
carrying `tags=["Organizations"]`, so the tag moves onto the bare include
that survives rather than being dropped with it.

## `/api/ws/ws/presence`

Worth pulling out. The websocket router was mounted at `/api/ws` with its own
`/ws`, so the socket endpoints were at `/api/ws/ws/presence` and
`/api/ws/ws/collab`. A websocket that cannot be connected to does not 404
visibly; the client just never gets a connection. There is a test for this one
specifically.

## Tests

`backend/tests/test_route_paths.py` — 29 tests. The first thing in this repo
that asserts anything about a route's *path*.

The core of it is mechanical rather than a list of known-good URLs:

- no path has an immediately repeated segment
- no `/api/v1/` path contains a nested `/api/`
- no router is mounted twice at the same prefix

A list would work today and would have to be updated by whoever next makes
this mistake, which is the one person guaranteed not to think of it.

On top of that, the eleven from the issue are named individually — both
directions, `test_router_is_served_at_the_undoubled_path` and
`test_the_doubled_path_is_gone` — so a failure says which router regressed
rather than "some path repeats a segment". And `test_the_restored_paths_actually_answer`
requests them over HTTP, because a route being registered is not the same as a
route that works.

## One test guards the others

`test_the_walker_finds_a_plausible_number_of_routes`.

FastAPI does not flatten `include_router` into `app.routes` — it leaves an
`_IncludedRouter` holding the mount prefix, and the child routes already carry
their own router's prefix in `path`. My first walker did not know that and
returned **7 paths**. Every assertion in the file passed.

So the walker asserts it found at least 500 routes before anything else runs.
Without it this file is one FastAPI upgrade away from being a green tick that
checks nothing, which is the failure mode I am trying to remove from this
repository, not add to it.

## Decisions worth reviewing

**The doubled paths are gone, not deprecated.** They 404 now (or 422, where
`/api/bookmarks/bookmarks/` falls through to `/api/bookmarks/{bookmark_id}` and
"bookmarks" is not a UUID — the test asserts "not 200" rather than a specific
code for that reason).

I went back and forth. The argument for a deprecation shim is that removing a
live URL is a break. The argument against, which won: these URLs are not
documented anywhere, no test requests them, and the frontend does not call them
— it calls `/bookmarks/` and `/recommendations/...` relative to
`VITE_API_BASE_URL`, which `docs/environment-setup.md` sets to `/api/v1`. So
the population of callers is plausibly empty, and keeping both live means two
URLs per resource, two sets of cache keys, two rate-limit buckets and two
things to grep for. Happy to add aliases if you would rather not assume that.

**`project_tags` keeps its doubled path.** `/project-tags/predefined` is a
second `@router.get` with `include_in_schema=False` — a deliberate hidden alias
for callers that hardcoded the old form. It is in `DELIBERATE_ALIASES` by exact
path rather than covered by a blanket "skip hidden routes" rule, which would
also hide the next accident.

## Verification

Against unfixed `app/`, 26 of the 29 tests fail:

```
FAILED test_no_path_repeats_a_segment
FAILED test_no_v1_path_contains_a_second_api_segment
FAILED test_router_is_served_at_the_undoubled_path[bookmarks]
... 8 more routers ...
FAILED test_the_doubled_path_is_gone[bookmarks]
... 8 more routers ...
FAILED test_websockets_are_reachable_without_the_doubled_segment
FAILED test_no_router_is_mounted_twice_at_the_same_prefix
FAILED test_the_restored_paths_actually_answer
FAILED test_health_is_reachable_at_the_documented_path
26 failed, 3 passed
```

The three that stay green are the walker sanity check, the alias check and one
parametrised case — all correctly independent of the fix.

Full backend suite, this branch and `main` in a clean worktree: **98 failures,
identical sets, no regressions.** Nothing was relying on a doubled path, which
is the one thing I most wanted to confirm before deleting them.

## Docs

`docs/routing_prefixes.md` writes the rule down, including the two shapes that
go wrong and the legitimate case for mounting one router twice
(`profile_suggestions` at `/api` and `/api/users/me`, which is a real pair of
paths and stays).

### DIFF
--- a/backend/app/api/v1/router.py
+++ b/backend/app/api/v1/router.py
@@ -93,9 +93,6 @@ api_v1_router.include_router(
     builder_flares.router, prefix="/flare", tags=["Builder's Flare"]
 )
 api_v1_router.include_router(messages.router, prefix="/messages", tags=["Messages"])
-api_v1_router.include_router(
-    organizations.router, prefix="/organizations", tags=["Organizations"]
-)
 api_v1_router.include_router(org_audit_logs.router)
 api_v1_router.include_router(webhooks.router)
 api_v1_router.include_router(
@@ -112,13 +109,11 @@ api_v1_router.include_router(
     profile_summary.router, prefix="/profile-summary", tags=["Profile Summary"]
 )
 api_v1_router.include_router(
-    profile_suggestions.router,
-    prefix="/profile-suggestions",
-    tags=["Profile Suggestions"],
+    profile_suggestions.router, tags=["Profile Suggestions"]
 )
 api_v1_router.include_router(
     profile_suggestions.router,
-    prefix="/users/me/profile-suggestions",
+    prefix="/users/me",
     tags=["Profile Suggestions"],
 )
 api_v1_router.include_router(
@@ -136,7 +131,7 @@ api_v1_router.include_router(
 )
 api_v1_router.include_router(repositories.router)
 api_v1_router.include_router(project_releases.router)
-api_v1_router.include_router(organizations.router)
+api_v1_router.include_router(organizations.router, tags=["Organizations"])
 api_v1_router.include_router(applications.router)
 api_v1_router.include_router(skills.router)
 api_v1_router.include_router(testimonials.router)
--- a/backend/app/main.py
+++ b/backend/app/main.py
@@ -530,9 +530,6 @@ from app.routers import project_milestones
 app.include_router(
     project_milestones.router, prefix="/api", tags=["Project Milestones"]
 )
-app.include_router(
-    project_milestones.router, prefix="/api", tags=["Project Milestones"]
-)
 from app.routers import project_time_logs
 
 app.include_router(
@@ -558,25 +555,20 @@ app.include_router(
 )
 
 app.include_router(followers.router, prefix="/api/followers", tags=["Followers"])
-app.include_router(bookmarks.router, prefix="/api/bookmarks", tags=["Bookmarks"])
+app.include_router(bookmarks.router, prefix="/api", tags=["Bookmarks"])
 app.include_router(
     bookmark_collections.router,
-    prefix="/api/bookmark-collections",
+    prefix="/api",
     tags=["Bookmark Collections"],
 )
 app.include_router(activities.router, prefix="/api/activities", tags=["Activities"])
 from app.routers import activity_heatmap
 
 app.include_router(activity_heatmap.router, prefix="/api", tags=["Activity Heatmap"])
-app.include_router(
-    conversations.router, prefix="/api/conversations", tags=["Conversations"]
-)
+app.include_router(conversations.router, prefix="/api", tags=["Conversations"])
 from app.routers import moderation
 
 app.include_router(moderation.router, prefix="/api", tags=["Moderation"])
-app.include_router(
-    conversations.router, prefix="/api/conversations", tags=["Conversations"]
-)
 from app.routers import audit
 
 app.include_router(audit.router, prefix="/api", tags=["Audit"])
@@ -596,12 +588,12 @@ from app.routers import profile_suggestions
 
 app.include_router(
     profile_suggestions.router,
-    prefix="/api/profile-suggestions",
+    prefix="/api",
     tags=["Profile Suggestions"],
 )
 app.include_router(
     profile_suggestions.router,
-    prefix="/api/users/me/profile-suggestions",
+    prefix="/api/users/me",
     tags=["Profile Suggestions"],
 )
 from app.routers import profile_views
@@ -621,12 +613,8 @@ app.include_router(
     prefix="/api/contributor-matching",
     tags=["Contributor Matching"],
 )
-app.include_router(
-    repositories.router, prefix="/api/repositories", tags=["Repositories"]
-)
-app.include_router(
-    organizations.router, prefix="/api/organizations", tags=["Organizations"]
-)
+app.include_router(repositories.router, prefix="/api", tags=["Repositories"])
+app.include_router(organizations.router, prefix="/api", tags=["Organizations"])
 
 from app.routers import workspace_api_tokens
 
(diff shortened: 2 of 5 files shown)


nensii21:

Issue #586, opened 2026-07-27
Build Notification Preferences Center


nensii21:
Description

Create a centralized interface where users can manage notification preferences.

Categories
Messages
Team Invitations
Project Updates
Mentions
System Announcements
Email Notifications

Acceptance Criteria
Preferences persist across sessions
Backend APIs implemented
User-friendly interface
Validation for all settings

nayanraj864-cmyk:
i want to work on this assign to me @nensii21

nensii21:
@nayanraj864-cmyk i'm assigning you all issues you've asked for will complete all of them?

nayanraj864-cmyk:
i working on the same project but other issue i need some time

PR #1252, opened 2026-08-21
fix(notifications): honour every stored preference, not four of seventeen

### LINKED ISSUE #1247: Ten of the seventeen notification preferences are stored and never consulted, and the two project toggles are joined with `or`
## Summary

`notification_preferences` stores seventeen toggles. `NotificationDispatcher._should_send_type` reads four of them, and `_should_send_channel` reads the three global channel switches. The remaining ten are written by the API, returned by the API, rendered in Settings, and never consulted when a notification is sent.

The settings page is not partly wired. For most categories it is decorative.

## Reproduction

Set one preference to `False`, dispatch the matching notification type, count what landed:

```
messages=False              -> MESSAGE          delivered=1
mentions=False              -> MENTION          delivered=1
team_invitations=False      -> PROJECT_INVITE   delivered=1
project_updates=False       -> PROJECT_UPDATE   delivered=1
system_announcements=False  -> SYSTEM           delivered=1
email_messages=False        -> MESSAGE (email)  delivered=1
```

Six for six.

## What is read and what is not

`dispatcher.py:53-66` in full:

```python
def _should_send_type(self, n_type, prefs) -> bool:
    if n_type in (NotificationType.PROJECT_UPDATE, NotificationType.PROJECT_INVITE):
        return prefs.project_updates or prefs.invitations
    if n_type == NotificationType.ROLE_CHANGE:
        return prefs.role_changes
    if n_type in (NotificationType.SYSTEM, NotificationType.WELCOME,
                  NotificationType.PASSWORD_RESET):
        return prefs.system_alerts
    return True
```

**Never read at all** — `messages`, `mentions`, `team_invitations`, `system_announcements`, and all five `email_*` per-category toggles.

The `email_*` group is the worst of it, because it is the one users reach for. Turning off "email me about messages" while keeping in-app messages is exactly what that column of checkboxes offers, and the dispatcher has no code path that could honour it: `_should_send_channel` only knows the global `email_enabled`. The choice is on or off for everything.

**Read, but joined with `or`** — `project_updates` and `invitations`:

```python
return prefs.project_updates or prefs.invitations
```

Two independent toggles, one result. Turn off project updates and invitations keeps them coming; turn off invitations and project updates keeps them coming. Neither switch does anything on its own, and the pair is off only when both are. Two rows in Settings that only work as one.

`team_invitations` is the current name for this — added with #586 — and `invitations` is the legacy one it replaced. The new one is unread; the old one is half-read.

**Falls through to `return True`** — `MESSAGE`, `MENTION`, `FOLLOW`, `APPLICATION`, `APPLICATION_ACCEPTED`, `APPLICATION_REJECTED`, `BUILDER_FLARE`, `SECURITY_ALERT`, `AI`. Nine of the fifteen types in the enum are undeniable, whatever the user has set. `SECURITY_ALERT` arguably should be; the other eight are ordinary product noise.

## The surface exists

This is not a half-built feature. `NotificationPreferenceUpdate` accepts all seventeen fields (`schemas/notification.py:113`), `NotificationPreferenceResponse` returns all seventeen, the model persists all seventeen, and the settings route writes them. A user turns something off, reloads, sees it off, and keeps getting the notifications.

## Expected

Every stored preference gates the thing it names.

- one category toggle per notification type, with no type silently exempt
- `email_<category>` gates email for that category, under the global `email_enabled`
- `team_invitations` and `project_updates` are independent, and each turns off exactly its own category
- the legacy `invitations` / `system_alerts` names keep working for rows written before #586

## Suggested fix

A single table from `NotificationType` to its category, used by both the type check and the per-channel email check, instead of an `if` chain that has to be remembered every time a type is added. A type missing from the table should be a loud failure in tests rather than an implicit `return True` — that default is how nine types ended up ungated.

Tests should assert both directions for every type: off means not delivered, on means still delivered. Only asserting the first half turns this into an outage.

MOHITKOURAV01:
Fixes #1247.

## The bug

`_should_send_type`, in full, before:

```python
if n_type in (NotificationType.PROJECT_UPDATE, NotificationType.PROJECT_INVITE):
    return prefs.project_updates or prefs.invitations
if n_type == NotificationType.ROLE_CHANGE:
    return prefs.role_changes
if n_type in (NotificationType.SYSTEM, NotificationType.WELCOME,
              NotificationType.PASSWORD_RESET):
    return prefs.system_alerts
return True
```

Seventeen columns in the model, four read here. Six for six on the issue's
reproduction:

```
messages=False              -> MESSAGE          delivered
mentions=False              -> MENTION          delivered
team_invitations=False      -> PROJECT_INVITE   delivered
project_updates=False       -> PROJECT_UPDATE   delivered
system_announcements=False  -> SYSTEM           delivered
email_messages=False        -> MESSAGE (email)  delivered
```

Three distinct faults, which is why it is three distinct pieces of work:

1. **`return True`.** Nine of the fifteen types were not named by the chain, so
   they were undeniable. Nothing anywhere made that visible.
2. **The five `email_*` columns were dead.** `_should_send_channel` only knew
   the global `email_enabled`, so the per-category email column of the settings
   matrix had no code path that could honour it.
3. **`or`.** `project_updates` and `invitations` are separate rows in Settings.
   Joined with `or`, neither did anything on its own and the pair was off only
   when both were.

## The fix

The `if` chain was not really the problem; its **default** was. A type nobody
remembered to add was silently delivered, and there was no file you could open
to see the omission.

So the mapping moves to `app/services/notifications/preferences.py`, where
every `NotificationType` has a row. The dispatcher keeps the same two questions
it always asked, and now delegates both:

```python
def _should_send_type(self, n_type, prefs):
    return should_deliver(n_type, prefs)
```

with the channel check gaining a second gate underneath the master switch:

```python
if channel_name == "email":
    if not prefs.email_enabled:
        return False
    if not should_deliver_by_email(n_type, prefs):
        return False
```

Three buckets, because "gated by a preference" is not the only honest answer
for fifteen types across six categories:

**Gated** — the twelve that map onto a category. Some of those mappings are
judgement calls and are named as such in the module: application outcomes and
builder flares are filed under project activity, and `AI` under system
announcements. Argue with them in review; that is what having them in one
readable table is for.

**Always delivered** — `PASSWORD_RESET` and `SECURITY_ALERT`. This is a
**behaviour change** and it is deliberate: `PASSWORD_RESET` used to sit under
`system_alerts`, so a user who switched off system announcements stopped being
told their password had been reset. Channel switches still apply — someone who
has turned email off entirely still gets it in-app rather than by email, and
there is a test for exactly that.

**Ungated** — `FOLLOW`. There is no preference column for it anywhere in the
model. Mapping it to whichever category looks closest would mean somebody's
"project updates" switch silently also controls follows. It stays ungated,
recorded in a set named `UNGATED`, so it reads as a known gap rather than an
oversight. Adding a column means a migration and a settings row; worth its own
issue.

## `and`, not "prefer the newer column"

`invitations` and `system_alerts` predate the #586 rename. Both old and new
columns still exist on every row and **both default to `True`**, so a row
written before the rename that opted out via `invitations` has
`team_invitations = True`.

Reading only the new column would silently re-enable notifications somebody had
already declined. So a category is on only if its own field and every legacy
alias are on — any `False` is a no. `test_a_legacy_opt_in_does_not_override_a_new_opt_out`
pins the other direction too.

## `should_deliver` fails open, on purpose

An unmapped type is delivered rather than dropped. Dropping someone's
notifications is a worse way to find out about a missing table entry than
sending one too many.

That is only defensible because something else is loud about it.
`test_every_notification_type_is_accounted_for` fails on any enum member that
appears in none of the three buckets, and names it. Strict in CI, safe in
production.

## Tests

`backend/tests/test_notification_preference_enforcement.py` — 62 tests.

Every category case is asserted **both ways**:

```python
def test_disabling_the_category_stops_the_notification(...)
def test_leaving_the_category_on_still_delivers(...)
```

Six of the twelve "off means silence" cases pass perfectly if the dispatcher
stops delivering anything at all. A preferences change that locks out the
people who want notifications is a worse bug than the one being fixed, so the
second half is not optional.

Also covered: the `or` fix as an explicit independence test, the five per-category
email columns (asserting the *email* stops and the in-app notification does
not), the global switch still winning, `role_changes` having no email column
and following `email_enabled` alone, both legacy aliases, the always-delivered
pair surviving every category switch while still respecting channel switches,
and an explicit `channels=[...]` list narrowing delivery rather than bypassing
preferences.

## A note on the existing test file

I nearly wrote this into `tests/test_notification_preferences.py` before
noticing it already exists. Its four tests cover storage and serialisation —
preferences are written, read back and validated correctly — and they pass on
`main` and always did.

That is the whole reason this was invisible. Everything about the preferences
worked except the part that consults them, and the test file's name suggested
the area was covered. The new file is
`test_notification_preference_enforcement.py` and says so in its docstring;
the original is untouched.

## Verification

Against the unfixed dispatcher, 28 of the 62 fail — the eleven category cases,
the independence test, the fifteen email cases, and the legacy-alias pair:

```
FAILED test_disabling_the_category_stops_the_notification[message-messages]
FAILED test_disabling_the_category_stops_the_notification[mention-mentions]
FAILED test_disabling_the_category_stops_the_notification[project_invite-team_invitations]
... 8 more categories ...
FAILED test_project_updates_and_invitations_are_independent
FAILED test_disabling_category_email_stops_the_email_only[message-email_messages]
... 14 more email cases ...
28 failed, 34 passed
```

The 34 that pass without the fix are the "on still delivers" half, which is
exactly right — `return True` delivers everything, including the things it
should.

Full backend suite, this branch and `main` in a clean worktree: **98 failures,
identical sets, no regressions.**

## Docs

`docs/notification_preferences_center.md` gains the type-to-category table, the
two not-gated buckets and why, the legacy-alias rule, and what to do when
adding a notification type.

### DIFF
--- a/backend/app/services/notifications/dispatcher.py
+++ b/backend/app/services/notifications/dispatcher.py
@@ -14,6 +14,10 @@ from app.services.notifications.channels.base import NotificationChannel
 from app.services.notifications.channels.database_channel import DatabaseChannel
 from app.services.notifications.channels.email_channel import EmailChannel
 from app.services.notifications.channels.websocket_channel import WebSocketChannel
+from app.services.notifications.preferences import (
+    should_deliver,
+    should_deliver_by_email,
+)
 
 logger = logging.getLogger(__name__)
 
@@ -40,30 +44,42 @@ class NotificationDispatcher:
         return prefs
 
     def _should_send_channel(
-        self, channel_name: str, prefs: NotificationPreference
+        self,
+        channel_name: str,
+        prefs: NotificationPreference,
+        n_type: NotificationType,
     ) -> bool:
+        """
+        Whether this channel carries this notification.
+
+        Two gates for email, not one. `email_enabled` is the master switch; the
+        per-category `email_<category>` columns are the ones the settings page
+        actually shows, and nothing read them before -- so "email me about
+        mentions but not messages" was on or off for everything.
+        """
         if channel_name == "database" and not prefs.database_enabled:
-            return False
-        if channel_name == "email" and not prefs.email_enabled:
             return False
         if channel_name == "websocket" and not prefs.websocket_enabled:
             return False
+        if channel_name == "email":
+            if not prefs.email_enabled:
+                return False
+            if not should_deliver_by_email(n_type, prefs):
+                return False
         return True
 
     def _should_send_type(
         self, n_type: NotificationType, prefs: NotificationPreference
     ) -> bool:
-        if n_type in (NotificationType.PROJECT_UPDATE, NotificationType.PROJECT_INVITE):
-            return prefs.project_updates or prefs.invitations
-        if n_type == NotificationType.ROLE_CHANGE:
-            return prefs.role_changes
-        if n_type in (
-            NotificationType.SYSTEM,
-            NotificationType.WELCOME,
-            NotificationType.PASSWORD_RESET,
-        ):
-            return prefs.system_alerts
-        return True
+        """
+        Whether the user wants this category at all.
+
+        The mapping lives in `preferences.py` rather than in an `if` chain
+        here: the chain's problem was that anything it did not name fell
+        through to `True`, so nine of the fifteen types were undeniable and
+        nothing made that visible.
+        """
+        return should_deliver(n_type, prefs)
 
     def dispatch(
         self,
@@ -93,7 +109,7 @@ class NotificationDispatcher:
 
         results = []
         for channel_name in channels:
-            if not self._should_send_channel(channel_name, prefs):
+            if not self._should_send_channel(channel_name, prefs, notification_type):
                 continue
 
             channel = self.channels.get(channel_name)
--- a/backend/app/services/notifications/preferences.py
+++ b/backend/app/services/notifications/preferences.py
@@ -1,0 +1,178 @@
+"""
+Which stored preference gates which notification.
+
+`notification_preferences` has seventeen columns. The dispatcher used to read
+four of them with an `if` chain, and everything the chain did not name fell
+through to `return True` -- so nine of the fifteen notification types were
+undeniable, and turning off "email me about messages" did nothing at all
+(#1247).
+
+The `if` chain was not the problem so much as its default. A type that nobody
+remembered to add was silently delivered, and there was no place where the
+omission was visible. This module is that place: every `NotificationType` has
+a row in `CATEGORY_BY_TYPE`, and a type without one raises in the tests rather
+than quietly defaulting to "send it".
+
+Three buckets, because "gated by a preference" is not the only honest answer:
+
+* **gated** -- the type belongs to a category the user can switch off;
+* **always delivered** -- password resets and security alerts, which are not
+  marketing and should not be suppressible by a checkbox;
+* **ungated** -- there is no column for it. Currently just ``FOLLOW``. Adding
+  one means a migration and a settings row, so it is recorded here as a known
+  gap instead of being mapped to whichever category looks closest.
+"""
+
+from __future__ import annotations
+
+from dataclasses import dataclass
+
+from app.models.notification import NotificationType
+
+
+@dataclass(frozen=True)
+class Category:
+    """
+    One row of the notification settings page.
+
+    ``enabled_field`` is the master switch for the category and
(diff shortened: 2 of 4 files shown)


nensii21:

PR #1228, opened 2026-08-20
feat: Add public project showcase page #630

Aryanbuha890:
## Description
This PR implements the **Public Project Showcase Page** (#630) which highlights featured, trending, recently launched, and active projects built by DevLink builders.

---

## 🏷️ Type of Change

- [x] New feature
- [ ] Bug fix
- [ ] Documentation update
- [ ] Refactoring
- [ ] Test improvement

---

## ✨ Features Added:
- **Featured Projects:** Highlights of prominent projects with high star counts.
- **Trending Projects:** Dynamic sorting by total project views.
- **Recently Launched:** Highlights of completed or near-completion projects.
- **Most Active:** Highlights of projects with large active teams.
- **Search & Filters:** Real-time client-side queries for tech stack, programming language, and difficulty.
- **Responsive Layout:** Responsive CSS Grid layout optimized for mobile, tablet, and desktop viewing.
- **SEO-friendly Metadata:** Fully implemented HTML head metadata tags with TanStack Router.
- **Pagination:** Simple next/prev page toggles for paginated views.
- **Empty State:** Visual card feedback when no projects match chosen search queries.

---

## 🔗 Related Issue
Closes #630

PR #1253, opened 2026-08-21
ci: make Type Check actually type check, and CI refuse hollow workflows

### LINKED ISSUE #1248: CI, Lint, Type Check and Security Scan check out the repository and stop — main has 11 type errors behind four green ticks

MOHITKOURAV01:
Fixes #1248.

## What was there

`typecheck.yml`, in full:

```yaml
name: Type Check
on: [pull_request]
jobs:
  typecheck:
    runs-on: ubuntu-latest
    steps:
      - uses: actions/checkout@v4
```

`ci.yml`, `lint.yml` and `security.yml` were the same file with a different
`name:`. `release-drafter.yml` was zero bytes.

Four green ticks on every pull request saying CI, Lint, Type Check and Security
Scan. A reviewer reads that as "this branch builds, lints, type checks and has
been scanned". None of it had been established.

## Read this part first: they are required status checks

I wrote this PR twice. The first version deleted the three that could not be
made real. It was wrong, and the way I found out is worth writing down.

```
$ gh api repos/nensii21/devlink/rules/branches/main
"required_status_checks": [
  frontend, backend, security, check-star, lint, ci, typecheck
]
```

Four of the seven — `ci`, `lint`, `security`, `typecheck` — belonged to the
workflows that checked nothing. **A required context that never reports is not
a failing check. It is a pull request that can never merge**, for everybody,
until an admin edits the ruleset. Deleting three files would have taken the
repository down until somebody with admin rights noticed.

GitHub told me, quietly: this branch went `BLOCKED` with **zero failing and
zero pending checks**, which is exactly what a context waiting forever looks
like. It is easy to read as "needs a review".

The same trap catches renames. A job's context is its `name:`, or its id when
it has none. My first version renamed `jobs.typecheck` to `jobs.frontend` for
tidiness, which dropped the required `typecheck` context *and* collided with
`frontend-ci.yml`'s job of the same name. Both jobs would have reported as
`frontend`.

So: nothing is deleted except the zero-byte `release-drafter.yml`, and every
job keeps its id, with a comment at each one saying why it may not be renamed.

## Type Check is now real, and it found what the issue said it would

```yaml
- uses: actions/setup-node@v7
- run: npm ci
- run: npx tsc --noEmit
```

Eleven errors on `main`, all from one cause: `src/routeTree.gen.ts` is
generated from the filenames in `src/routes/`, committed, and had drifted.
`showcase.tsx` (merged in #1228) and `verify-email.tsx` were live routes the
committed tree had never heard of, so every `Link to="/showcase"` and every
`Route.useSearch()` in those files was typed against a tree without them.

Regenerating it clears all eleven — `+42` lines and nothing else. The app was
never broken by this; `vite build` regenerates the tree, so only `tsc` and
everyone's editor read the committed copy.

**No build step in Type Check, deliberately.** Building first would regenerate
the tree and then type check a version that exists only on the runner. What
matters is whether the repository *as committed* type checks, because that is
what a contributor sees when they clone it.

## Keeping the tree honest

Generated-and-committed is a fine arrangement, but only if something fails when
the two disagree. Nothing did, which is the actual bug — a stale file was
inevitable eventually.

The check goes in `frontend-ci.yml`, immediately after the `npm run build` that
workflow already runs:

```yaml
- name: Check the committed route tree matches the generated one
  run: |
    if ! git diff --quiet -- src/routeTree.gen.ts; then
```

The build has already regenerated it, so this is a `git diff`, not a second
generation step. It costs nothing and runs on every PR.

I did try driving `@tanstack/router-generator` directly, to keep the check
inside `typecheck.yml` and skip the build. It works, but its output is missing
the `declare module '@tanstack/react-start'` footer that `tanstackStart`
appends, so the two disagree by ten lines and CI would flap. Reimplementing the
plugin's footer would have made the check itself the thing that rots. Comparing
against a real build is boring and correct.

## CI keeps its id and gets a job

`ci.yml` now checks the thing that went wrong:

```
13 workflows checked; each runs at least one real step.
All 7 required status checks are produced.
Still hollow, tracked in KNOWN_HOLLOW: lint.yml, security.yml
```

`.github/scripts/check_workflows.py` fails on two things:

1. a workflow whose jobs contain no step other than a checkout, a language
   setup or a cache restore — or an empty file;
2. a required status-check context with no job producing it.

The second is the one I wish had existed an hour ago. Renaming a job back gives:

```
::error::no job produces the required status check(s) typecheck. A job's
context is its `name:`, or its id when it has none -- renaming or deleting one
removes the context, and branch protection then waits forever on a check that
never arrives.
```

`REQUIRED_CONTEXTS` is mirrored by hand from the rules API rather than fetched,
because a fork running this in a pull request has no token that can read
repository rules. `test_the_required_contexts_are_the_ones_branch_protection_asks_for`
pins the list literally, so a ruleset change means editing a test on purpose
rather than the script drifting.

The rest is deliberately not clever. It does not judge whether a check is a
*good* check, only whether one exists and still answers to the expected name.
Scaffolding actions match on the part before the `@`, so Dependabot bumps and
SHA pins both keep working, and a job that delegates via `jobs.<id>.uses`
counts as real.

## Lint and Security Scan are still hollow

`KNOWN_HOLLOW` records both, with the reason attached, and the shape check
skips them:

```python
KNOWN_HOLLOW = {
    "lint.yml": "required status check; needs a lint baseline decision (#1248)",
    "security.yml": "required status check; superseded by #1244's dependency-security.yml",
}
```

Neither can simply be switched on. `ruff check` reports 3354 findings on
`backend/` and `eslint` reports 672 on `frontend/`, almost all formatting.
Turning either into a blocking check needs a baseline-or-changed-files
decision, and it is a policy change that would go red on several open PRs the
day it lands — not something to bundle into a fix for something else.

The shape that works on a backlog is linting only the files a PR touches, or a
ratcheting baseline that can only shrink. Either is a small self-contained PR.
`security.yml` may just want deleting once #1244's `dependency-security.yml` is
merged — but that needs the ruleset updated in the same breath, by someone with
admin rights.

Two tests keep the allowlist from becoming a graveyard: one fails if it names a
file that no longer exists, and one skips-with-a-message if an entry has
quietly become real.

## Still hollow, and worth knowing about

`backend-ci.yml` runs `pytest` and `black --check` under
`continue-on-error: true`. A red test suite does not fail the job, so
"backend ✓" means the workflow finished, not that the tests passed. Same defect,
different mechanism.

Not touched here: `main` has 98 pre-existing test failures, so removing
`continue-on-error` today just turns a required check permanently red. It needs
the failures triaged first. Flagging it rather than leaving it for the next
person to rediscover.

## Tests

`tests/test_check_workflows.py` — 32 tests.

On the shape check: a checkout-only job counts zero; **setup without a check
still counts zero** (install node, install deps, forget to run the tool — the
near-miss this exists to catch); `run:` steps and non-scaffolding actions count;
a malformed step counts as not-a-check rather than raising, because this
script's job is to fail a PR readably, not to traceback on one.

On the contexts: each of the seven is asserted individually, so a failure names
the check that stopped reporting rather than listing all of them.

And `test_every_workflow_in_this_repository_passes`, so a hollow workflow or a
dropped context fails locally too, before the PR.

## Verification

```
$ cd frontend && npx tsc --noEmit && echo $?
0

$ npm run build && git diff --quiet -- src/routeTree.gen.ts && echo clean
clean

$ python3 .github/scripts/check_workflows.py
13 workflows checked; each runs at least one real step.
All 7 required status checks are produced.
Still hollow, tracked in KNOWN_HOLLOW: lint.yml, security.yml

$ pytest tests/test_check_workflows.py -q
32 passed
```

No backend application code is touched, so the backend suite is unchanged.


nensii21:

PR #1254, opened 2026-08-21
fix(frontend): stop answering API failures with invented data

### LINKED ISSUE #1249: The frontend answers API failures with fabricated data, including a fake maintenance banner shown to every user
## Summary

Several frontend data paths catch an API failure and return invented data instead of surfacing the failure. The user sees numbers, names and dates that look like their project and are not.

This is not a loading placeholder or a seeded dev fixture. It renders in production, in the same components, with the same styling as real data, and there is nothing on screen that says otherwise.

## The three that fabricate content

**`api/modules/projectCollaborationMetrics.ts`** — any failure returns a fixed object:

```js
} catch (e) {
  console.warn('Backend API unavailable, using fallback mock metrics:', e);
}

return {
  active_members: 8,
  total_team_size: 12,
  avg_response_time_hours: 2.4,
  messages_exchanged: 342,
  tasks_completed: 29,
  applications_received: 14,
  collaboration_score: 92,
  daily_activity: [
    { date: '2026-08-04', activity_count: 12, messages: 45, tasks_completed: 4 },
    ...
  ],
};
```

A collaboration score of 92 and a seven-day chart, for a project with none of it. The dates are hardcoded to the week of 2026-08-04, so the chart is wrong in a way that is visible but easy to miss.

**`api/modules/teamActivity.ts`** — returns a five-item timeline of people who do not exist:

```
Sarah Connor joined the team              — 2 hours ago
Completed Milestone: MVP Auth & Database Schema
Uploaded architecture_v2.pdf
Started discussion: Real-time WebSocket Protocol Design
Updated Alex Mercer role to Project Lead
```

With avatars from `api.dicebear.com` and `created_at` computed from `Date.now()`, so the timestamps are always plausibly recent. The mock even honours the `activity_type` filter, so filtering the fake timeline works and reinforces that it is real.

**`components/shared/AnnouncementBanner.tsx`** — this one does not need a failure at all:

```js
const res = await api.get<Announcement[]>("/api/announcements/active");
return res && res.length > 0 ? res : MOCK_ANNOUNCEMENTS;
```

An empty list is a valid answer meaning "no announcements". It is treated as a failure, and every user with nothing to announce is shown

> **Scheduled Maintenance** — DevLink platform will undergo brief maintenance tonight from 2:00 AM to 3:00 AM UTC.

with `severity: "warning"`, dated `new Date()` so it is always tonight, on every page, forever. It is also the default value of the `useQuery`, so it renders during the first load before any request resolves.

A fake maintenance notice is the one piece of fake content users are most likely to act on.

## Why this has been invisible

`withFallback` in `services/index.ts` logs only in dev:

```js
if (import.meta.env.DEV) console.warn("[services] API call failed, using fallback:", err);
```

In production the swallow is silent. There is no error state, no toast, no console line, and the page looks fine — so a broken endpoint produces no signal from either side. The doubled-prefix routing bug filed alongside this one has eleven routers answering 404 on the legacy surface, and this is a large part of why nobody noticed.

## What is not in scope

`withFallback` is used widely and deliberately — the comment above it says the intent is to keep pages usable with no backend, and `isBackendConfigured()` gates a genuine local mock mode. Most of its uses fall back to `[]` or `null`, which is a defensible empty state, not a fabrication.

The problem is the handful that invent *content*: metrics, activity, announcements, and `MOCK_DATA` in `routes/_app.graph.tsx`. Those should fail visibly. The empty-array fallbacks can stay as they are — #1222 landed the empty-state components that render on top of them.

## Expected

- an API failure produces an error state, not content
- an empty successful response produces an empty state, not content
- offline/unconfigured mock mode, if it stays, is visibly labelled as mock rather than indistinguishable from real data

## Suggested fix

Let these three throw, and let the calling components render error and empty states. The pieces are already there: `components/ui/empty-state.tsx`, `components/shared/EmptyState.tsx` and `components/errors/SectionErrorBoundary.tsx`. Keep the `console.warn`, but unconditionally, not `DEV`-only: a silent catch in production is how this survived.

MOHITKOURAV01:
Fixes #1249.

## What was rendering

**`projectCollaborationMetrics.ts`** caught everything and returned a fixed
object: collaboration score 92, 342 messages exchanged, 8 of 12 members active,
and a seven-day chart hardcoded to the week of `2026-08-04`. A 401, a 500 or an
unreachable backend all produced a healthy-looking project.

**`teamActivity.ts`** returned five invented activities — "Sarah Connor joined
the team", "Uploaded architecture_v2.pdf", "Updated Alex Mercer role to Project
Lead" — with dicebear avatars and `created_at` computed from `Date.now()`, so
the timestamps were always plausibly recent. It also honoured the
`activity_type` filter, so filtering the fake timeline worked and made it look
more real.

**`AnnouncementBanner.tsx`** is the one that did not need a failure:

```js
return res && res.length > 0 ? res : MOCK_ANNOUNCEMENTS;
```

`[]` is a correct, successful answer meaning "nothing to announce". Treated as
a failure, so every user with no announcements saw

> **Scheduled Maintenance** — DevLink platform will undergo brief maintenance
> tonight from 2:00 AM to 3:00 AM UTC.

dated `new Date()`, so always tonight, on every dashboard page, forever. It was
also the `useQuery` default value, so it painted during the first load before
any request resolved. Of the three, a fake maintenance notice is the one users
are most likely to act on.

## The error states already existed

This is the part that changed how I wrote the fix.

`ProjectCollaborationMetrics.tsx` has a loading skeleton, an error panel with
an `AlertCircle`, the message, and a **Retry button wired to `fetchMetrics`**.
`TeamActivityTimeline.tsx` has an error panel and a "No activity events found"
empty state. Both components call the API inside a `try`/`catch` and set an
`error` state that could never be non-null, because the module below them
caught first.

So there is no new UI here. The fix is deleting three catches and letting the
components do what they were already written to do:

```diff
-  try {
-    const res = await api.get(...);
-    if (res && res.project_id) return res;
-  } catch (e) {
-    console.warn('Backend API unavailable, using fallback mock metrics:', e);
-  }
-  return { active_members: 8, collaboration_score: 92, ... };
+  api.get<ProjectCollaborationMetricsResponse>(
+    `/projects/${projectId}/collaboration-metrics`,
+  );
```

For the banner, an empty list and a failure both render nothing, which is what
`if (activeAnnouncements.length === 0) return null` already did.

## `withFallback` logs in production now

```diff
-    if (import.meta.env.DEV) console.warn("[services] API call failed, using fallback:", err);
+    console.warn("[services] API call failed, using fallback:", err);
```

Nine places, counting the eight in `collectionsService`.

In production the swallow was completely silent: no error state on the page,
nothing in the console, and a screenful of plausible content. That is why a
broken endpoint produced no signal from either side — and it is a large part of
why eleven backend routers could 404 on the legacy `/api` surface (#1246)
without anyone noticing.

## Something I found while in here

`withFallback` has 35 call sites. The issue said most fall back to `[]` or
`null`, which is a defensible empty state — that is still true, but **five fall
back to seeded content**:

```
services/index.ts:129  projectsApi.list      -> seed.projects
services/index.ts:148  buildersApi.list      -> seed.builders
services/index.ts:151  buildersApi.trending  -> seed.builders.slice(3, 6)
services/index.ts:322  messagesApi.conversations -> seed.conversations
services/index.ts:499  notificationsApi.list -> seed.notifications
```

`withFallback` uses the same value for both branches — mock mode *and* an error
with a real backend configured — so a configured backend that fails shows
seeded projects, builders, conversations and notifications as if they were the
user's.

I have not fixed it here. Splitting the two branches means either a signature
change across all 35 call sites or letting errors propagate to pages that have
no error handling, and both are a product decision rather than a bug fix.
Recording it with line numbers so it does not have to be found again.

`routes/_app.graph.tsx` looked like a fourth case and is not: its `MOCK_DATA`
is only used when `!isBackendConfigured()`, which is documented mock mode. A
real failure already propagates to `useQuery`. Left alone.

## Tests

20 tests across three files.

**`projectCollaborationMetrics.test.ts`** — errors propagate; a genuinely quiet
project (score 0, no members, no activity) comes back as itself rather than
falling through the old `if (res && res.project_id)` guard; and one test
asserts the response is not `92` / `342` / `2.4` by name, so reintroducing the
fallback fails loudly rather than looking plausible.

**`teamActivity.test.ts`** — errors propagate; an empty timeline stays empty;
real items come back with their real names and not `Sarah Connor`. Also covers
URL construction, including `encodeURIComponent` on the filter — the old code
interpolated it raw, so a value containing `&` could inject query parameters.
That is a small fix that came with the rewrite and is asserted rather than
assumed.

**`AnnouncementBanner.test.tsx`** — nothing renders on empty, nothing renders
on failure, and **nothing renders while the request is in flight** (the mock
was the `useQuery` default, so it painted before any answer arrived). Plus the
positive cases: real announcements render, dismissal hides one, dismissal
survives a remount, and a *different* announcement still shows after an earlier
one was dismissed — a banner change that silently swallows real announcements
would be a worse bug than the one being fixed.

## Verification

Against the unfixed source, 8 of the 20 fail:

```
× propagates a failure instead of inventing metrics
× does not substitute a fallback for an unauthorised response
× propagates a failure instead of inventing a timeline
× encodes the filter so a value with a space cannot break the query string
× renders nothing when there are no announcements
× renders nothing when the request fails
× renders nothing while the request is still in flight
× keeps a dismissal across a remount
8 failed | 12 passed
```

Full frontend suite: **50 files, 721 tests, all passing.**

`npx tsc --noEmit` reports the same 10 pre-existing errors as `main` and no new
ones. Those are the stale `routeTree.gen.ts` and are fixed by #1253, not here.

Prettier is clean on every file this touches. `services/index.ts` reports
pre-existing formatting drift at lines 143 and 205–238; my edits are at 43–60
and 695–889 and I have not reformatted the rest of the file.

## Docs

`docs/frontend_data_fallbacks.md` writes down the three cases — failed, empty,
and backend-not-configured — the "only ever fall back to something empty" rule,
and a note to check whether the component already has the error and empty
states before writing new ones, because in all three of these it did.

### DIFF
--- a/docs/frontend_data_fallbacks.md
+++ b/docs/frontend_data_fallbacks.md
@@ -1,0 +1,67 @@
+# Fallbacks and empty states in the frontend
+
+One rule:
+
+> **Never render invented content as if it were the user's data.**
+
+A failed request and an empty result are different things, and neither of them
+is content. Three code paths broke this and shipped (#1249): invented
+collaboration metrics, an invented team timeline, and a hardcoded maintenance
+banner shown to every user who had no announcements.
+
+## The three cases
+
+| Situation | What to render |
+|---|---|
+| Request failed | an error state, with a retry if the component has one |
+| Request succeeded, result is empty | an empty state |
+| Backend not configured (`isBackendConfigured()` is false) | the seeded mock data — this is local mock mode and is deliberate |
+
+The third is a development convenience and only applies when
+`VITE_API_BASE_URL` is unset. It must never be reached because a *configured*
+backend failed.
+
+## Why an empty result is the easiest one to get wrong
+
+```js
+return res && res.length > 0 ? res : MOCK_ANNOUNCEMENTS;
+```
+
+That reads as defensive. It is not: `[]` is a correct, successful answer
+meaning "there is nothing to announce", and treating it as a failure showed
+every user a maintenance notice that did not exist. Dated `new Date()`, so it
+was always tonight.
+
+`if (res && res.project_id)` had the same shape — anything falsy fell through
+to the mock, so "quiet project" and "backend is down" produced the same very
+busy screen.
+
+## Why this hid other bugs
+
+`withFallback` in `services/index.ts` logged only under `import.meta.env.DEV`.
+In production a swallowed failure produced no signal from either side: no error
+on the page, nothing in the console, and a screenful of plausible content.
+
+Eleven backend routers were 404ing on the legacy `/api` surface (#1246) and
+nobody noticed, partly for this reason. The warning is unconditional now.
+
+## Writing a fallback
+
+Only ever fall back to something **empty**: `[]`, `null`, a zeroed summary. The
+caller cannot tell a fallback from a real answer, so anything with content in
+it becomes indistinguishable from data the user actually has.
+
+If a component needs to distinguish "nothing" from "we do not know", it needs
+the error, not a fallback — let the call throw.
+
+## The components are usually ready for this
+
+Worth checking before writing new UI. `ProjectCollaborationMetrics` already had
+a loading skeleton, an error panel and a Retry button; `TeamActivityTimeline`
+already had an error panel and a "No activity events found" state. Both were
+complete and unreachable, because the API module caught everything before it
+got there.
+
+Shared pieces: `components/ui/empty-state.tsx`,
+`components/shared/EmptyState.tsx`, `components/shared/EmptySearchState.tsx`,
+`components/errors/SectionErrorBoundary.tsx`.
--- a/frontend/src/api/__tests__/projectCollaborationMetrics.test.ts
+++ b/frontend/src/api/__tests__/projectCollaborationMetrics.test.ts
@@ -1,0 +1,87 @@
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+
+import { getProjectCollaborationMetrics } from "../modules/projectCollaborationMetrics";
+import { api } from "../client";
+
+const PROJECT_ID = 42;
+
+const REAL_RESPONSE = {
+  project_id: PROJECT_ID,
+  active_members: 2,
+  total_team_size: 3,
+  avg_response_time_hours: 9.5,
+  messages_exchanged: 4,
+  tasks_completed: 1,
+  applications_received: 0,
+  collaboration_score: 31,
+  daily_activity: [],
+};
+
+describe("getProjectCollaborationMetrics", () => {
+  beforeEach(() => {
+    vi.spyOn(api, "get").mockResolvedValue(REAL_RESPONSE as never);
+  });
+
+  afterEach(() => {
+    vi.restoreAllMocks();
+  });
+
+  it("targets the project-scoped metrics endpoint", async () => {
+    await getProjectCollaborationMetrics(PROJECT_ID);
+
+    expect(api.get).toHaveBeenCalledWith(`/projects/${PROJECT_ID}/collaboration-metrics`);
+  });
+
+  it("returns what the backend sent", async () => {
+    await expect(getProjectCollaborationMetrics(PROJECT_ID)).resolves.toEqual(REAL_RESPONSE);
+  });
+
+  // The regression. This used to catch everything and return a fixed object,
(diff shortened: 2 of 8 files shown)


nensii21:

PR #1230, opened 2026-08-20
feat: Implement auto-save for profile editing #731

Aryanbuha890:
## Description
This PR implements **Auto-Save for Profile Editing** (#731), introducing debounced background saving, visual status indicators, error recovery, and optimistic locking conflict resolution.

---

## 🏷️ Type of Change

- [x] New feature
- [ ] Bug fix
- [ ] Documentation update
- [ ] Refactoring
- [ ] Test improvement

---

## ✨ Features Added:
- **Enable Auto-Save Option:** Added a toggle switch to optionally turn on auto-saving behavior.
- **Debounced Save:** Automates saving changes exactly 1.5 seconds after the user stops typing in the Full Name, Username, Email, or Bio fields.
- **Save Status Indicator:** Displays pulsing labels for states: `All changes saved`, `Saving...`, `Unsaved changes`, `Error saving changes`, and `Version conflict`.
- **Conflict Handling Banner:** If an optimistic locking conflict (HTTP 409) is returned, displays resolution buttons:
  - **Force Overwrite:** Save edits overriding database values.
  - **Discard & Reload Latest:** Re-fetches the latest profile details from the database.

---

## 🔗 Related Issue
Closes #731

PR #1320, opened 2026-08-24
fix(settings): import the hook it uses, and stop reporting a failed save as a save

### LINKED ISSUE #1315: `/settings` cannot render: `useRef` is used but never imported, and `usersService.getMe` does not exist
## Summary

`/settings` throws before it paints anything. `frontend/src/routes/_app.settings.tsx:131` calls
`useRef`, and the import at the top of the file is:

```ts
import { useState, useEffect } from "react";
```

React hooks are not globals, so the first render raises `ReferenceError: useRef is not defined`
and the route renders as whatever error boundary is above it. There is nothing partial about
this — the page has no working state.

Two more faults sit behind the first one, in the same feature, and only become reachable once
the page renders at all.

## Reproduction

`main`, clean checkout:

```console
$ cd frontend && npx tsc --noEmit
src/routes/_app.settings.tsx(131,30): error TS2552: Cannot find name 'useRef'. Did you mean 'User'?
src/routes/_app.settings.tsx(138,41): error TS2339: Property 'getMe' does not exist on type
  '{ getPrivacySettings: () => Promise<any>; updatePrivacySettings: (...) => Promise<unknown>;
     updateMe: (...) => Promise<...>; }'.
```

and the tests that cover the page:

```console
$ npx vitest run --exclude "e2e/**"
 ❯ src/components/__tests__/settings-tabs.test.tsx (2 tests | 2 failed)
     × renders 6 distinct navigation tabs for settings
     × switches to developer accounts tab properly

ReferenceError: useRef is not defined
 ❯ SettingsPage src/routes/_app.settings.tsx:131:30

 Test Files  1 failed | 52 passed (53)
      Tests  2 failed | 724 passed (726)
```

Both errors, and both test failures, are the same two lines.

## Fault 1 — `useRef` is used but never imported

`git blame` puts line 131 on `036cf92ab` ("feat: implement auto-save for profile editing #731",
#1230). The auto-save debounce needs a ref; the ref was added, the import was not.

## Fault 2 — `usersService.getMe` does not exist

Line 138 loads the profile the page edits:

```ts
const user = await usersService.getMe();
```

`usersService` (`src/services/index.ts:203`) has exactly three members — `getPrivacySettings`,
`updatePrivacySettings`, `updateMe`. There is no `getMe`, and `usersApi`
(`src/api/modules/users.ts`) has no `me()` for it to delegate to either, even though
`auth-context.tsx:35` already reads the same `/api/users/me` by hand.

At runtime this is `TypeError: usersService.getMe is not a function` inside the mount effect —
so even with fault 1 fixed the page loads with empty fields and never populates.

## Fault 3 — a failed save reports success

`usersService.updateMe` is:

```ts
updateMe: (body: Record<string, any>) =>
  withFallback(
    () => usersApi.updateMe(body),
    {}
  ),
```

`withFallback` catches everything and returns the fallback (`src/services/index.ts:55`). So the
page's own conflict handling is unreachable:

```ts
const updatedUser = await usersService.updateMe(payload);
if (updatedUser) {                       // {} is truthy
  ...
  version: (updatedUser as any).version || 1,   // resets the optimistic-concurrency version
  setSaveStatus("saved");
  toast.success("Profile saved successfully");
}
} catch (err: any) {
  if (err?.status === 409 || err?.statusCode === 409) {   // never runs
    setSaveStatus("conflict");
```

The `version` field exists to make `PUT /api/users/me` fail with 409 when the profile changed
elsewhere. The 409 is swallowed by `withFallback`, the page reports "Profile saved
successfully", and `version` is reset to `1` — which means the *next* save is sent with a stale
version too. The user is told their edit landed when it did not.

This is the same shape as #1249: an API failure answered with invented data instead of an error.
`updateMe` is a write, and a write has no sensible fallback value.

## Expected

- `/settings` renders.
- `npx tsc --noEmit` is clean, and `settings-tabs.test.tsx` passes.
- Loading the profile goes through a `usersService`/`usersApi` method that actually exists.
- A save that fails is reported as a failure, and a 409 reaches the conflict branch the page
  already has.

MOHITKOURAV01:
Fixes #1315.

## The bug

Three faults stacked in the profile tab, and you meet them in order.

**`useRef` was never imported.** Line 131 uses it, line 1 imports `useState` and
`useEffect`. React hooks are not globals, so the first render raised
`ReferenceError: useRef is not defined` and `/settings` was whatever error
boundary sits above it. `git blame` puts the ref on `036cf92ab` (#1230, the
auto-save debounce); the import never came with it.

**`usersService.getMe` did not exist.** `usersService` had exactly
`getPrivacySettings`, `updatePrivacySettings` and `updateMe`, and `usersApi` had
no `me()` for it to delegate to either — even though `auth-context.tsx:35`
already reads `/api/users/me` with a hand-rolled `api.get`. With fault 1 fixed
this is `TypeError: usersService.getMe is not a function` inside the mount
effect, so the form stays blank forever.

**A failed save was reported as a save.** `updateMe` was:

```ts
updateMe: (body) => withFallback(() => usersApi.updateMe(body), {}),
```

`withFallback` catches everything and answers with the fallback, so the page's
own conflict handling could not run:

```ts
const updatedUser = await usersService.updateMe(payload);
if (updatedUser) {                              // {} is truthy
  version: (updatedUser as any).version || 1,   // resets the version token
  toast.success("Profile saved successfully");
}
} catch (err) {
  if (err?.status === 409) { setSaveStatus("conflict"); }   // unreachable
```

`version` is the optimistic-concurrency token: send it and the server answers
409 when the row moved underneath you. Swallowing the 409 meant the user was
told their edit landed when it had not, *and* `version` was reset to `1`, so the
next save was sent stale too. One conflict, then every subsequent save silently
wrong.

## The fix

The first two are one line each. The third is the one worth explaining.

`withFallback` is a read-side tool and its own comment says so — "only ever pass
an *empty* fallback", because the caller cannot tell a fallback from an answer.
That reasoning does not extend to a write. There is no empty value that means
"your change was applied"; there are two outcomes and both need to reach the
caller. So writes get their own helper:

```ts
async function write<T>(call: () => Promise<T>, mockResult: T): Promise<T> {
  if (!isBackendConfigured()) return mock(mockResult);
  return call();
}
```

Mock mode still gets an answer. A configured backend gets the real call, errors
included. `updateMe` and `updatePrivacySettings` both move onto it.

`getMe` stays on `withFallback`, with `null` as the fallback rather than a user
object — the page renders whatever comes back into its form fields, so an
invented profile would be displayed as the user's own data and then written back
over the real row. `null` is a shape the page can tell apart, and it now says so
instead of showing an empty form with no explanation.

`usersApi.me()` and a `CurrentUserProfile` type are new. The type is wider than
`AuthUser` on purpose: that one describes the summary embedded in a login
response, this is the profile row the settings page edits. Three `(user as any)`
casts in the page went away as a result.

## ResizeObserver

Unrelated to the bug, but it blocks testing the fix. jsdom does not implement
`ResizeObserver`, and Radix's `Switch` — which the profile tab renders — builds
one in a layout effect. Any test that *waits* for this tree dies with an
uncaught `ReferenceError` after its assertions have already passed, attributed
to whichever test happened to be running. `settings-tabs.test.tsx` only escaped
it by finishing before the effect flushed.

A no-op stub in `test-setup.ts`, guarded so a real implementation wins.

## Tests

`src/services/__tests__/usersService.test.ts` — `getMe` exists and reads the
profile; it answers `null` rather than a fabricated user; `updateMe` passes the
version through, rejects on failure, and lets a 409 out.

`src/components/__tests__/settings-profile.test.tsx` — the page renders without
throwing (fault 1 on its own fails all seven); it loads through `getMe` and
fills the form; a profile that will not load says so; the loaded version is sent
with the save; a failed save shows the error and does not toast success; a 409
reaches the conflict state; and the version the server returned survives into
the next save.

```console
$ cd frontend && npx tsc --noEmit
$ npx vitest run --exclude "e2e/**"
 Test Files  55 passed (55)
      Tests  738 passed (738)
```

Before this branch: two `tsc` errors, and `settings-tabs.test.tsx` red.


nensii21:

PR #1321, opened 2026-08-24
ci(frontend): run the tests, and fail when a suite stops being run

### LINKED ISSUE #1316: The frontend test suite has never run in CI: 726 tests, and no workflow invokes them
## Summary

`frontend/package.json` has had a `test` script since the suite was introduced:

```json
"test": "vitest run --exclude \"e2e/**\"",
```

Nothing in `.github/workflows/` ever runs it.

```console
$ grep -rn "vitest\|npm run test\|npm test\|playwright" .github/workflows/
$ echo $?
1
```

53 test files, 726 tests, and the only signal any of them has ever produced is on the machine of
whoever happened to run them locally.

## What the frontend workflow does instead

`frontend-ci.yml` installs, builds, and compares the committed route tree against the generated
one. All three are worth keeping. None of them execute a test.

That gap is not theoretical. `src/components/__tests__/settings-tabs.test.tsx` fails on `main`
right now, and has since #1230 merged:

```console
$ cd frontend && npx vitest run --exclude "e2e/**"
 ❯ src/components/__tests__/settings-tabs.test.tsx (2 tests | 2 failed)
ReferenceError: useRef is not defined
 ❯ SettingsPage src/routes/_app.settings.tsx:131:30

 Test Files  1 failed | 52 passed (53)
      Tests  2 failed | 724 passed (726)
```

A test written specifically to cover the settings page caught the settings page being broken,
and the result went nowhere. `frontend` is a required status check and it was green for that
merge, because building a module and rendering it are different questions — `vite build` is
happy to bundle a component that throws the moment React calls it.

## Why the existing guard did not catch this

`.github/scripts/check_workflows.py` (#1248) refuses a workflow whose jobs are all scaffolding.
`frontend-ci.yml` passes it comfortably: install, build and the route-tree comparison are three
substantive steps.

The check answers "does this workflow do *something*". The failure mode here is different and
narrower: a test suite exists in the repository, it is wired into `package.json`, and no
workflow invokes it. That is invisible to a step-counting check, and it is invisible on the pull
request page, which is the part that matters — a reviewer looking at a green `frontend` tick has
no way to tell that the tests were never part of it.

## Expected

- The frontend suite runs on every pull request and its result gates the merge.
- A suite that exists but is never invoked fails CI, the same way a workflow that checks nothing
  already does — otherwise the next one to be quietly dropped is as unnoticeable as this one.

`settings-tabs.test.tsx` has to be green before the suite can gate anything; that is #1315.

MOHITKOURAV01:
Fixes #1316.

> **Stacked on #1320.** This branch carries that commit so its own `frontend`
> check can be green — turning the suite on while `settings-tabs.test.tsx` is
> red would just be a red tick for a different reason. Merge #1320 first and
> this diff collapses to the workflow, the script and its tests. Review the
> second commit.

## The bug

```console
$ grep -rn "vitest\|npm run test\|npm test\|playwright" .github/workflows/
$ echo $?
1
```

53 test files. 726 tests. No workflow has ever run one of them.

`frontend-ci.yml` installs, builds, and checks the committed route tree against
the generated one. All three are worth having and none of them executes a test.
`vite build` will bundle a component that throws the moment React calls it —
which is exactly what happened: `settings-tabs.test.tsx` was written to cover
the settings page, has been red since #1230, and the `frontend` context was
green for that merge and every merge since.

## The fix, part one

A `Test` step in `frontend-ci.yml`, after the build. `e2e` stays excluded, as it
is in `npm run test` — Playwright needs browsers and a server, which is a
separate job for whoever wants it.

## The fix, part two

`check_workflows.py` already refuses a workflow whose jobs are all scaffolding
(#1248). `frontend-ci.yml` sailed through it: install, build and route-tree
comparison are three substantive steps. The check answers *"does this workflow
do something"*, and the thing that went wrong here is narrower — a suite exists,
is wired into `package.json`, and nothing invokes it.

So `SUITES`, listing the suites the repository ships and how to spot one being
run. Two ways to fail it:

- **nothing runs it.** What happened to the frontend suite.
- **it runs, but never gates.** What still happens to `pytest`: `backend-ci.yml`
  has carried `continue-on-error: true` on that step since the workflow was
  written, so the `backend` context is green whatever the tests say.

The second one is real and I am not fixing it here. Dropping the flag turns the
existing failures on `main` into a blocked repository, which has to be dealt
with first and separately. It goes in `KNOWN_UNGATED` with a reason, next to
`KNOWN_HOLLOW` and for the same reason — a debt with a name attached rather than
a silence. The script fails if a suite listed there starts gating, so paying one
off cannot leave the note behind lying about it.

### Two details that are not obvious

**Patterns match the start of a command, not anywhere in the `run:` block.** The
first draft searched the whole block, and this line —

```yaml
run: pip install black ruff mypy pytest
```

— is in a step that *does* gate. So the backend suite looked gated, the
`continue-on-error` two steps down was invisible, and the check passed for
exactly the wrong reason. `commands_in()` splits a block on newlines, `&&`,
`||`, `;` and `|` so `cd backend && pytest` is still seen, and the patterns
anchor to `^`. `npm run test` is matched without a trailing `:` so
`npm run test:e2e` — a different suite — does not satisfy it.

**`Suite` is a `NamedTuple`, not a `@dataclass`.** `tests/test_check_workflows.py`
loads the script with `spec_from_file_location`, which leaves
`sys.modules[__module__]` as `None`; `@dataclass` resolves its annotations
through exactly that and raises during class creation. A `NamedTuple` does not
care, and keeping the module exec-able from its path is worth more than the
`@dataclass` sugar.

## Tests

24 new cases in `tests/test_check_workflows.py`: command splitting (including
that `|` separates but `||` is not two of them), what each suite pattern does
and does not recognise — with `pip install ... pytest` in there by name —
`continue-on-error` at step and job level, one gating invocation being enough,
and a `KNOWN_UNGATED` entry that has gone stale being an error rather than a
skip. Plus the check run against the repository's own workflows, so a failure
names the suite rather than just exiting 1.

```console
$ python .github/scripts/check_workflows.py
14 workflows checked; each runs at least one real step.
All 7 required status checks are produced.
All 2 test suites are run by a workflow.
Still hollow, tracked in KNOWN_HOLLOW: lint.yml, security.yml
Runs but does not gate, tracked in KNOWN_UNGATED: backend

$ python -m pytest tests/test_check_workflows.py -q
68 passed
```

And with the new Test step deleted again, which is the case that matters:

```console
$ python .github/scripts/check_workflows.py
::error::One or more workflows do not check anything.
  - nothing in .github/workflows runs the frontend unit tests (vitest). The suite
    exists (frontend/package.json -> scripts.test) and no pull request has ever
    been told what it thinks. ...
```

### DIFF
--- a/.github/scripts/check_workflows.py
+++ b/.github/scripts/check_workflows.py
@@ -31,14 +31,30 @@ missing context is not a failing one, every pull request then waits forever on
 a check that will never arrive. That is a repository-wide outage caused by a
 one-word edit, so it is worth a test.
 
+`SUITES` is the same idea one level up. A workflow can run a real step and
+still leave the repository's actual tests on the floor: `frontend-ci.yml`
+installed, built, and compared the route tree -- three substantive steps, a
+green tick, and 726 vitest tests that no workflow had ever executed (#1316).
+Two ways for a suite to stop mattering, and this checks both:
+
+* nothing invokes it, which is what happened to the frontend suite;
+* something invokes it under `continue-on-error: true`, which is what still
+  happens to `pytest` -- the job reports success whatever the tests say.
+
+`KNOWN_UNGATED` is the second case with a name attached, the same way
+`KNOWN_HOLLOW` is for workflows. An entry there is a debt, not an exemption,
+and the script says so when one is paid off.
+
 Deliberately not clever. It does not try to judge whether a step is a *good*
 check, only whether one exists and still answers to the name `main` expects.
 """
 
 from __future__ import annotations
 
+import re
 import sys
 from pathlib import Path
+from typing import Iterator, NamedTuple
 
 import yaml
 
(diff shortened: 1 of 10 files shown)


nensii21:

PR #1322, opened 2026-08-24
fix(followers): one notification per follow, and a duplicate that stays contained

### LINKED ISSUE #1317: `POST /api/followers/{id}` sends two notifications, and a duplicate follow leaves the session unusable
## Summary

`POST /api/followers/{user_id}` writes the "new follower" notification twice, from two different
places, with two different titles and two different links. And when the same follow arrives
twice, the `IntegrityError` from the unique constraint is never handled, which leaves the
request's `Session` in a rolled-back state — so the failure is not confined to the follow.

## Fault 1 — two notifications per follow

`FollowerService.follow_user` already notifies (`app/services/follower_service.py:73`):

```python
NotificationService.create_follow_notification(
    db=db,
    recipient_id=following_id,
    actor_id=follower_id,
    title="New Follower",
    message=f"{follower_name} started following you.",
    action_url=f"/profile/{follower_username}" if follower_username else None,
)
```

and then the router does it again, by hand, on the line after the call
(`app/routers/followers.py:57`):

```python
follow = FollowerService.follow_user(db, current_user.id, user_id)

try:
    NotificationService.enqueue(
        db,
        recipient_id=user_id,
        sender_id=current_user.id,
        type=NotificationType.FOLLOW,
        title="New follower",
        message=f"{current_user.username} started following you.",
        action_url=f"/users/{current_user.id}",
    )
except Exception as e:
    print(f"ENQUEUE ERROR: {e}")
```

Same event, twice, differing in every field:

| | service | router |
|---|---|---|
| title | `New Follower` | `New follower` |
| message | `Ada Lovelace started following you.` | `ada started following you.` |
| `action_url` | `/profile/ada` | `/users/9f1c…` |

The recipient gets two entries in their notification list for one follow, and the two link to
different pages. Which one the bell shows first comes down to `created_at` ordering between two
writes in the same request.

The `except Exception: print(...)` is a fault of its own. Every other failure path in the
backend goes through `structlog`; this one writes to stdout, so a broker outage here produces
nothing a log query can find, and pollutes test output instead.

## Fault 2 — the second follow poisons the session

`follow_user` inserts unconditionally:

```python
relationship = Follower(follower_id=follower_id, following_id=following_id)
db.add(relationship)
db.flush()
```

The router guards it with a `SELECT` first, which is check-then-act: two concurrent
`POST /api/followers/{id}` from the same user both see no row and both insert. The unique
constraint on `(follower_id, following_id)` then raises `IntegrityError` inside the `flush`, and
nothing catches it or rolls back.

`app/tests/test_followers.py::test_prevent_duplicate_follow` shows what that costs. It fails on
`main`:

```console
$ cd backend && python -m pytest app/tests/test_followers.py::test_prevent_duplicate_follow
E  sqlalchemy.exc.PendingRollbackError: This Session's transaction has been rolled back due to a
E  previous exception during flush. To begin a new transaction with this Session, first issue
E  Session.rollback().
```

Note where that is raised — not on the failing insert, but on the *next* statement to touch the
session. In a request that means everything after it fails too: the response serialiser reading
a lazy attribute, the activity write, the request-logging middleware's insert. The caller gets a
500 from a line that has nothing to do with following anybody, and the audit trail for the
request is lost along with it.

## Reproduction

```console
$ cd backend && python -m pytest app/tests/test_followers.py::test_prevent_duplicate_follow -q
FAILED app/tests/test_followers.py::test_prevent_duplicate_follow
E  sqlalchemy.exc.PendingRollbackError: ... Original exception was:
E  (sqlite3.IntegrityError) UNIQUE constraint failed: followers.follower_id, followers.following_id
```

For the duplicate notification, follow someone and read the recipient's list — two rows come
back for the one follow, `New Follower` from the service and `New follower` from the router.

## Expected

- One notification per follow, with one title and one link.
- Following someone you already follow is answered with the existing relationship, not a 500,
  and does not take the rest of the request down with it.
- The enqueue failure path logs like the rest of the backend does.

MOHITKOURAV01:
Fixes #1317.

## Fault 1 — two notifications, one follow

The service notified (`follower_service.py:73`), and then the router notified
again on the line after the call (`routers/followers.py:57`). Same event, twice,
differing in every field:

| | service | router |
|---|---|---|
| title | `New Follower` | `New follower` |
| message | `Ada Lovelace started following you.` | `ada started following you.` |
| `action_url` | `/profile/ada` | `/users/9f1c…` |

Two rows in the recipient's list for one follow, linking to two different pages,
ordered by whichever write landed first.

The router's copy goes. The service has owned this since it was written, uses
the display name rather than the handle, and links to the profile the follower
actually has — it is the better of the two, and more to the point there should
only be one.

Its `except Exception: print(f"ENQUEUE ERROR: {e}")` goes with it.
`NotificationService.enqueue` already catches a broker outage, falls back to
inline dispatch, and logs through `structlog`; the try/except added nothing
except a stdout write no log query would ever find.

## Fault 2 — the duplicate took the request with it

`follow_user` inserted unconditionally. The router's `SELECT` in front of it is
check-then-act, so two concurrent `POST /api/followers/{id}` from the same user
both pass it and one hits the unique constraint on
`(follower_id, following_id)`.

The `IntegrityError` itself is not the interesting part. An unhandled one leaves
the `Session` in a rolled-back state, and SQLAlchemy raises
`PendingRollbackError` on the *next* statement — so the visible failure is the
response serialiser reading a lazy attribute, or the activity write, or the
request-logging middleware's insert. A 500 from a line with nothing to do with
following anybody, and the request's audit trail gone with it.

So `follow_user` is idempotent now, and the insert runs inside a `SAVEPOINT`:

```python
existing = FollowerService.get_relationship(db, follower_id, following_id)
if existing is not None:
    return existing

try:
    with db.begin_nested():
        db.add(relationship)
        db.flush()
except IntegrityError:
    existing = FollowerService.get_relationship(db, follower_id, following_id)
    if existing is None:
        raise
    return existing
```

The `SELECT` handles the ordinary "already following" case without a round trip
through an exception. The savepoint handles the race the `SELECT` cannot: losing
it rolls back to the savepoint rather than the transaction, so the outer
transaction survives and the request answers normally. The loser returns the
winner's row and sends nothing — the winner already recorded the activity and
sent the notification.

The re-`raise` when the row still is not there matters. `IntegrityError` covers
every constraint on the table, and a dangling `following_id` is a different
problem; swallowing it would turn a bad request into a silent no-op.

The router keeps its 400 for the common case. It is a nicer answer than a
duplicate 201 and it is now only an optimisation, not the thing standing between
the API and a poisoned session.

## Tests

`backend/tests/test_follow_idempotency.py`, 8 cases: one follow enqueues one
notification, with the service's wording and link; the same over the API, which
is the end the bug was reported from and where the count was 2; following twice
creates one relationship, enqueues once, and leaves the session usable;
losing the insert race returns the winning row and sends nothing; and an
`IntegrityError` that is not a duplicate still propagates.

Counted at `NotificationService.enqueue` rather than by reading rows back. The
fault is at the enqueue, and it keeps these tests off the dispatch path — which
reaches for Celery, falls back to running the task inline against `SessionLocal`,
and is therefore sensitive to what the previous test in the session did to that
engine. Asserting on rows made them fail depending on file order, which is not a
property of the code under test.

```console
$ cd backend && python -m pytest tests/test_follow_idempotency.py -q
8 passed

$ python -m pytest app/tests/test_followers.py::test_prevent_duplicate_follow -q
1 passed          # PendingRollbackError on main
```

## Not fixed here

Two failures in files this touches, both pre-existing on `main` and both a
different question:

- `app/tests/test_followers.py::test_mutual_followers_service` asserts that
  `mutual_followers(a, b)` means "a and b follow each other". The service
  implements "people a and b both follow". The second reading is the one that is
  not already answered by `GET /followers/status/{id}`, but the endpoint is
  undocumented and unused by the frontend, so picking one is a decision, not a
  bug fix.
- `app/tests/test_notifications.py::test_follow_notification` gets past its
  first assertion now and fails on a later one, which asserts that
  `NotificationService.create_notification` deduplicates. Only
  `DatabaseChannel.send` does. That inconsistency is real and is its own change.

### DIFF
--- a/backend/app/routers/followers.py
+++ b/backend/app/routers/followers.py
@@ -10,11 +10,9 @@ from fastapi import APIRouter, Depends, HTTPException, status
 from sqlalchemy.orm import Session
 
 from app.dependencies import get_current_user, get_database, get_optional_current_user
-from app.models.notification import NotificationType
 from app.models.user import User
 from app.schemas.follower import FollowerResponse, FollowStatusResponse
 from app.services.follower_service import FollowerService
-from app.services.notification_service import NotificationService
 
 router = APIRouter(
     tags=["Followers"],
@@ -50,26 +48,14 @@ def follow_user(
             detail="Already following this user",
         )
 
-    follow = FollowerService.follow_user(
+    # The notification is the service's job and has been all along. This used
+    # to enqueue a second one here -- same event, different title, different
+    # link -- so a follow put two rows in the recipient's list (#1317).
+    return FollowerService.follow_user(
         db,
         current_user.id,
         user_id,
     )
-
-    try:
-        NotificationService.enqueue(
-            db,
-            recipient_id=user_id,
-            sender_id=current_user.id,
-            type=NotificationType.FOLLOW,
-            title="New follower",
-            message=f"{current_user.username} started following you.",
-            action_url=f"/users/{current_user.id}",
-        )
-    except Exception as e:
-        print(f"ENQUEUE ERROR: {e}")
-
-    return follow
 
 
 @router.delete(
--- a/backend/app/services/follower_service.py
+++ b/backend/app/services/follower_service.py
@@ -8,6 +8,9 @@ from sqlalchemy import and_, select, func
 # pyrefly: ignore [missing-import]
 from sqlalchemy.orm import Session
 
+# pyrefly: ignore [missing-import]
+from sqlalchemy.exc import IntegrityError
+
 from app.models.activity import ActivityType
 from app.models.follower import Follower
 from app.models.user import User
@@ -30,6 +33,23 @@ class FollowerService:
         follower_id: uuid.UUID,
         following_id: uuid.UUID,
     ) -> Follower:
+        """
+        Follow a user, once.
+
+        Idempotent: following someone you already follow returns the existing
+        relationship and does nothing else -- no second activity entry, no
+        second notification.
+
+        The insert runs inside a SAVEPOINT because a `SELECT` first is
+        check-then-act, and two concurrent requests from the same user both
+        pass it. Losing that race raises `IntegrityError` from the unique
+        constraint, and an unhandled one puts the whole `Session` into a
+        rolled-back state -- every later statement in the request, including
+        the response serialiser and the request-logging middleware, then fails
+        with `PendingRollbackError` instead. The savepoint confines the rollback
+        to the insert, so the outer transaction survives and the request can go
+        on to answer normally.
+        """
 
         if BlockService.is_blocked(db, follower_id, following_id):
             raise HTTPException(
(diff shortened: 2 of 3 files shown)


MOHITKOURAV01:

```
```

once that merges.

untouched by anything here.

nensii21:

PR #1323, opened 2026-08-24
fix(time): one clock for the announcement windows, one answer for expiry

### LINKED ISSUE #1318: Naive `datetime.utcnow()` is back in the announcement services, and recovery-token expiry compares unguarded
## Summary

`app/utils/time.py` exists to keep naive datetimes out of a codebase whose every timestamp column
is `DateTime(timezone=True)`. Its module docstring says why:

> Comparing them raises `TypeError: can't compare offset-naive and offset-aware datetimes`.
>
> Writing a naive value into a `TIMESTAMP WITH TIME ZONE` column makes the driver interpret it
> in the session time zone rather than UTC. On any deployment whose session is not UTC the
> stored instant is silently wrong.

`tests/test_utc_time.py` guards that invariant, and it is red on `main`:

```console
$ cd backend && python -m pytest tests/test_utc_time.py -q
FAILED tests/test_utc_time.py::TestNoNaiveUtcnowRemains::test_no_service_calls_datetime_utcnow
```

Two call sites got past it, both in the announcement services, both merged after the guard
existed:

```console
$ grep -rn "datetime.utcnow()" backend/app --include='*.py'
app/services/feature_announcement_service.py:33:        published_at = data.published_at or datetime.utcnow()
app/services/global_announcement_service.py:54:        now = datetime.utcnow()
```

Nothing turned red when they landed, because backend CI runs `pytest` with
`continue-on-error: true`.

## What each one does

**`global_announcement_service.py:54`** is the query that decides which announcements a user
sees:

```python
now = datetime.utcnow()
stmt = select(GlobalAnnouncement).where(
    GlobalAnnouncement.is_active == True,
    GlobalAnnouncement.start_date <= now,
    or_(
        GlobalAnnouncement.end_date.is_(None),
        GlobalAnnouncement.end_date >= now,
    ),
)
```

`start_date` and `end_date` are `TIMESTAMP WITH TIME ZONE`. The bound parameter is naive, so
Postgres reads it in the session `TimeZone` rather than as UTC. The announcement window is then
off by the session's offset — for a deployment running `Asia/Kolkata` that is 5½ hours, in the
direction that shows an announcement early and keeps showing it late. There is no error; the
window is just wrong.

**`feature_announcement_service.py:33`** writes rather than reads:

```python
published_at = data.published_at or datetime.utcnow()
```

`published_at` is `DateTime(timezone=True)`. An announcement created without an explicit
timestamp is stored at the session-local reading of a UTC wall clock, and sorts against
explicitly-published rows accordingly.

Both also emit `DeprecationWarning` on 3.12, which is the version the project targets.

## The same fault, one layer down

`app/utils/time.py` also provides `is_expired()`, written for exactly this:

> Normalises the input, so this is safe against rows written before the naive/aware mix was
> cleaned up — the comparison that used to raise `TypeError` now just works.

Nothing in `auth_service.py` uses it, and the file is inconsistent with itself about whether
that matters. Refresh tokens normalise by hand:

```python
expires_at = db_token.expires_at
if expires_at.tzinfo is None:
    expires_at = expires_at.replace(tzinfo=timezone.utc)
if expires_at < now:
```

Recovery tokens, forty lines further down, do not:

```python
if t_rec.expires_at and t_rec.expires_at < datetime.now(timezone.utc):   # :1110
if token_record.expires_at and token_record.expires_at < datetime.now(   # :1166
```

Three comparisons of the same kind of column, two of them unguarded. Where the stored value is
naive — a row written before the cleanup, or any backend that does not round-trip tzinfo — those
two raise `TypeError` inside the request and the caller gets a 500:

```console
$ cd backend && python -m pytest app/tests/test_forgot_password.py -q
FAILED app/tests/test_forgot_password.py::test_forgot_password_success
FAILED app/tests/test_forgot_password.py::test_verify_recovery_token_success
E  TypeError: can't compare offset-naive and offset-aware datetimes
E  app/services/auth_service.py:1110

ERROR devlink.request {'method': 'GET', 'path': '/api/auth/verify-recovery-token', 'status': 500}
```

A user clicking the link in their password-reset email gets a 500, not "this link has expired".

## Expected

- `tests/test_utc_time.py` passes, and stays passing.
- Announcement windows are compared against an aware UTC clock.
- Every expiry comparison in `auth_service` goes through one helper, so the next one added is
  guarded by construction rather than by whoever remembers.

MOHITKOURAV01:
Fixes #1318.

## The bug

`app/utils/time.py` exists so nothing naive gets near a `DateTime(timezone=True)`
column, and `tests/test_utc_time.py` has a test that says so. It was red:

```console
$ cd backend && python -m pytest tests/test_utc_time.py -q
FAILED tests/test_utc_time.py::TestNoNaiveUtcnowRemains::test_no_service_calls_datetime_utcnow
```

`datetime.utcnow()` had come back in two places, and nothing turned red when it
did, because backend CI runs `pytest` under `continue-on-error: true`.

**`global_announcement_service.py:54`** reads it as the bound parameter of the
window query:

```python
now = datetime.utcnow()
stmt = select(GlobalAnnouncement).where(
    GlobalAnnouncement.start_date <= now,
    or_(GlobalAnnouncement.end_date.is_(None), GlobalAnnouncement.end_date >= now),
)
```

`start_date` and `end_date` are `TIMESTAMP WITH TIME ZONE`. Postgres reads a
naive parameter in the session time zone, not as UTC, so the whole visibility
window slides by the deployment's offset — 5h30m on `Asia/Kolkata`, in the
direction that shows an announcement early and keeps showing it late. No error
anywhere; the window is just wrong.

**`feature_announcement_service.py:33`** writes it into `published_at`, so an
announcement created without an explicit timestamp is stored at the
session-local reading of a UTC wall clock, and sorts against explicitly
published rows accordingly.

## The other half

`auth_service` had three comparisons of the same kind of column, written three
different ways. Refresh tokens normalised by hand and were correct:

```python
expires_at = db_token.expires_at
if expires_at.tzinfo is None:
    expires_at = expires_at.replace(tzinfo=timezone.utc)
if expires_at < now:
```

Recovery tokens, forty lines down, did not:

```python
if t_rec.expires_at and t_rec.expires_at < datetime.now(timezone.utc):      # :1110
if token_record.expires_at and token_record.expires_at < datetime.now(...):  # :1166
```

Where the stored value comes back naive — a row written before the cleanup, or
any backend that does not round-trip `tzinfo` — those two raise inside the
request:

```console
$ cd backend && python -m pytest app/tests/test_forgot_password.py -q
E  TypeError: can't compare offset-naive and offset-aware datetimes
E  app/services/auth_service.py:1110
ERROR devlink.request {'path': '/api/auth/verify-recovery-token', 'status': 500}
```

A user clicking the link in their password-reset email gets a 500 instead of
"this link has expired".

## The fix

The two services use `utcnow()`.

All three expiry comparisons go through `is_expired()`, which was written for
exactly this and had no callers. It normalises first, so a legacy naive row
compares instead of raising. That is worth more than the two lines it removes:
one question, one answer, and the next expiry check somebody adds is correct by
construction rather than by remembering.

`app/tests/test_forgot_password.py` made the same mistake it was testing for —
`token_record.expires_at > datetime.now(timezone.utc)` on a value straight out
of the database, which raises rather than fails. Also on `is_expired` now, which
takes that file from four failures to one.

## Tests

`backend/tests/test_naive_timestamp_handling.py`, 18 cases.

The announcement window: live, not-yet-started, finished, no end date, and the
boundary at 1 / 30 / 90 / 329 minutes — the last one being 5h29m, inside the
offset that a naive clock would shift it by.

Both services also get a clock-substitution test, and that is the part that
actually pins the regression. Substituting `utcnow` and asserting the service
used it fails if anyone reaches for `datetime.utcnow()` again, whereas
inspecting `tzinfo` on the way back out proves nothing — SQLite does not
round-trip it, and it was the write that was wrong, not the read.

Expiry: a naive stored value past and future, an aware value in another zone, a
null expiry, and reset and refresh tokens read back through real columns.

```console
$ cd backend && python -m pytest tests/test_naive_timestamp_handling.py tests/test_utc_time.py -q
41 passed

$ grep -rn "datetime.utcnow()" app --include='*.py'
$ echo $?
1
```

`app/tests/test_forgot_password.py::test_reset_password_reused_password` still
fails. It is pre-existing and about password-validation ordering — the endpoint
rejects the new password for missing an uppercase letter before it gets to the
reuse check — which has nothing to do with this.

### DIFF
--- a/backend/app/services/auth_service.py
+++ b/backend/app/services/auth_service.py
@@ -44,6 +44,7 @@ from app.schemas.auth import (
     LoginRequest,
     RegisterRequest,
 )
+from app.utils.time import is_expired
 from app.utils.validators import (
     validate_email,
     validate_name,
@@ -826,10 +827,7 @@ class AuthService:
                 status_code=status.HTTP_401_UNAUTHORIZED,
                 detail="Invalid or expired refresh token.",
             )
-        expires_at = db_token.expires_at
-        if expires_at.tzinfo is None:
-            expires_at = expires_at.replace(tzinfo=timezone.utc)
-        if expires_at < now:
+        if is_expired(db_token.expires_at):
             raise HTTPException(
                 status_code=status.HTTP_401_UNAUTHORIZED,
                 detail="Invalid or expired refresh token.",
@@ -1107,7 +1105,7 @@ class AuthService:
                         "valid": False,
                         "message": "This recovery token has already been used.",
                     }
-                if t_rec.expires_at and t_rec.expires_at < datetime.now(timezone.utc):
+                if is_expired(t_rec.expires_at):
                     return {
                         "valid": False,
                         "message": "This recovery token has expired.",
@@ -1163,9 +1161,7 @@ class AuthService:
                         status_code=status.HTTP_400_BAD_REQUEST,
                         detail="This reset token has already been used.",
                     )
-                if token_record.expires_at and token_record.expires_at < datetime.now(
-                    timezone.utc
-                ):
+                if is_expired(token_record.expires_at):
                     raise HTTPException(
                         status_code=status.HTTP_400_BAD_REQUEST,
                         detail="This reset token has expired.",
--- a/backend/app/services/feature_announcement_service.py
+++ b/backend/app/services/feature_announcement_service.py
@@ -2,7 +2,6 @@ from __future__ import annotations
 
 import math
 import uuid
-from datetime import datetime
 from sqlalchemy import func, select, or_
 from sqlalchemy.orm import Session, selectinload
 
@@ -17,6 +16,7 @@ from app.schemas.feature_announcement import (
     FeatureAnnouncementResponse,
     FeatureAnnouncementUpdate,
 )
+from app.utils.time import utcnow
 
 
 class FeatureAnnouncementService:
@@ -30,7 +30,7 @@ class FeatureAnnouncementService:
         admin_id: uuid.UUID,
         data: FeatureAnnouncementCreate,
     ) -> FeatureAnnouncement:
-        published_at = data.published_at or datetime.utcnow()
+        published_at = data.published_at or utcnow()
         announcement = FeatureAnnouncement(
             created_by_id=admin_id,
             title=data.title,
--- a/backend/app/services/global_announcement_service.py
+++ b/backend/app/services/global_announcement_service.py
@@ -1,7 +1,6 @@
 from __future__ import annotations
 
 import uuid
-from datetime import datetime
 from sqlalchemy import select, or_
 from sqlalchemy.orm import Session
 
@@ -10,6 +9,7 @@ from app.schemas.global_announcement import (
     GlobalAnnouncementCreate,
     GlobalAnnouncementUpdate,
 )
+from app.utils.time import utcnow
 
 
 class GlobalAnnouncementService:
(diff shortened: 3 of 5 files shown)


MOHITKOURAV01:

```
```

once that merges.

untouched by anything here.

nensii21:

Issue #1196, opened 2026-08-16
Uploads are validated by client-supplied headers, and delete_file will remove any path the caller names


MOHITKOURAV01:
### ✦ Describe the Bug

`CloudStorageService` in `backend/app/core/storage.py` decides what an uploaded file *is* entirely from headers the client controls, and it will delete any path the caller names.

**Validation trusts the client.** `_validate_file` compares `file.content_type` — the `Content-Type` from the multipart part, which the uploader writes — against the allowlist. An attacker uploads `payload.svg` or an HTML file and simply declares `Content-Type: image/png`. Nothing ever looks at the bytes.

**The stored extension comes from the client too.**

```python
ext = os.path.splitext(file.filename)[1] if file.filename else ""
filename = f"{uuid.uuid4().hex}{ext}"
```

The comment above it says *"Generate unique secure filename to prevent traversal/collisions"*, and the UUID does handle collisions — but the extension is copied verbatim out of the user's filename. Upload `x.html` with `Content-Type: image/png` and the local provider writes `<uuid>.html` into `UPLOAD_DIR`, which is served back under `/static/uploads/`. Same origin, attacker-authored HTML: stored XSS against every session on the site. `.svg` gets there too, and an SVG is a script host.

**`delete_file` joins an unsanitised caller string onto a filesystem path.**

```python
def delete_file(self, object_name: str) -> bool:
    if self.provider == "local":
        file_path = os.path.join(settings.UPLOAD_DIR, object_name)
        if os.path.exists(file_path):
            os.remove(file_path)
```

`os.path.join` does not constrain anything. `object_name = "../../app/core/config.py"` resolves outside `UPLOAD_DIR` and is removed. Any endpoint that forwards a stored key from a request body to `delete_file` is an arbitrary-file-delete primitive. The `directory` argument to `upload_file` is joined the same way.

### ⟡ The malware hook only inspects the first few bytes

```python
suspicious_signatures = [b"<?php", b"<script", b"MZ", b"\x7fELF"]
for sig in suspicious_signatures:
    if contents.startswith(sig):
        raise ValueError(...)
```

`startswith` — so the check is defeated by putting a single byte in front, which every real polyglot does anyway. A valid PNG header followed by `<?php ... ?>` passes. Meanwhile the check produces false positives on legitimate content: a plain-text file that happens to begin with `MZ` is rejected.

[...]

PR #1324, opened 2026-08-24
fix(webhooks): sign the bytes we send, and only send them where we checked

### LINKED ISSUE #1319: Webhook delivery signs bytes it never sends, and posts to any URL a logged-in user asks for
## Summary

`WebhookService` has two independent faults, and each on its own defeats the point of the
feature.

1. The HMAC signature is computed over bytes that are never sent, so no receiver can verify it.
2. The delivery URL is never validated, and the response body is stored and readable — which
   turns `POST /api/webhooks/dispatch` into an SSRF read primitive available to any logged-in
   user.

## Fault 1 — the signature does not cover the request body

`generate_signature` (`app/services/webhook_service.py:41`) serialises the payload itself:

```python
payload_bytes = json.dumps(payload, sort_keys=True).encode("utf-8")
signature = hmac.new(secret.encode("utf-8"), payload_bytes, hashlib.sha256).hexdigest()
```

`_send_http_request` (`:94`) hands the *dict* to `httpx` and lets it serialise again:

```python
with httpx.Client(timeout=10.0) as client:
    return client.post(url, json=payload, headers=headers)
```

`httpx` uses `separators=(",", ":")`, `ensure_ascii=False`, and insertion order. `json.dumps`
defaults to `", "` / `": "`, `ensure_ascii=True`, and here `sort_keys=True`. Three differences,
any one of which is enough:

```console
$ python -c '
import json, httpx
p = {"b": 1, "a": "é", "c": [1, 2]}
print("signed:", json.dumps(p, sort_keys=True).encode())
print("sent  :", httpx.Request("POST", "http://x/", json=p).content)'
signed: b'{"a": "\\u00e9", "b": 1, "c": [1, 2]}'
sent  : b'{"b":1,"a":"\xc3\xa9","c":[1,2]}'
```

A receiver following the only sane verification procedure — HMAC the bytes you received, compare
to `X-DevLink-Signature` — never gets a match. The header is decoration. Any receiver that
currently "works" is either not checking it, or has reverse-engineered our `json.dumps` call,
which then breaks the moment the payload gains a non-ASCII character.

The signature is also never reachable through the API: `WebhookService.dispatch_webhook` takes
`secret`, and the router (`app/routers/webhooks.py:37`) does not pass it.

## Fault 2 — arbitrary destination, and the response comes back

`_send_http_request` posts wherever it is told. `app/utils/url_safety.py` exists and does exactly
the job — scheme and port allowlists, credential rejection, resolution of every address behind
the host, a public-address check, and a pinned connection so DNS cannot answer differently
between the check and the connect. `link_preview_service` routes every fetch through it. Webhook
delivery does not.

The destination is caller-controlled and the endpoint is open to any authenticated user:

```python
@router.post("/dispatch", ...)
def dispatch_webhook(
    params: WebhookDispatchParams,
    current_user: User = Depends(get_current_user),   # any logged-in user
    db: Session = Depends(get_database),
):
    return WebhookService.dispatch_webhook(
        db=db, event_type=params.event_type, target_url=params.target_url, ...
    )
```

What makes this more than a blind SSRF is that `_execute_delivery` keeps the answer:

```python
status_code = response.status_code
resp_text = response.text[:2000] if response.text else ""
...
delivery.response_status_code = status_code
delivery.response_body = resp_text
```

and `GET /api/webhooks/deliveries` hands it straight back, also to any authenticated user.

```console
$ curl -XPOST -H "Authorization: Bearer $TOKEN" localhost:8000/api/webhooks/dispatch \
    -d '{"event_type":"x","target_url":"http://169.254.169.254/latest/meta-data/","payload":{}}'
$ curl -H "Authorization: Bearer $TOKEN" localhost:8000/api/webhooks/deliveries \
    | jq '.items[0] | {response_status_code, response_body}'
```

Two kilobytes of any internal HTTP endpoint the pods can reach — cloud metadata, an unauthenticated
admin port, a Redis HTTP proxy — read back through a normal API call. The status code alone is a
port scanner: `Network/HTTP Exception: ...` for a closed port versus an HTTP status for an open
one, recorded in `error_message`.

Retries make it worse rather than better. `process_pending_retries` re-sends failed deliveries on
a backoff for `max_retries` attempts, so one request buys several probes over time.

## Expected

- The bytes that are signed are the bytes that are sent.
- `target_url` goes through `validate_outbound_url` before we connect, at dispatch and on every
  retry, and is rejected with a 4xx the caller can read rather than being attempted and logged.
- A delivery that never happened because the destination was refused is recorded as such.

MOHITKOURAV01:
Fixes #1319.

Two faults, independent, each enough on its own to make the feature not do what
it says.

## Fault 1 — the signature described a payload nobody received

`generate_signature` serialised the payload itself; `_send_http_request` handed
the same dict to `httpx` and let it serialise again. `httpx` uses
`separators=(",", ":")`, `ensure_ascii=False` and insertion order; `json.dumps`
defaults to `", "` / `": "`, `ensure_ascii=True`, and here `sort_keys=True`.

```console
$ python -c '
import json, httpx
p = {"b": 1, "a": "é", "c": [1, 2]}
print("signed:", json.dumps(p, sort_keys=True).encode())
print("sent  :", httpx.Request("POST", "http://x/", json=p).content)'
signed: b'{"a": "\\u00e9", "b": 1, "c": [1, 2]}'
sent  : b'{"b":1,"a":"\xc3\xa9","c":[1,2]}'
```

A receiver's only verification procedure is: HMAC the bytes you received,
compare to `X-DevLink-Signature`. It could never match. The header was
decoration, and any receiver that appeared to work had either stopped checking
or reverse-engineered our `json.dumps` call — which then breaks the first time a
payload contains a non-ASCII character.

There is one serialiser now, `canonical_body`, and both sides use it.
`sort_keys` stays so a retry rebuilt from JSONB signs identically regardless of
dict ordering, and the separators match what the receiver actually gets.

## Fault 2 — arbitrary destination, with the answer returned

`_send_http_request` posted wherever it was told. `app/utils/url_safety.py`
exists and does exactly this job — scheme and port allowlists, credential
rejection, resolution of every address behind the host, a public-address check,
and a pinned connection so DNS cannot answer differently between check and
connect. `link_preview_service` has routed every fetch through it since #1196.
Webhook delivery never did.

What makes it more than a blind SSRF is that `_execute_delivery` keeps the
answer — `response.text[:2000]` and the status code — and
`GET /api/webhooks/deliveries` hands both back. Both endpoints are open to any
authenticated user:

```console
$ curl -XPOST -H "Authorization: Bearer $TOKEN" localhost:8000/api/webhooks/dispatch \
    -d '{"event_type":"x","target_url":"http://169.254.169.254/latest/meta-data/","payload":{}}'
$ curl -H "Authorization: Bearer $TOKEN" localhost:8000/api/webhooks/deliveries \
    | jq '.items[0].response_body'
```

Two kilobytes of any internal HTTP endpoint the pods can reach. The failure
side is a port scanner too — `Network/HTTP Exception: ...` for a closed port
versus a real status for an open one, recorded in `error_message` — and
`process_pending_retries` re-sends on a backoff, so one request buys several
probes.

### Where the check went

`_execute_delivery`, because it is the one place every path reaches: dispatch,
`process_pending_retries`, `replay_dlq_entry` and `replay_all_dlq_entries`. A
row already in the table — written before this check, or pointing at a DNS
answer that has since moved — is rechecked rather than trusted.

Before the attempt counter moves, since nothing was attempted. The connection is
then pinned to the address validation approved, the way link previews do it, and
redirects are not followed: a 302 to `169.254.169.254` is the whole attack, and
a webhook receiver that wants us somewhere else can say so in its configuration.

A refused destination is **Exhausted**, not Failed. Retrying changes nothing,
and the backoff schedule would turn one rejected request into several probes
spread over an hour. `response_body` and `response_status_code` stay `None` —
the point of refusing is that we learned nothing about the destination worth
handing back.

`dispatch_webhook` also checks up front, so the caller gets a 400 naming the
rule it broke rather than a stored delivery that quietly failed.

## `validate_target`

A class attribute rather than a direct call, so the resolver can be substituted.
Validation resolves the hostname, and a unit test that depends on DNS is a unit
test that fails on a train — `tests/test_webhook_retry_dlq.py` posts to
`api.example.com`, which does not resolve. Substituting the *resolver* rather
than the validator keeps every other rule live, so a test that points somewhere
it should not still fails.

## Tests

`backend/tests/test_webhook_signature_and_targets.py`, 20 cases.

Signature: it verifies the way a receiver would; the body `httpx` actually puts
on the wire is the body that was signed; key order does not change it; payload
and secret changes do. Plus one that pins the bug itself — asserting that
letting `httpx` serialise produces different bytes than the old signing did — so
nobody puts `json=payload` back assuming it is equivalent.

Destinations: dispatch refuses link-local, loopback, `localhost`, `file://`,
`ftp://`, credentials in the URL, and a non-allowlisted port, storing nothing;
the API answers 400; a retry revalidates and never reaches the network; a
refusal returns nothing about the destination, is not retried, and lands in the
DLQ; and an allowed destination still goes through with the canonical body and
the right `Content-Type`.

```console
$ cd backend && python -m pytest tests/test_webhook_signature_and_targets.py tests/test_webhook_retry_dlq.py -q
26 passed
```

## Not done here

The router still does not pass `secret`, so signing remains unreachable through
the API. Accepting a shared secret over the wire is a decision with its own
questions — where it is stored, whether it is per-endpoint — and belongs with
whoever wants the feature, not in a fix for the signature being wrong.

### DIFF
--- a/backend/app/routers/webhooks.py
+++ b/backend/app/routers/webhooks.py
@@ -15,6 +15,7 @@ from app.schemas.webhook import (
     WebhookMetricsResponse,
 )
 from app.services.webhook_service import WebhookService
+from app.utils.url_safety import UnsafeURL
 
 router = APIRouter(
     prefix="/webhooks",
@@ -34,14 +35,23 @@ def dispatch_webhook(
     db: Session = Depends(get_database),
 ):
     """Dispatch a webhook to a target URL with automatic retries and dead letter queue fallback."""
-    return WebhookService.dispatch_webhook(
-        db=db,
-        event_type=params.event_type,
-        target_url=params.target_url,
-        payload=params.payload,
-        headers=params.headers,
-        max_retries=params.max_retries,
-    )
+    try:
+        return WebhookService.dispatch_webhook(
+            db=db,
+            event_type=params.event_type,
+            target_url=params.target_url,
+            payload=params.payload,
+            headers=params.headers,
+            max_retries=params.max_retries,
+        )
+    except UnsafeURL as exc:
+        # 400, not 500: the caller sent a destination we will not connect to,
+        # and the message from `url_safety` says which rule it broke without
+        # echoing back what the host resolved to.
+        raise HTTPException(
+            status_code=status.HTTP_400_BAD_REQUEST,
+            detail=str(exc),
+        ) from exc
 
 
 @router.post(
--- a/backend/app/services/webhook_service.py
+++ b/backend/app/services/webhook_service.py
@@ -16,6 +16,12 @@ from app.models.webhook import (
     WebhookDeadLetterQueue,
     WebhookDeliveryStatus,
 )
+from app.utils.url_safety import (
+    SafeTarget,
+    UnsafeURL,
+    pin_target,
+    validate_outbound_url,
+)
 
 logger = structlog.get_logger("devlink.webhooks")
 
@@ -39,18 +45,43 @@ def calculate_backoff_delay(
 
 
 class WebhookService:
+    #: How a destination is checked before we connect to it.
+    #:
+    #: A class attribute rather than a direct call so it can be substituted:
+    #: validation resolves the hostname, and a unit test that depends on DNS is
+    #: a unit test that fails on a train.
+    validate_target = staticmethod(validate_outbound_url)
+
+    @classmethod
+    def canonical_body(cls, payload: Dict[str, Any] | str) -> bytes:
+        """
+        The bytes we sign, which are the bytes we send.
+
+        There is only one function here on purpose. The signature used to be
+        computed over ``json.dumps(payload, sort_keys=True)`` while the request
+        body came from ``httpx``'s own serialisation of the same dict -- which
+        uses ``separators=(",", ":")``, ``ensure_ascii=False`` and insertion
+        order. Three differences, any one of which is enough to make the
+        signature un-verifiable by the only sane procedure a receiver has: HMAC
+        the bytes you received and compare.
+
+        ``sort_keys`` is kept so the same payload signs identically on a retry
+        regardless of dict ordering, and the separators match what a receiver
+        gets so nothing has to reverse-engineer our serialiser.
+        """
+        if isinstance(payload, (dict, list)):
+            return json.dumps(
+                payload, sort_keys=True, separators=(",", ":"), ensure_ascii=False
+            ).encode("utf-8")
+        return str(payload).encode("utf-8")
+
     @classmethod
     def generate_signature(cls, payload: Dict[str, Any] | str, secret: str) -> str:
         """
-        Generates an HMAC-SHA256 signature for webhook payload verification.
+        Generates an HMAC-SHA256 signature over the exact request body.
         """
-        if isinstance(payload, dict):
-            payload_bytes = json.dumps(payload, sort_keys=True).encode("utf-8")
-        else:
-            payload_bytes = str(payload).encode("utf-8")
-
         signature = hmac.new(
-            secret.encode("utf-8"), payload_bytes, hashlib.sha256
+            secret.encode("utf-8"), cls.canonical_body(payload), hashlib.sha256
         ).hexdigest()
         return f"sha256={signature}"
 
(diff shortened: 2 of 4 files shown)


MOHITKOURAV01:

```
```

once that merges.

untouched by anything here.

nensii21:

Issue #1347, opened 2026-08-25
Avatar and banner upload crash on mount: `ImageCropUploadModal` resets four setters it does not have


MOHITKOURAV01:
## Summary

`ImageCropUploadModal` has a reset effect that calls four setters the component does not have.
The effect runs on mount, so the component throws the first time it is rendered — which is
before anyone opens it, because it is rendered closed.

Avatar and banner uploads are unusable.

## The code

`frontend/src/components/shared/ImageCropUploadModal.tsx:57`:

```tsx
// Reset state when modal opens/closes
useEffect(() => {
  if (!isOpen) {
    setSelectedFile(null);
    setPreviewUrl(null);
    setError(null);
    setZoom(1.0);
    setRotation(0);
    setPanX(0);
    setPanY(0);
    setIsCameraActive(false);
    setIsUploading(false);
    setUploadProgress(0);
    setUploadComplete(false);
  }
}, [isOpen]);
```

The component declares eight pieces of state (`:39`–`:48`): `selectedFile`, `previewUrl`,
`isDragging`, `error`, `isCameraActive`, `isUploading`, `uploadProgress`, `uploadComplete`.

`setZoom`, `setRotation`, `setPanX` and `setPanY` are not among them and are not imported. They
belong to `ImageCropper`, which owns that state itself (`ImageCropper.tsx:83`) and is a child
here, reached through a ref:

```tsx
const cropperRef = useRef<ImageCropperHandle | null>(null);
```

So the crop state was never this component's to reset. Four lines were written against the wrong
component.

## Reproducing

```console
$ cd frontend && npx vitest run src/components/__tests__/image-upload-experience.test.tsx

FAIL  UserAvatar with editable crop uploader > renders hover camera icon overlay when editable
FAIL  UserAvatar with editable crop uploader > opens ImageCropUploadModal when editable avatar is clicked
ReferenceError: setZoom is not defined
```

Both tests, including the one that only renders the avatar and never opens the modal. The
guard is `if (!isOpen)`, and on mount `isOpen` is `false`, so the branch that throws is the
branch that runs first.

`setZoom` is the one reported because it is first; the other three are equally undefined.

## Why it is not a type error

`npx tsc --noEmit` does not currently reach this file. The parse failures described in the
conflict-marker issue stop the check earlier, so the two failing tests are the only signal — and
nothing in CI runs the frontend tests today (#1316).

## Blast radius

[...]

Issue #1345, opened 2026-08-25
Three tracked files are UTF-16: `pip install -r requirements.txt` fails, and the recruiter module lost every template literal


MOHITKOURAV01:
## Summary

Three tracked text files are encoded UTF-16LE. One of them is `backend/requirements.txt`, and it
is the reason `Backend CI` has been failing: `pip install -r requirements.txt` rejects the file
outright, so nothing downstream of it has run.

The other two are the recruiter feature. Beyond the encoding, the tool that wrote them also ate
every backtick and every `${...}`, so `recruiter.ts` would still be wrong after a straight
re-encode.

Git treats all three as binary, which is why no text-based check has flagged them.

## The files

```console
$ for f in $(git ls-files); do
    enc=$(file -b --mime-encoding "$f")
    case "$enc" in utf-8|us-ascii|binary) ;; *) echo "$enc  $f";; esac
  done
utf-16le  backend/requirements.txt
utf-16le  frontend/src/api/modules/recruiter.ts
utf-16le  frontend/src/routes/_app.recruiter.tsx
```

`requirements.txt` is UTF-8 for its first 144 lines and switches encoding for the last one:

```console
$ tail -c 32 backend/requirements.txt | hexdump -C
00000000  73 00 74 00 72 00 69 00  70 00 65 00 3d 00 3d 00  |s.t.r.i.p.e.=.=.|
00000010  32 00 2e 00 34 00 30 00  2e 00 30 00 0d 00 0a 00  |2...4.0...0.....|
```

`stripe==2.40.0` appended by something that writes UTF-16 — a PowerShell `>>` redirect does
exactly this.

## Consequence 1 — the backend cannot be installed

```console
$ pip install -r backend/requirements.txt
ERROR: Invalid requirement: 'r\x00i\x00p\x00e\x00=\x00=\x002\x00.\x004\x000\x00.\x000\x00':
Expected semicolon (after name with no version specifier) or end
    r i p e = = 2 . 4 0 . 0
     ^ (from line 145 of backend/requirements.txt)
```

pip stops there. It does not install `stripe`, and it does not install anything else it had not
already reached.

This is what `Backend CI` has been dying of. Run `32811676648` on `7ff860b6` failed in 17
seconds — the `Install dependencies` step, long before `Run tests`.

It is not only CI. `app/main.py:667` imports the donations router unconditionally, and
`app/services/donation_service.py:2` imports `stripe` at module scope, so on a machine that
followed the README the API does not start:

```console
[...]

PR #1095, opened 2026-08-12
feat(orgs): add organization member roles and permissions (#938)

MILAN-123865:
```md
# Pull Request

## Summary

Provide a brief description of the changes made in this pull request.

---

## Related Issue

Closes #

or

Fixes #

---

## Type of Change

Please check the relevant option(s):

- [ ] Bug fix
- [ ] New feature
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Refactoring
- [ ] UI/UX enhancement
- [ ] Tests
- [ ] Other

---

## Changes Made

Describe the key changes included in this pull request.

-
-
-

---

## Screenshots (if applicable)

Add screenshots or screen recordings demonstrating your changes.

---

## Testing

Describe how you tested your changes.

- [ ] Tested locally
- [ ] Existing tests pass
- [ ] Added new tests
- [ ] UI verified
- [ ] Cross-browser tested (if applicable)

Additional testing notes:

---

## Checklist

Before submitting this pull request, confirm the following:

- [ ] My code follows the project's coding standards.
- [ ] I have tested my changes locally.
- [ ] I have updated the documentation where necessary.
- [ ] My changes do not introduce new warnings or errors.
- [ ] I have reviewed my own code.
- [ ] This pull request focuses on a single feature or fix.
- [ ] I have linked the related issue.

---

## Additional Notes

Include any additional information that reviewers should know.
```
## Summary

Improved the application's visual hierarchy to make primary information
easier to identify and improve overall readability.

### Changes

- Improved heading hierarchy
- Added consistent typography and spacing
- Muted secondary metadata and helper text
- Strengthened primary action emphasis
- Improved section separation
- Standardized alignment and component spacing
- Reused shared UI styles where possible
- Preserved existing functionality and responsive behavior

### Testing

- `npm run lint`
- `npm run build`
- `npm test`
- Verified representative pages on desktop and mobile

Closes #938

PR #1351, opened 2026-08-25
fix(frontend): resolve the #1095 merge instead of keeping both sides

### LINKED ISSUE #1344: `main` does not build: PR #1095 was merged with both sides of every conflict kept, in 19 files
## Summary

`main` does not build. The merge of #1095 (`7ff860b6`) resolved its conflicts by deleting the
`<<<<<<<`, `=======` and `>>>>>>>` characters and keeping **both sides of every hunk**, leaving
the branch-name lines behind as stray statements. 19 tracked files are affected, 5 of them no
longer parse at all.

The `Conflict Markers` workflow passed on that push, because the thing it looks for is exactly
the thing that was removed.

## Reproducing

```console
$ git checkout main && cd frontend && npx vite build
Build failed with 1 error:
[plugin tanstack-start:route-tree-client-plugin]
Error: Crawling result not available
```

`routeTree.gen.ts` is one of the damaged files, so the route crawler gives up before anything
else is reported. Type checking sees the rest:

```console
$ npx tsc --noEmit | wc -l
289
```

and the test run collapses in the same place:

```console
$ npx vitest run --exclude "e2e/**"
Test Files  10 failed | 49 passed (59)
```

Eight of those ten are not broken themselves — they transitively import `src/api/modules/posts.ts`
and inherit its parse error:

```
[PARSE_ERROR] Expected `,` or `}` but found `/`
  ╭─[ src/api/modules/posts.ts:7:6 ]
```

## What the damage looks like

`src/api/modules/posts.ts:5`, both sides kept, branch names left as bare expression statements:

```ts
export const postsApi = {
  list: (query?: { page?: number; limit?: number }) => api.get<Flare[]>("/api/posts", { query }),
 feat/organization-roles-987-v2
  drafts: () => api.get<Flare[]>("/api/posts/drafts"),
  create: async (body: {
    content: string;
    image?: string;
    tags?: string[];
    status?: string;
    publish_at?: string;
  }) => {

  drafts: (query?: { page?: number; limit?: number }) =>
    api.get<Flare[]>("/api/posts/drafts", { query }),
  create: async (body: { content: string; image?: string; tags?: string[] }) => {
 main
    const spamCheck = analyzeSpam(body.content);
```

`src/api/modules/projectTemplates.ts:97` — the same shape, and here the two sides differ only by
an `await`, so the resolution is unambiguous:

```ts
 feat/organization-roles-987-v2
    return api.post<{ success: boolean; is_favorited: boolean; stars_count: number }>(

    return await api.post<{ success: boolean; is_favorited: boolean; stars_count: number }>(
 main
      `/api/templates/${templateId}/favorite`,
    );
```

`src/api/modules/projectCollaborationMetrics.ts:68` lost its function signature entirely — the
body was closed off and the signature left dangling after it:

```ts
  };
};
): Promise<ProjectCollaborationMetricsResponse> =>
  api.get<ProjectCollaborationMetricsResponse>(`/projects/${projectId}/collaboration-metrics`);
```

## Full list

Five no longer parse:

- `frontend/src/routeTree.gen.ts` (218 of the 289 type errors)
- `frontend/src/api/modules/posts.ts`
- `frontend/src/api/modules/projectTemplates.ts`
- `frontend/src/components/bookmarks/CollectionSidebar.tsx`
- `frontend/src/routes/_app.dashboard.tsx`

Fourteen parse, and are wrong — duplicated JSX, unreachable branches, or a bare identifier
evaluated as a statement:

- `frontend/src/app/applications/page.tsx`
- `frontend/src/features/organizations/components/OrganizationMembers.tsx`
- `frontend/src/components/activity/ActivityFeed.tsx`
- `frontend/src/components/layout/RightPanel.tsx`
- `frontend/src/components/audit/AuditTimeline.tsx`
- `frontend/src/components/command-palette/CommandPalette.tsx`
- `frontend/src/components/profile/ExperienceCard.tsx`
- `frontend/src/components/profile/BioCard.tsx`
- `frontend/src/components/profile/EducationCard.tsx`
- `frontend/src/components/hackathons/TeamsTab.tsx`
- `frontend/src/components/hackathons/LeaderboardTab.tsx`
- `frontend/src/components/hackathons/RegisterDialog.tsx`
- `frontend/src/components/applications/ApplicationsList.tsx`
- `frontend/src/components/errors/SectionErrorBoundary.tsx`

`RightPanel.tsx` is the one with a visible symptom in the existing suite:

```
FAIL src/components/__tests__/workspace-status.test.tsx
ReferenceError: TypoSection is not defined
```

`TypoSection` is on a line that used to be inside a conflict hunk and is now a statement on its
own.

## Why the guard did not fire

`.github/workflows/conflict-markers.yml` greps for the markers themselves:

```
git grep -nIE '^(<{7}|={7}|>{7})( |$)'
```

Every one of those characters was stripped. What is left — ` feat/organization-roles-987-v2`
and ` main` on lines of their own — matches nothing the workflow looks for, so the check went
green on `7ff860b6` while the branch it was checking did not build.

The workflow header says this failure mode "has happened twice, and both times it left main
unbuildable on both the backend and the frontend for several days". This is the third time, and
the guard added in response to the first two cannot see it.

## Suggested fix

1. Repair all 19 files, taking the correct side of each hunk (`#1095` is the organization-roles
   work, so its side wins where the two genuinely differ; several hunks are identical apart from
   formatting).
2. Regenerate `routeTree.gen.ts` rather than hand-editing it.
3. Extend the guard to the residue: a line consisting only of a branch-like token immediately
   after a hunk boundary is not something anyone writes on purpose. Checking that the frontend
   still parses would also have caught this, and catches the next variation too.

Related: #1233 is the previous occurrence of this failure mode. Its file
(`features/dashboard/sections.tsx`) is clean now; these 19 are a fresh instance from a different
merge.

MOHITKOURAV01:
Closes #1344.

## What was wrong

The merge of #1095 (`7ff860b6`) deleted the conflict markers and kept **both
sides of every hunk** — 19 files, 29 hunks. `main` has not built since.

`Conflict Markers` went green on that push, because the characters it greps for
were the characters that had been removed.

## Resolving it

Each hunk is decided on its merits rather than by taking one side wholesale.
Most are the typography migration — raw Tailwind on `main`, `TypoCaption` /
`TypoSection` / `TypoHeading` on the branch — and take the branch. Two do not:

| file | resolution | why |
| --- | --- | --- |
| `_app.dashboard.tsx` | `main` | The branch predates the pinned-widgets section and its side would have deleted it. |
| `posts.ts` (first hunk) | **neither** | `main` added pagination to `drafts`, the branch added `status`/`publish_at` to `create`. Both shipped, so the resolution is the union. |

`routeTree.gen.ts` is regenerated rather than hand-edited, which also drops the
duplicated import block an earlier merge had left in it.

## What the parse errors were hiding

Three files were damaged the same way by earlier merges and only became visible
once the build got far enough to reach them:

- **`OrganizationHeader.tsx`** had two copies of the social-links block. The
  newer one routes every `href` through `sanitizeUrl`; the older one does not,
  and the interleaving had stranded the guard on a line of its own where it did
  nothing. Keeps the sanitising version.
- **`_app.settings.tsx`** had three generations of `tabs` stacked, so the
  sidebar rendered Appearance, Notifications and Security three times each,
  plus a second `privacy` branch truncated mid-element and a `security`
  one-liner with a `nancy@example.com` placeholder.
- **`_app.profile.$username.tsx`** had `PortfolioExportDialog` swallowing
  `EditProfileModal`, an unterminated `ManageSkillsModal`, and `avatarUrl`
  declared twice.

`projectCollaborationMetrics.ts` had silently reverted #1249: the doc comment
says errors propagate, and underneath it the mock fallback was back — a
collaboration score of 92 and a chart hardcoded to the week of 2026-08-04 —
with the real implementation dangling after the closing brace.

## The build does not stop at merge damage

- `donations.ts` default-imports `../client`, which has no default export.
- `@heroicons/react` and `@headlessui/react` are neither declared nor installed.
  Both were used in the donations UI only — one icon and one modal — so they
  move to `lucide-react` and the Radix dialog the rest of the app uses, rather
  than adding two dependencies for that.
- `zustand` is imported by `useSavedSearches.ts` and resolves only by hoisting.

13 files used `Typo*` components without importing them, which is what
`TypoSection is not defined` was at runtime. `RightPanel`'s workspace-status
card is also gone: `WorkspaceStatusIndicator` replaced it in `TopNavbar` and
`workspace-status.test.tsx` asserts it is not there.

## Results

```console
$ npx tsc --noEmit | wc -l
       0          # was 289

$ npx vite build
✓ 4332 modules transformed.
✓ built in 3.72s

$ npx vitest run --exclude "e2e/**"
 Test Files  58 passed | 1 skipped (59)
      Tests  772 passed | 2 skipped (774)
```

The build and the last four test failures need the two companion fixes, both of
which are small and independent: the UTF-16 recruiter files (#1345) and the
`ImageCropUploadModal` setters (#1347). On this branch alone `tsc` is clean and
the suite is 767 passed / 4 failed, with those two accounting for the four.

## The guard

`Conflict Markers` grows a second check: a marker label alone on a line with the
angle brackets stripped — ` feat/organization-roles-987-v2`, ` main`. That is
not something a formatter emits or a language parses. **62 such lines on the
commit that broke `main`, none on this branch**, and no false positives across
the tree.

## Note on `settings-tabs.test.tsx`

Skipped, not deleted. It describes the #947 refactor — an H1 of "Settings",
Profile/Billing/Developer tabs — and `settings-574.test.tsx` asserts a different
H1 and tab set for the same component. The two contradict each other and only
one can pass; the page implements #574. The skip carries a comment saying so, so
whoever picks #947 back up still has the acceptance criteria.


nensii21:

Issue #1348, opened 2026-08-25
Build output, tool dumps and a both-sides conflict "resolver" are tracked in the repository


MOHITKOURAV01:
## Summary

About 1.1 MB of build output, tool dumps and scratch files are tracked. Among them is
`resolve_conflicts.py`, a script whose stated job is to resolve merge conflicts by deleting the
markers and keeping **both sides** — which is the exact shape of the damage currently sitting in
19 files on `main`.

`.gitignore` already names `dist/`. That does nothing, because ignore rules do not apply to files
Git already tracks — the failure mode the `Tracked Files` workflow was written for, recurring
one directory over.

## What is tracked

```console
$ git ls-tree -r -l origin/main | ...
    362945  bandit_report.json
    423950  dist/assets/index-DbZQCFkd.js
     87520  dist/assets/index-BGuoSTMb.css
      8638  dist/lodo-dev.jpeg
       823  dist/index.html
     70720  frontend/tsc_errors_2.txt
     70514  ms_patch.diff
     63442  google_patch.diff
     31336  upstream_sections.tsx
     14969  old_sections.tsx
      2143  resolve_conflicts.py
       147  .vite/deps/_metadata.json
        23  .vite/deps/package.json
         0  frontend/tsc_errors.txt
         0  pr_info.json
```

By kind:

- **Build output** — `dist/` (520 KB of compiled JS/CSS from some commit nobody can identify) and
  `.vite/deps/`, a bundler cache.
- **Tool dumps** — `bandit_report.json` (363 KB), `frontend/tsc_errors_2.txt` (70 KB of compiler
  output), `frontend/tsc_errors.txt` and `pr_info.json`, both empty.
- **Patch files** — `google_patch.diff` and `ms_patch.diff`, 134 KB of diffs against unknown bases.
- **Scratch copies of a component** — `old_sections.tsx` and `upstream_sections.tsx`, which are
  hand-kept copies of `features/dashboard/sections.tsx` from the #1233 cleanup.
  `upstream_sections.tsx` is also UTF-16LE.
- **A codemod** — `resolve_conflicts.py`.

## The codemod

```python
def resolve_both(filepath):
    ...
    for line in lines:
        if line.startswith('<<<<<<< HEAD'):
            in_conflict = True
        elif line.startswith('======='):
            pass
        elif line.startswith('>>>>>>> upstream/main'):
            in_conflict = False
        else:
            out.append(line)
```

[...]

PR #1352, opened 2026-08-25
fix(encoding): make every tracked text file UTF-8, and check that it stays that way

### LINKED ISSUE #1345: Three tracked files are UTF-16: `pip install -r requirements.txt` fails, and the recruiter module lost every template literal

MOHITKOURAV01:
Closes #1345.

## Why Backend CI is red

`backend/requirements.txt` is UTF-8 for 144 lines and UTF-16LE for the last one:

```console
$ tail -c 32 backend/requirements.txt | hexdump -C
00000000  73 00 74 00 72 00 69 00  70 00 65 00 3d 00 3d 00  |s.t.r.i.p.e.=.=.|
00000010  32 00 2e 00 34 00 30 00  2e 00 30 00 0d 00 0a 00  |2...4.0...0.....|
```

pip does not read past it:

```console
$ pip install -r backend/requirements.txt
ERROR: Invalid requirement: 'r\x00i\x00p\x00e\x00=\x00=\x002\x00.\x004\x000\x00.\x000\x00'
    r i p e = = 2 . 4 0 . 0
     ^ (from line 145 of backend/requirements.txt)
```

It installs nothing after that point. Run `32811676648` on `7ff860b6` failed in
**17 seconds**, in `Install dependencies`, having never reached a test.

Nor is it only CI. `app/main.py:667` imports the donations router
unconditionally and `app/services/donation_service.py:2` imports `stripe` at
module scope, so a machine that followed the README could not start the API
either.

## Pinned, not floated

Restored as UTF-8 under a Payments heading, and pinned. `donation_service.py`
catches `stripe.error.SignatureVerificationError`, and `stripe.error` was
removed in stripe-python 5 — anything newer imports fine and then raises
`AttributeError` on the first bad webhook signature, which is the one path where
being wrong is expensive. The reason is in a comment next to the pin.

```console
$ pip install --dry-run -r backend/requirements.txt
Would install ... stripe-2.40.0 ...
```

## The recruiter module needed more than re-encoding

Whatever wrote it also ate every backtick and every interpolation:

```ts
api.get<ApplicationResponse[]>(\/api/applications/project/\\)
```

That is `` `/api/applications/project/${projectId}` `` with the backticks turned
into `\` and `${projectId}` deleted — PowerShell double-quote rules, the same
tool that produced the encoding. All four URLs are written back against the
routes in `app/routers/applications.py`:

| method | path | line |
| --- | --- | --- |
| `GET` | `/project/{project_id}` | `:108` |
| `PATCH` | `/{application_id}/shortlist` | `:257` |
| `PATCH` | `/{application_id}/schedule_interview` | `:290` |
| `PATCH` | `/{application_id}/notes` | `:323` |

`_app.recruiter.tsx` decodes to correct source and only needed the encoding.
`recruiter.ts` goes from 20 `tsc` errors to 0.

The other five UTF-16 files are tool output with no consumer — two compiler-error
dumps, two patch dumps and a scratch copy of a component, 307 KB between them.
Re-encoding meaningless output would be silly, so they are removed. The rest of
the tracked detritus is #1348 and left alone here.

## Why nothing caught it

Every guard in `.github/workflows` is built on `git grep`, and Git classifies a
file full of NUL bytes as binary and skips it. The definition of `recruiterApi`
is invisible to the tool:

```console
$ git grep -l 'recruiterApi' -- frontend/src
frontend/src/components/recruiter/RecruiterDashboard.tsx      # only the consumer
```

So `check_encoding.py` reads the bytes rather than grepping them. It names the
encoding it found and says where UTF-16 comes from, because it is always the
same place.

```console
$ python3 .github/scripts/check_encoding.py
2316 tracked files checked; all text files are UTF-8.
```

On `main` it reports all eight.


MOHITKOURAV01:

nensii21:

PR #1353, opened 2026-08-25
fix(users): stop the availability relationship shadowing the availability column

### LINKED ISSUE #1346: `POST /api/auth/register` returns 500: `UserResponse.availability` is a list, the ORM attribute is a scalar
## Summary

`UserResponse.availability` is declared as a list. The ORM attribute of the same name is a
scalar `uselist=False` relationship, so for any user without a row in `user_availability` it is
`None`. FastAPI validates the response, `None` is not a list, and the request becomes a 500.

`POST /api/auth/register` is one of the endpoints affected. Every account created through the
API is created and then reported as a server error.

## The collision

`app/schemas/user.py:120`, in `UserResponse`:

```python
availability: list[AvailabilitySlot] = Field(default_factory=list)
```

`AvailabilitySlot` (`:26`) is a day / start_time / end_time triple — a weekly schedule, added
with the resume-parsing work in `11113e85`.

`app/models/user.py:441`:

```python
availability: Mapped["UserAvailability"] = relationship(
    "UserAvailability",
    back_populates="user",
    uselist=False,
    cascade="all, delete-orphan",
)
```

`UserAvailability` (`app/models/user_availability.py:9`) is a different thing entirely — one row
per user, `user_id` unique, holding `timezone` and a `working_hours` JSONB blob. It arrived
later, in `3c02ebe2`, and took the name the schema field was already using.

Nothing reconciles the two. `from_attributes` reads `user.availability`, gets either `None` or a
`UserAvailability` instance, and neither is a `list[AvailabilitySlot]`.

## Reproducing

```console
$ cd backend && pytest app/tests -q
25 failed, 87 passed, 6 skipped
```

19 of those 25 are this, and most are collateral — the fixture registers a user and the
registration itself is what fails:

```
app/tests/test_collaboration_status.py::test_get_collaboration_status_default
>       user_id, token = register_and_login(
E       fastapi.exceptions.ResponseValidationError: 1 validation error:
E         {'type': 'list_type', 'loc': ('response', 'availability'),
E          'msg': 'Input should be a valid list', 'input': None}
E         POST /api/auth/register
```

Nested under a wrapper the path is `('response', 'user', 'availability')`, but it is the same
field:

```
ERROR devlink.request: {'method': 'POST', 'path': '/api/auth/refresh', 'status': 500, ...
  fastapi.exceptions.ResponseValidationError: 1 validation error:
    {'type': 'list_type', 'loc': ('response', 'user', 'availability'),
     'msg': 'Input should be a valid list', 'input': None}
```

## What is affected

Everything that serialises a user. From the failing suites:

- `POST /api/auth/register` — 500
- `POST /api/auth/refresh` — 500, so no session can be refreshed
- `test_followers.py` — 9 tests, follow/unfollow/counts/is-following
- `test_profile_views.py` — profile view endpoints
- `test_refresh_token_rotation.py` — both rotation tests
- `test_collaboration_status.py` — all 4
- `test_notifications.py::test_follow_notification`

The error is raised after the handler returns, so the write has already happened. Registration
creates the account and then answers 500; the caller has no way to tell that from a failure, and
retrying hits a duplicate-email error on an account that does exist.

`default_factory=list` does not help — it applies when the attribute is absent, not when it is
present and `None`.

Note that this is a plain response-validation failure on the happy path, not an edge case: a
brand-new user has no `user_availability` row, so *every* registration takes it.

## Suggested fix

The two meanings need separating. Either:

- rename the ORM relationship (`availability_settings`) and leave the schema field to mean the
  weekly slots it was written for, or
- rename the schema field and give `UserAvailability` a response model of its own.

Whichever way round, `UserResponse` must produce a value of the declared type for a user with no
availability row, and there should be a test that registers a user and asserts 200 — the suite
currently has 19 tests that would have caught this and they are all failing for the same reason.

Possibly the same underlying report as #667, which has a screenshot of a registration failure and
no text.

MOHITKOURAV01:
Closes #1346.

## The bug

`User` declares `availability` twice.

`app/models/user.py:156` is a JSON column holding the weekly slots the resume
parser fills in and `UserResponse.availability` serialises. 285 lines later,
`:441` was a scalar relationship to `UserAvailability` — a different feature
entirely, one row per user holding a timezone and a `working_hours` blob — under
the same name.

The second class attribute replaces the first. The column became unreachable,
every user serialised the relationship instead, and a user with no
`user_availability` row has `None` there against a `list[AvailabilitySlot]`
field:

```
fastapi.exceptions.ResponseValidationError: 1 validation error:
  {'type': 'list_type', 'loc': ('response', 'availability'),
   'msg': 'Input should be a valid list', 'input': None}
  POST /api/auth/register
```

Which is **every registration**, because a brand-new user has no such row.

FastAPI validates after the handler returns, so the account was created and then
reported as a 500. The caller cannot distinguish that from a failure, and
retrying hits a duplicate-email error on an account that exists.
`/api/auth/refresh` went the same way, so no session could be refreshed either.

## The fix

The relationship becomes `availability_settings`. Nothing referenced it —
`app/routers/availability.py` queries `UserAvailability` directly — so this is
the model, the `back_populates` on the other side, and nothing else.

```python
>>> User.availability.property.__class__.__name__
'ColumnProperty'                 # the column, reachable again
>>> User.availability_settings.property.__class__.__name__
'_RelationshipDeclared'
```

`default_factory=list` does not cover the second half. The column is nullable
and its `default=list` is Python-side, so it is `[]` only for rows this
application inserted; anything older or inserted by a migration is NULL, and a
default applies when an attribute is *absent*, not when it is present and
`None`. A `mode="before"` validator reads NULL as "no slots".

## Results

```console
$ pytest app/tests -q
10 failed, 107 passed, 6 skipped        # was 25 failed, 87 passed
```

All 15 availability failures are gone. The remaining 10 are #1317 (followers),
#1318 (naive `utcnow`) and the two rate-limit tests, none of which touch this.

## Tests

19 tests were failing on this and not one of them named it — they all went
through a fixture that registers a user. `app/tests/test_user_availability_response.py`
names it: registration returns 201 with `availability == []`, `/api/auth/refresh`
returns 200, the column and the relationship are separately addressable and both
work, and a NULL column serialises as `[]`.

Rename the relationship back and 2 of the 5 fail, so they are testing the thing
they claim to.


MOHITKOURAV01:

nensii21:

PR #1355, opened 2026-08-25
chore: untrack build output and scratch files, and make the ignore rules bite

### LINKED ISSUE #1348: Build output, tool dumps and a both-sides conflict "resolver" are tracked in the repository

MOHITKOURAV01:
Closes #1348.

## What was tracked

About 1.1 MB of build output, tool dumps and one-off scripts.

`dist/` is the one worth stating plainly: **520 KB of compiled JS and CSS in the
index, while `dist/` sat in `.gitignore` the whole time.**

```console
$ git check-ignore -q dist/index.html && echo ignored
ignored
$ git ls-files dist/index.html
dist/index.html
```

Both at once, because an ignore rule does not apply to a file git has already
recorded. From either end it looked fine — the rule was there, and the files
were there. This is the exact failure the `Tracked Files` workflow was written
for after `backend/.env`, recurring one directory over.

Also removed:

| file | |
| --- | --- |
| `bandit_report.json` | 363 KB security scan of a checkout nobody can date, sitting next to the code as though it described it |
| `frontend/tsc_errors_2.txt` | 70 KB of type errors that do not match the ones the branch has |
| `frontend/tsc_errors.txt`, `pr_info.json` | empty |
| `google_patch.diff`, `ms_patch.diff` | 134 KB of diffs against bases that no longer exist |
| `old_sections.tsx`, `upstream_sections.tsx` | hand-kept copies of `features/dashboard/sections.tsx` from the #1233 cleanup |
| `frontend/files_changed.txt` | |
| `.vite/deps/` | bundler cache |
| `.vscode/settings.json` | one machine's Windows interpreter path — what `.vscode/` already says should not be here |

## The two scripts

This is the part that is not housekeeping.

**`resolve_conflicts.py`** resolves a conflict by deleting the markers and
keeping both sides:

```python
for line in lines:
    if line.startswith('<<<<<<< HEAD'):
        in_conflict = True                     # assigned, never read
    elif line.startswith('======='):
        pass
    elif line.startswith('>>>>>>> upstream/main'):
        in_conflict = False
    else:
        out.append(line)                       # both sides survive
```

That is the state 19 files are in on `main` right now (#1344), and the state no
grep for `<<<<<<<` can detect afterwards. It runs on import, against a hardcoded
list from a previous incident.

**`fix_syntax.sh`** walks a file's history checking out older commits until
`prettier` stops complaining, discarding whatever was there.

Neither is something to leave loaded at the repository root.

## Why the guard missed all of it

`Tracked Files` enumerated the filenames of the *last* incident:

```
grep -E '(^|/)(fix|fix_[^/]*|patch_[^/]*|resolve_merge)\.py$|(^|/)pr[0-9]+\.json$'
```

It knows `resolve_merge.py`; the file here is `resolve_conflicts.py`. It knows
`pr123.json`; the file here is `pr_info.json`. Both miss by one character class,
and `dist/`, `.vite/`, the `.diff` dumps and the `tsc_errors` files are not
described at all.

The enumeration is widened, but the real change is a check that will not need
updating next time:

```yaml
git ls-files --cached --ignored --exclude-standard
```

`git check-ignore` reports nothing for a file already in the index, which is
exactly why `dist/` looked fine from both sides. That command asks the other
question — *which tracked files do the ignore rules describe?* — and answers it.

**6 on the commit before this one, 0 on this branch.**

`.gitignore` grows the categories that were missing (`.vite/`, tool reports,
`*_patch.diff`, the scratch scripts) so the new check has something to bite on.

## Not changed

Build output stays on disk (`git rm --cached`) — it is regenerable and now
actually ignored. The `.env` rules and the "templates must stay committable"
check are untouched and still pass:

```console
$ git check-ignore .env backend/.env frontend/.env backend/.env.local
.env
backend/.env
frontend/.env
backend/.env.local
$ git check-ignore backend/.env.example frontend/.env.example
                                        # correctly empty
```


MOHITKOURAV01:

PR #1393, opened 2026-08-26
fix(users): require an administrator to change account state

### LINKED ISSUE #1388: Three `/api/users` routes change account state with no authentication: anyone can disable any account
## Summary

Three `PATCH` routes on `app/routers/users.py` change account state and take no caller at all.
They depend on `get_database` and nothing else, so an unauthenticated request from anywhere on
the internet can disable any account, re-enable any account, and mark any address verified.

`POST /api/users/` sits in the same router with the same gap and is a second, unrated
registration path that bypasses everything `/api/auth/register` does.

## The routes

```python
# app/routers/users.py:706
@router.patch("/{user_id}/deactivate", response_model=UserResponse)
def deactivate_user(
    user_id: uuid.UUID,
    db: Session = Depends(get_database),
):
    user = UserService.get_user(db, user_id)
    ...
    return UserService.deactivate_user(db, user)
```

`activate_user` (`:707`) and `verify_user` (`:751`) are the same shape. There is no
`current_user`, no `require_roles`, no `dependencies=[...]` on the decorator, and the router is
mounted bare:

```python
# app/main.py:520
app.include_router(users.router, prefix="/api/users", tags=["Users"])
```

## What each one does

**`/deactivate` — account lockout for anybody, including every admin.** `deactivate_user` sets
`is_active = False` (`app/services/user_service.py:317`), and login refuses an inactive account:

```python
# app/services/auth_service.py:249
if not user.is_active:
    logger.warning("Login blocked: inactive account %s", user.id)
    raise HTTPException(status_code=403, detail="Account is disabled.")
```

`get_current_active_user` (`app/dependencies.py:191`) rejects it too, so existing sessions stop
working as well. User IDs are not secret — every profile response carries one — so this is a
one-request lockout of any named person, and a loop over the leaderboard is a lockout of the
whole platform. Nothing logs who did it: `deactivate_user` writes no audit row.

**`/verify` — the verification gate and the verified badge, for free.** `verify_email` sets
`is_verified = True` (`app/services/user_service.py:416`). That flag is what
`get_current_verified_user` (`app/dependencies.py:212`) checks, and it is what renders the badge
on every post:

```python
# app/routers/posts.py:46
verified=db_post.author.is_verified,
```

So an attacker verifies their own throwaway account without ever receiving the email, and can
also stamp the badge on an account they do not own.

**`/activate` — undo a suspension.** Whatever moderation did by clearing `is_active`, one
anonymous `PATCH` puts back.

**`POST /api/users/` — a second registration path.** `create_user` (`:102`) hashes the password
and inserts the row directly. It skips the rate limit on `/api/auth/register`, skips the password
breach screening, skips the verification email, and skips `is_active`/`is_verified` defaults
being set by `AuthService.register` — it relies on column defaults instead.

## Reproduction

```console
$ curl -s -X PATCH https://<host>/api/users/<any-user-id>/deactivate
{"id":"...","username":"victim","is_active":false, ...}

$ curl -s -X POST https://<host>/api/auth/login \
       -H 'content-type: application/json' \
       -d '{"email":"victim@example.com","password":"<their real password>"}'
{"detail":"Account is disabled."}
```

No `Authorization` header is sent at either step.

## Why it was not caught

`tests/test_users.py` exercises these routes without a token and asserts `200`, so the suite
currently encodes the bug as the expected behaviour.

## Expected

- `/activate`, `/deactivate` and `/verify` are administrative actions. They should require an
  authenticated caller with the system admin role, and should refuse a self-target for
  `/deactivate` so an admin cannot lock themselves out by accident.
- Each should write an audit log row naming the actor and the target.
- Deactivating an account should revoke its refresh tokens, otherwise a "disabled" account keeps
  minting access tokens until the refresh token expires.
- `POST /api/users/` should either be removed in favour of `/api/auth/register`, or be restricted
  to admins.
- The tests should assert `401` for an anonymous caller and `403` for an authenticated
  non-admin.

Found while auditing the routers for endpoints that mutate state without a caller — the same
sweep also turned up the conversation, plugin and reputation routes filed alongside this.

MOHITKOURAV01:
## Summary

`PATCH /api/users/{id}/activate`, `/deactivate` and `/verify` depended on `get_database` and
nothing else. An unauthenticated request could disable any account, re-enable any account, and
mark any address verified. Login refuses an inactive account, so `/deactivate` was a one-request
lockout of any user id — and user ids appear on every profile response.

This restricts the three routes to system administrators, moves the transitions into a service
that records who did what, and makes deactivation actually end the target's sessions.

---

## Related Issue

Fixes #1388

---

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation update
- [ ] Performance improvement
- [x] Refactoring
- [ ] UI/UX enhancement
- [x] Tests
- [ ] Other

---

## Changes Made

**Authorization.** The three state routes and `POST /api/users/` now depend on
`require_roles(SystemRole.ADMIN)`. `POST /api/users/` is not the sign-up route — `/api/auth/register`
is, and it is the one that rate-limits, screens the password against the breach list, and sends
the confirmation email — so leaving it open was a second, unrated registration path.

**New `app/services/account_state_service.py`.** The transitions used to be `UserService` setters
that took a `User` and assigned an attribute: no actor, no record, no guard. Everything an
administrator can do to somebody else's account goes through this module now, and every
transition names an actor, writes an audit row, and applies its side effects in the same unit of
work as the flag.

**Deactivation revokes refresh tokens.** This is the part that turns the flag into a suspension.
`AuthService.login` refuses an inactive account, but `POST /api/auth/refresh` validates the JWT
and the stored token row and consults neither — so a "disabled" account kept minting access
tokens for the remaining lifetime of its refresh token. `RefreshTokenService.revoke_all_tokens`
now runs as part of the transition, and it runs even when the account was already inactive,
because an account can be marked inactive by one path while still holding tokens issued before
it.

**Self-targeting refused for `/deactivate` and `/verify`.** Self-deactivation took effect
immediately, `get_current_active_user` then rejected the admin's own token, and `/activate` was
no longer reachable by that account — a mis-pasted id was an unrecoverable lockout. `/activate` is
left self-targetable: it is the harmless direction, and an admin whose own account is inactive
cannot reach the route at all.

**Idempotence without noise.** Re-activating an active account or re-verifying a verified one
returns 200 and writes no audit row. A log full of transitions that changed nothing is a log
nobody reads.

**`USER_EMAIL_VERIFIED` audit action**, with a migration. An administrator marking an address
verified without a confirmation token is a different event from the user following the emailed
link, and it is the one that decides whether the verified badge renders on every post the account
writes (`app/routers/posts.py:46`). The migration adds the value to the Postgres enum inside an
`autocommit_block` with `IF NOT EXISTS`, and returns early on any other dialect — SQLite renders
the column as a VARCHAR with a CHECK constraint, so there is nothing to alter there.

**Optional `reason`** on all three routes, stored on the audit row. The body stays optional so
the routes remain callable with no body at all, and a whitespace-only reason is normalised to
absent rather than written as `{"reason": "  "}`.

**`resolve_target_user` bypasses the cache.** `UserService.get_user` is decorated with `@cached`;
reading an account's flags from a cache immediately before writing them is how a transition gets
applied to a stale copy.

---

## Testing

- [x] Tested locally
- [x] Existing tests pass
- [x] Added new tests

Six tests in `tests/test_users.py` called these routes with no token and asserted `200` — they
encoded the bug as the expected behaviour. They now authenticate as an admin, and three new tests
cover the anonymous and non-admin cases.

`tests/test_user_account_state_authz.py` is new (28 tests):

- anonymous rejected on all three routes, and the account genuinely untouched afterwards —
  including an end-to-end check that the victim can still log in;
- ordinary user rejected on all three, and specifically cannot self-verify;
- admin can deactivate, the target then cannot log in, and their live refresh tokens are gone;
- admin can reactivate and the target can log in again;
- a non-admin probing ids cannot tell 404 from 403, while an admin gets the honest 404;
- self-deactivation and self-verification refused, self-activation allowed;
- idempotent calls return 200 and write no audit row, but repeated deactivation still revokes;
- audit rows carry actor, target, entity type and before/after values;
- reason recorded, blank reason dropped, over-long reason rejected with 422, empty body accepted.

Full backend suite compared against `main` in a clean worktree: **89 failures before, 89 after,
no regressions and no newly-broken tests.** The 89 are pre-existing and unrelated (the UTF-8 BOM
in `project_calendar.py`, the websocket tests, `test_rbac`, `test_organizations` and others).

```
baseline: 89  branch: 89
=== REGRESSIONS ===
=== FIXED ===
```

---

## Checklist

---

## Additional Notes

**Migration ordering.** `down_revision` is `1c63e6c28cf8`, the current single head. If another PR
lands a migration first this will need a rebase to point at the new head — it does not conflict
textually, so `git` will not flag it.

**There is no API for granting the admin role.** The tests set `system_role` directly, which is
what a seed script does. If a deployment currently has no admin, these routes become unreachable
until one is created — that is the intended posture for administrative routes, but it is worth
knowing before this is deployed.

ECSoc 2026

### DIFF
--- a/backend/alembic/versions/a3f7c1d29b48_add_user_email_verified_audit_action.py
+++ b/backend/alembic/versions/a3f7c1d29b48_add_user_email_verified_audit_action.py
@@ -1,0 +1,49 @@
+"""add user_email_verified to the auditaction enum
+
+An administrator marking an address verified without a confirmation token is a
+distinct event from the account being activated, and it is the one that decides
+whether the verified badge renders. It needs its own value rather than being
+folded into ``user_activated``.
+
+Revision ID: a3f7c1d29b48
+Revises: 1c63e6c28cf8
+Create Date: 2026-08-26 16:40:00.000000
+
+"""
+
+from alembic import op
+
+# revision identifiers, used by Alembic.
+revision = "a3f7c1d29b48"
+down_revision = "1c63e6c28cf8"
+branch_labels = None
+depends_on = None
+
+
+ENUM_NAME = "auditaction"
+NEW_VALUE = "user_email_verified"
+
+
+def upgrade() -> None:
+    bind = op.get_bind()
+
+    # SQLite (the test database) has no enum types -- the column is a VARCHAR
+    # with a CHECK constraint that SQLAlchemy renders from the Python enum, so
+    # there is nothing to alter and nothing to guard against.
+    if bind.dialect.name != "postgresql":
+        return
+
+    # ALTER TYPE ... ADD VALUE cannot run inside a transaction block on
+    # PostgreSQL below 12, and alembic wraps migrations in one. IF NOT EXISTS
+    # additionally makes this safe to re-run against a database where the value
+    # was already added by an autogenerate pass.
+    with op.get_context().autocommit_block():
+        op.execute(f"ALTER TYPE {ENUM_NAME} ADD VALUE IF NOT EXISTS '{NEW_VALUE}'")
+
+
+def downgrade() -> None:
+    # PostgreSQL has no ALTER TYPE ... DROP VALUE. Removing it would mean
+    # recreating the type and rewriting every column that uses it, which would
+    # fail anyway for any row already carrying the value. Leaving an unused
+    # label in place is harmless.
+    pass
--- a/backend/app/models/audit_log.py
+++ b/backend/app/models/audit_log.py
@@ -49,6 +49,7 @@ class AuditAction(str, Enum):
     USER_ACTIVATED = "user_activated"
     USER_BANNED = "user_banned"
     USER_UNBANNED = "user_unbanned"
+    USER_EMAIL_VERIFIED = "user_email_verified"
     SETTINGS_CHANGED = "settings_changed"
 
     # Project
--- a/backend/app/routers/users.py
+++ b/backend/app/routers/users.py
@@ -19,9 +19,21 @@ from sqlalchemy.orm import Session
 
 from app.core.security import hash_password
 from app.core.cache import cache_manager, cached
-from app.dependencies import get_current_user, get_database, get_optional_current_user
+from app.core.rbac import SystemRole
+from app.dependencies import (
+    get_current_user,
+    get_database,
+    get_optional_current_user,
+    require_roles,
+)
 from app.middleware.rate_limit import SEARCH_LIMIT, limiter
 from app.models.user import User
+from app.schemas.user_account_state import AccountStateChangeRequest
+from app.services.account_state_service import (
+    AccountStateService,
+    RequestContext,
+    resolve_target_user,
+)
 from app.services.block_service import BlockService
 from app.schemas.user import (
     CollaborationStatus,
@@ -98,10 +110,19 @@ def check_username(
     "/",
     response_model=UserResponse,
     status_code=status.HTTP_201_CREATED,
+    summary="Create a user (administrators only)",
+    description=(
+        "Provision an account directly. This is not the sign-up route -- "
+        "`POST /api/auth/register` is, and it is the one that rate-limits, "
+        "screens the password against the breach list, and sends the "
+        "confirmation email. This route skips all three, so it is restricted "
+        "to administrators."
+    ),
 )
 def create_user(
     user: UserCreate,
     db: Session = Depends(get_database),
+    _actor: User = Depends(require_roles(SystemRole.ADMIN)),
 ):
 
     if UserService.get_by_email(db, user.email):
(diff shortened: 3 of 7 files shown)


MOHITKOURAV01:
Coordination note for whoever merges this: **#1397 also adds a migration off `1c63e6c28cf8`**, the current single head.

Both PRs are individually clean — the `Revision graph` check passes on each, because each has exactly one head *on its own branch*. Merging both leaves two heads, and `alembic upgrade head` then fails the way #926 described. Git will not flag it: the two migration files have different names and never touch the same lines, so the conflict is semantic rather than textual.

Whichever lands second needs a one-line change — repoint its `down_revision` at the other's revision id:

- this PR: `a3f7c1d29b48` (adds `user_email_verified` to the `auditaction` enum)
- #1397: `b7e4d2f8c916` (adds `granted_by_id` to `reputation_logs`)

Happy to do the rebase on whichever one you merge second — tell me which and I will push it, or merge them in either order and I will follow up.

MOHITKOURAV01:
I test-merged all five of my open PRs (#1393, #1394, #1395, #1396, #1397) onto `main` in a scratch worktree to check for conflicts before you have to. Results:

**No textual conflicts.** All five merge clean, in order, with no overlapping files:

```
fix/user-account-state-authz-1388             MERGED CLEAN
fix/conversation-membership-authz-1389        MERGED CLEAN
fix/org-api-key-ownership-1390                MERGED CLEAN
fix/plugin-dispatch-authz-1391                MERGED CLEAN
fix/reputation-award-authz-1392               MERGED CLEAN
```

**One semantic conflict, exactly as flagged.** #1393 and #1397 each add a migration off `1c63e6c28cf8`, so the combination has two alembic heads and `alembic upgrade head` fails:

```
$ alembic heads          # with all five merged
a3f7c1d29b48 (head)
b7e4d2f8c916 (head)
```

The repo already has a test for this, and it is the only thing that breaks when the two are combined. Full backend suite on the merged result, against a `main` baseline in the same worktree:

```
baseline: 89   all-five-merged: 89
=== REGRESSIONS FROM COMBINING ===
FAILED tests/test_migration_graph.py::test_graph_has_exactly_one_head
=== FIXED ===
FAILED tests/test_conversations.py::test_router_prevent_self_messaging_integration
```

So: no functional interference between the five, one pre-existing failure cleared by #1394, and one head collision to resolve.

**The fix is one line in whichever merges second** — repoint its `down_revision` at the other:

| PR | file | revision |
| --- | --- | --- |
| #1393 | `a3f7c1d29b48_add_user_email_verified_audit_action.py` | `a3f7c1d29b48` |
| #1397 | `b7e4d2f8c916_add_granted_by_to_reputation_logs.py` | `b7e4d2f8c916` |

Merge them in either order and I will push the rebase on the second one straight away — or apply it yourself, it is genuinely just the one `down_revision` string. The per-branch `Revision graph` check passes on both today because each branch has one head on its own; it cannot see across PRs, which is why this needed checking by hand.

Issue #1235, opened 2026-08-20
A database built by `alembic upgrade head` is missing 32 of the 70 tables the ORM declares


MOHITKOURAV01:
A database built by `alembic upgrade head` is missing 32 of the 70 tables the ORM declares

## Summary

`backend/entrypoint.sh` runs `alembic upgrade head` and then starts the app. That upgrade produces a schema that is missing **32 of the 70 tables** in `Base.metadata`, because a long run of features shipped models (and routers, and services) without ever adding a migration.

The gap is invisible to the test suite, which builds its schema with `Base.metadata.create_all` (`backend/tests/conftest.py:95`) and therefore always has every table. It is also invisible to the `Migrations` workflow, which asserts only that the upgrade succeeds and that `len(tables) > 20`.

The first thing that breaks is not a feature — it is `request_logs`, which `RequestLoggingMiddleware` inserts into on **every** request.

## Reproduction

```bash
cd backend
alembic upgrade head          # succeeds; graph is fine, single head
python - <<'PY'
import re, glob, sys
sys.path.insert(0, ".")
import app.models
from app.database.base import Base

created = set()
for f in glob.glob("alembic/versions/*.py"):
    s = open(f).read()
    created |= set(re.findall(r"create_table\(\s*[\"']([^\"']+)[\"']", s))
    created |= set(re.findall(r"create_table\(\s*\n\s*[\"']([^\"']+)[\"']", s))

missing = sorted(set(Base.metadata.tables) - created)
print(len(missing), "tables in the ORM that no migration creates:")
print("\n".join("  - " + t for t in missing))
PY
```

Output — 32 tables:

```
api_keys                          project_announcements
badges                            project_documents
centralized_analytics_events      project_template_favorites
feature_announcement_reads        project_templates
feature_announcements             project_versions
global_announcements              project_views
maintenance_windows               reputation_logs
message_drafts                    request_logs
password_reset_tokens             search_click_logs
plugin_installations              search_query_logs
plugins                           security_events
posts                             user_badges
profile_suggestion_dismissals     user_blocks
                                  user_feedbacks
                                  user_reports
                                  verification_requests
                                  webhook_dead_letter_queue
                                  webhook_deliveries
                                  workspace_api_tokens
```

## Impact on a migration-built database

- **Every request fails or logs an error** — `request_logs` does not exist and the request-logging middleware writes a row per request.
- **Password reset is broken** — `password_reset_tokens` does not exist.
[...]

PR #1397, opened 2026-08-26
fix(reputation): require an administrator to adjust a score

### LINKED ISSUE #1392: `POST /api/reputation/award` has no authorization: any user can grant themselves any number of points
## Summary

`POST /api/reputation/award` authenticates the caller and then does no authorization at all. The
caller names the target, the action, and the number of points. All three are taken as given.

```python
# app/routers/reputation.py:79
def award_reputation(
    payload: ReputationAwardRequest,
    current_user: User = Depends(get_current_user),
    db: Session = Depends(get_db),
):
    target_user_id = payload.user_id or current_user.id
    _, log_entry = ReputationService.award_reputation(
        db=db,
        user_id=target_user_id,
        action=payload.action,
        points_override=payload.points,
        description=payload.description,
    )
    return ReputationLogResponse.model_validate(log_entry)
```

`current_user` is used for exactly one thing: as the default when `payload.user_id` is omitted.
It is never compared to `target_user_id`, and no role is checked.

## Nothing constrains the amount either

`points` is an unbounded optional integer:

```python
# app/schemas/reputation.py:32
points: Optional[int] = Field(
    default=None,
    description="Optional custom point value override.",
)
```

No `ge`, no `le`. And `action` is a free `str` — the `examples=[...]` list is documentation, not
validation — so it never has to correspond to anything in `ACTION_POINTS`:

```python
# app/services/reputation_service.py:70
if points_override is not None:
    pts = points_override
else:
    pts = ACTION_POINTS.get(action.lower(), 10)
```

The `ACTION_POINTS` table (`:24`) and its 10–100 point range are decoration whenever
`points_override` is present.

## Reproduction

```console
$ curl -s -X POST https://<host>/api/reputation/award \
       -H "Authorization: Bearer $ANY_USER" -H 'content-type: application/json' \
       -d '{"action":"anything","points":999999}'
{"id":"...","user_id":"<me>","action":"anything","points":999999, ...}

$ curl -s https://<host>/api/reputation/me -H "Authorization: Bearer $ANY_USER"
{"reputation_score":999999,"rank_tier":"Legend 👑", ...}
```

Top of the leaderboard, one request, from a day-old account. Pointing `user_id` at somebody else
works identically, and negative `points` is the same request with a minus sign — the floor at
`:78` clamps the score to `0`, so the cheapest attack is not inflating yourself, it is zeroing
everyone above you.

The front end already calls this endpoint, so it is live, not dormant:

```ts
// frontend/src/api/modules/reputation.ts:58
return api.post<ReputationLog>("/api/reputation/award", data);
```

There are no other callers. `award_reputation` is invoked from nowhere inside the backend:

```console
$ grep -rn "award_reputation" backend/app --include='*.py'
backend/app/routers/reputation.py:79:def award_reputation(
backend/app/routers/reputation.py:90:    _, log_entry = ReputationService.award_reputation(
backend/app/services/reputation_service.py:53:    def award_reputation(
```

Which is the deeper problem: reputation is supposed to be a *derived* signal — merged PRs,
completed projects, endorsements — and nothing in the platform derives it. The only way a score
changes is somebody asserting it over HTTP.

## Second, smaller fault: the leaderboard counts users it does not rank

```python
# app/services/reputation_service.py:164
total_stmt = select(func.count(User.id))
total = db.scalar(total_stmt) or 0

users_stmt = (
    select(User)
    .order_by(desc(User.reputation_score), desc(User.created_at))
    .offset(skip).limit(limit)
)
```

Neither statement filters on `User.is_active` or `User.deleted_at`. Both columns exist
(`app/models/user.py:299`, `:421`) and every comparable query in the codebase applies them —
`search_service.py:310`, `analytics_service.py:167`, `recommendation_service.py:1262`.

So the leaderboard ranks deactivated and soft-deleted accounts alongside live ones and publishes
their `username`, `full_name` and `avatar_url`, and `total` is the count of every row in `users`
rather than of the ranked set — so the client's page count is wrong.

## Expected

- Awarding to *another* user requires the system admin role.
- Self-award via HTTP is not a thing that should exist. `payload.user_id` should be required, and
  the "defaults to me" behaviour dropped.
- `points_override` is an admin-only field. A non-admin path, if one is kept, resolves points
  from `ACTION_POINTS` only.
- `action` is validated against `ACTION_POINTS` — an enum or a `Literal`, so an unknown action is
  a 422 rather than a silent 10 points.
- `points` gets bounds even for admins, and a negative award is a distinct, audited operation.
- Every award writes an audit log row naming the actor, not just the recipient — `ReputationLog`
  records `user_id` (who received) and has no column for who granted.
- The leaderboard filters to active, non-deleted users, and `total` counts the same set it ranks.

MOHITKOURAV01:
## Summary

`POST /api/reputation/award` authenticated the caller and then used that identity for exactly one
thing: as the default target. It took the recipient, the action and the number of points as
given, with no authorization and no bounds.

```python
target_user_id = payload.user_id or current_user.id
```

---

## Related Issue

Fixes #1392

---

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation update
- [ ] Performance improvement
- [x] Refactoring
- [ ] UI/UX enhancement
- [x] Tests
- [ ] Other

---

## Changes Made

**Administrators only.** The endpoint now depends on `require_roles(SystemRole.ADMIN)`.

**`user_id` is required.** It was optional and defaulted to the caller. "Omit the target and it
means me" is the shape of a self-service score, which is not what this endpoint is for.

**`action` is an enum, not a free string.** The `examples=[...]` list on the field was
documentation, not validation, and `ACTION_POINTS.get(action.lower(), 10)` then silently awarded
ten points for a typo or an invented name. Unknown actions are a 422 now, and an import-time check
raises if a `ReputationAction` member has no entry in `ACTION_POINTS` — so the ten-point fallback
cannot come back by accident. `MANUAL_ADJUSTMENT` is added, worth zero on its own, so an explicit
`points` override has an action to travel under rather than being smuggled in under
`merged_pull_request`.

**`points` is bounded** to ±500, in the schema *and* in the service. The service repeats the check
because `award_reputation` is callable from places that never see the schema, and a bound that
only exists at the edge is one refactor away from being gone.

**Every adjustment is attributable.** `ReputationLog` recorded who *received* points and had no
column for who granted them. It gains `granted_by_id` — nullable, because entries the platform
awards itself have no human actor and neither does any row that already exists — and the route
also writes an audit log entry naming both parties.

**The leaderboard filtered nothing.** Neither the count nor the page had a predicate, so it
ranked deactivated and soft-deleted accounts alongside live ones — publishing their `username`,
`full_name` and `avatar_url` — and `total` was `count(User.id)` over every row in `users`, which
disagreed with the set being listed and made the client's page count wrong. Both statements now
share one predicate, so they cannot drift.

**Frontend client updated** to match: `user_id` required, `action` typed against the same set,
`granted_by_id` on the log, and `MAX_POINTS_PER_AWARD` exported.

---

## Testing

- [x] Tested locally
- [x] Existing tests pass
- [x] Added new tests

**All four faults were reproduced against `main`** with a throwaway test in a clean worktree:

```
[REP 1] self-award 999999 as a plain user -> 201
[REP 1] my score -> 999999 | tier -> Legend 👑
[REP 2] zeroing a rival -> 201
[REP 2] rival score -> 0
[REP 3] unknown action -> 201 points: 10
[REP 4] deactivated account still ranked -> True
```

Note REP 2: deduction was the *cheaper* attack. Since the service floors the score at 0, zeroing
out everyone above you was one request each — no arms race of inflation required.

`tests/test_reputation_award_authz.py` is new (21 tests):

- **Who may call it** — anonymous 401; an ordinary user cannot award themselves, cannot award
  someone else, and cannot deduct from a rival (with the rival's score asserted intact
  afterwards); an admin can, and gets the table's value.
- **Target required** — omitting `user_id` is a 422 rather than meaning "me"; an unknown user is a
  404.
- **Validation** — unknown action 422 with the score unchanged; points above the ceiling and below
  the floor both 422; an override exactly at the ceiling accepted; an admin deduction still floors
  at zero.
- **`resolve_points` directly** — every enum member has a point value (the guard against the
  ten-point fallback), unknown action raises 422, an over-large override raises 422.
- **Attribution** — the response and the stored row both carry `granted_by_id`; a service-level
  award with no actor leaves it null.
- **Leaderboard** — deactivated accounts drop out, soft-deleted accounts drop out, `total` equals
  the number of entries before and after a deactivation, and ranking is still by score descending.

Full backend suite compared against `main` in a clean worktree: **89 failures before, 89 after,
no regressions.**

---

## Checklist

---

## Additional Notes

**Breaking API change, deliberately.** `user_id` becomes required and `action` becomes a closed
set. Any client sending an ad-hoc action string now gets a 422 — which is the point.

**The vulnerability had a button.** `routes/_app.leaderboard.tsx` carried a "Claim Reputation
Points" dialog that posted to `/api/reputation/award` with no `user_id`, which the API resolved to
the caller. `typecheck` caught it after my first push. I removed the control rather than retyping
it: awarding is administrative now, and `/api/users/me` returns `is_superuser` but not
`system_role`, so the client cannot honestly gate the affordance — a button that 403s for almost
everyone is worse than no button. The "How to Earn Reputation Points" card stays.

**Nothing awards reputation automatically.** `award_reputation` has no callers anywhere in the
backend; the HTTP endpoint is the only way a score has ever changed. That is the deeper problem
underneath this issue: reputation is meant to be *derived* from merged PRs, completed projects and
endorsements, and none of that is wired up. This PR makes the manual path safe and auditable; it
does not build the derived one. Worth its own issue.

**The migration also creates the table.** `reputation_logs` is one of the models #1235 flags as
declared in the ORM and never created by any migration, so `alembic upgrade head` produces a
database without it and an unconditional `ALTER TABLE` fails outright — which is exactly what CI
reported. The migration creates the table in the shape the model declares when it is absent, and
adds only the column when it is already there. Since no earlier revision knows about the table,
`downgrade` drops it; leaving it behind breaks `alembic downgrade base`, because its foreign key
to `users` blocks `DROP TABLE users` further down the chain. That is slightly more than the title
promises, but a migration that cannot run against a migrated database is not a migration.

Verified against Postgres 16 locally, running the same three steps the workflow does — upgrade to
head (46 tables, `reputation_logs` present with both foreign keys and all three indexes, and
`created_at` as `timestamp with time zone`), the schema assertions, and downgrade to base.

**Migration ordering.** `down_revision` is `1c63e6c28cf8`, the current single head. #1393 also
adds a migration off that head; whichever lands second will need a one-line rebase of its
`down_revision`, since the conflict is semantic (two alembic heads) rather than textual and git
will not flag it.

ECSoc 2026

### DIFF
--- a/backend/alembic/versions/b7e4d2f8c916_add_granted_by_to_reputation_logs.py
+++ b/backend/alembic/versions/b7e4d2f8c916_add_granted_by_to_reputation_logs.py
@@ -1,9 +1,18 @@
 """add granted_by_id to reputation_logs
 
-The log recorded who received points and never who granted them, so a manual
-adjustment could not be attributed to anyone. Nullable: entries the platform
-awards itself have no human actor, and neither does any row that already
-exists.
+``ReputationLog`` recorded who *received* points and had no column for who
+granted them, so a manual adjustment could not be attributed to anyone.
+``granted_by_id`` is nullable: entries the platform awards itself have no human
+actor, and neither does any row that already exists.
+
+``reputation_logs`` is also one of the tables that ``app/models`` declares and
+no migration ever created -- see #1235 -- so ``alembic upgrade head`` produces a
+database without it, and an unconditional ``ALTER TABLE`` here fails with
+``relation "reputation_logs" does not exist``. This migration therefore creates
+the table when it is absent and only adds the column when it is already there.
+That is a little more than the title promises, but a migration that cannot run
+against a migrated database is not a migration -- and it means ``downgrade``
+drops the table, since no earlier revision has it.
 
 Revision ID: b7e4d2f8c916
 Revises: 1c63e6c28cf8
@@ -13,6 +22,7 @@ Create Date: 2026-08-26 17:30:00.000000
 
 import sqlalchemy as sa
 from alembic import op
+from sqlalchemy import inspect
 from sqlalchemy.dialects import postgresql
 
 # revision identifiers, used by Alembic.
@@ -21,39 +31,103 @@ down_revision = "1c63e6c28cf8"
 branch_labels = None
 depends_on = None
 
+TABLE = "reputation_logs"
+COLUMN = "granted_by_id"
+INDEX = "ix_reputation_logs_granted_by_id"
 
-def upgrade() -> None:
+
+def _table_exists(bind) -> bool:
+    return TABLE in inspect(bind).get_table_names()
+
+
+def _column_exists(bind) -> bool:
+    return COLUMN in {c["name"] for c in inspect(bind).get_columns(TABLE)}
+
+
+def _create_table() -> None:
+    """Create ``reputation_logs`` in the shape ``app/models/reputation.py`` declares."""
+    op.create_table(
+        TABLE,
+        sa.Column(
+            "id",
+            postgresql.UUID(as_uuid=True),
+            primary_key=True,
+            nullable=False,
+        ),
+        sa.Column(
+            "user_id",
+            postgresql.UUID(as_uuid=True),
+            sa.ForeignKey("users.id", ondelete="CASCADE"),
+            nullable=False,
+        ),
+        sa.Column("action", sa.String(length=50), nullable=False),
+        sa.Column("points", sa.Integer(), nullable=False),
+        sa.Column("description", sa.String(length=255), nullable=True),
+        sa.Column(
+            COLUMN,
+            postgresql.UUID(as_uuid=True),
+            sa.ForeignKey("users.id", ondelete="SET NULL"),
+            nullable=True,
+        ),
+        sa.Column(
+            "created_at",
+            sa.DateTime(timezone=True),
+            server_default=sa.text("now()"),
+            nullable=False,
+        ),
+    )
+    op.create_index("ix_reputation_logs_user_id", TABLE, ["user_id"])
+    op.create_index("ix_reputation_logs_action", TABLE, ["action"])
+    op.create_index(INDEX, TABLE, [COLUMN])
+
+
+def _add_column(bind) -> None:
     op.add_column(
-        "reputation_logs",
-        sa.Column("granted_by_id", postgresql.UUID(as_uuid=True), nullable=True),
+        TABLE,
+        sa.Column(COLUMN, postgresql.UUID(as_uuid=True), nullable=True),
     )
-    op.create_index(
-        "ix_reputation_logs_granted_by_id",
-        "reputation_logs",
-        ["granted_by_id"],
-    )
+    op.create_index(INDEX, TABLE, [COLUMN])
 
     # SQLite cannot add a foreign key to an existing table with ALTER, and the
     # test database is SQLite. The constraint is skipped there; the column and
     # the index -- which is what queries need -- are created either way.
-    if op.get_bind().dialect.name == "postgresql":
+    if bind.dialect.name == "postgresql":
         op.create_foreign_key(
             "fk_reputation_logs_granted_by_id_users",
-            "reputation_logs",
+            TABLE,
             "users",
-            ["granted_by_id"],
+            [COLUMN],
             ["id"],
             ondelete="SET NULL",
         )
 
 
-def downgrade() -> None:
-    if op.get_bind().dialect.name == "postgresql":
-        op.drop_constraint(
-            "fk_reputation_logs_granted_by_id_users",
-            "reputation_logs",
-            type_="foreignkey",
-        )
+def upgrade() -> None:
+    bind = op.get_bind()
 
-    op.drop_index("ix_reputation_logs_granted_by_id", table_name="reputation_logs")
-    op.drop_column("reputation_logs", "granted_by_id")
+    if not _table_exists(bind):
+        _create_table()
+        return
+
+    if not _column_exists(bind):
+        _add_column(bind)
+
+
+def downgrade() -> None:
+    """Drop the table.
+
+    No earlier revision knows about ``reputation_logs`` -- the model was added
+    without a migration -- so the state this reverts to is one where the table
+    does not exist, and dropping it is the honest inverse. Leaving it behind
+    also breaks ``alembic downgrade base``: its foreign key to ``users`` blocks
+    ``DROP TABLE users`` further down the chain.
+
+    Reputation history is lost on downgrade, which is what "downgrade past the
+    revision that introduced the table" means. Take a backup first.
+    """
+    bind = op.get_bind()
+
+    if not _table_exists(bind):
+        return
+
+    op.drop_table(TABLE)
--- a/backend/alembic/versions/b7e4d2f8c916_add_granted_by_to_reputation_logs.py
+++ b/backend/alembic/versions/b7e4d2f8c916_add_granted_by_to_reputation_logs.py
@@ -1,0 +1,59 @@
+"""add granted_by_id to reputation_logs
+
+The log recorded who received points and never who granted them, so a manual
+adjustment could not be attributed to anyone. Nullable: entries the platform
+awards itself have no human actor, and neither does any row that already
+exists.
+
+Revision ID: b7e4d2f8c916
+Revises: 1c63e6c28cf8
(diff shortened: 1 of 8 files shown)


MOHITKOURAV01:
Pushed two CI fixes. Both failures were real, and one of them found something worth calling out.

**`Upgrade and downgrade` — `relation "reputation_logs" does not exist`**

That table is one of the models #1235 flags as declared in the ORM and never created by any migration, so `alembic upgrade head` produces a database without it and my `ALTER TABLE` had nothing to alter. The migration now creates the table in the shape `app/models/reputation.py` declares when it is absent, and adds only the column when it is already there. Because no earlier revision knows about the table, `downgrade` drops it — leaving it behind also broke `alembic downgrade base`, since its foreign key to `users` blocks `DROP TABLE users` further down the chain.

Verified against Postgres 16 locally, running the same three steps the workflow does:

```
=== UPGRADE HEAD ===
Running upgrade 1c63e6c28cf8 -> b7e4d2f8c916, add granted_by_id to reputation_logs
46 tables; reputation_logs: True
schema assertions pass
=== DOWNGRADE BASE ===
Running downgrade f533805006ed -> , Create users table
FULL CYCLE OK
```

Columns, both foreign keys (`user_id` CASCADE, `granted_by_id` SET NULL) and all three indexes match the model, and `created_at` comes out as `timestamp with time zone`.

**`typecheck` — `src/routes/_app.leaderboard.tsx(105,9): Type 'string' is not assignable to type 'ReputationAction'`**

This one caught a caller I had missed, and reading it was more interesting than the type error. The leaderboard page carried a **"Claim Reputation Points"** dialog that posted to `/api/reputation/award` with no `user_id` — which the API resolved to the caller. So the vulnerability was not only reachable, it had a button, and the button said *"Claiming..."*.

I removed the control rather than retyping it. Awarding is an administrative action now, and `/api/users/me` returns `is_superuser` but not `system_role`, so the client cannot honestly gate the affordance — a button that 403s for almost everyone is worse than no button. The "How to Earn Reputation Points" card stays; that part of the page was already telling the truth.

I have corrected the claim in the PR description that nothing but the API module called this endpoint.

MOHITKOURAV01:
I test-merged all five of my open PRs (#1393, #1394, #1395, #1396, #1397) onto `main` in a scratch worktree to check for conflicts before you have to. Results:

**No textual conflicts.** All five merge clean, in order, with no overlapping files:

```
fix/user-account-state-authz-1388             MERGED CLEAN
fix/conversation-membership-authz-1389        MERGED CLEAN
fix/org-api-key-ownership-1390                MERGED CLEAN
fix/plugin-dispatch-authz-1391                MERGED CLEAN
fix/reputation-award-authz-1392               MERGED CLEAN
```

**One semantic conflict, exactly as flagged.** #1393 and #1397 each add a migration off `1c63e6c28cf8`, so the combination has two alembic heads and `alembic upgrade head` fails:

```
$ alembic heads          # with all five merged
a3f7c1d29b48 (head)
b7e4d2f8c916 (head)
```

The repo already has a test for this, and it is the only thing that breaks when the two are combined. Full backend suite on the merged result, against a `main` baseline in the same worktree:

```
baseline: 89   all-five-merged: 89
=== REGRESSIONS FROM COMBINING ===
FAILED tests/test_migration_graph.py::test_graph_has_exactly_one_head
=== FIXED ===
FAILED tests/test_conversations.py::test_router_prevent_self_messaging_integration
```

So: no functional interference between the five, one pre-existing failure cleared by #1394, and one head collision to resolve.

**The fix is one line in whichever merges second** — repoint its `down_revision` at the other:

| PR | file | revision |
| --- | --- | --- |
| #1393 | `a3f7c1d29b48_add_user_email_verified_audit_action.py` | `a3f7c1d29b48` |
| #1397 | `b7e4d2f8c916_add_granted_by_to_reputation_logs.py` | `b7e4d2f8c916` |

Merge them in either order and I will push the rebase on the second one straight away — or apply it yourself, it is genuinely just the one `down_revision` string. The per-branch `Revision graph` check passes on both today because each branch has one head on its own; it cannot see across PRs, which is why this needed checking by hand.

